PR #2, opened 2017-07-07
cp exec util from kubernetes/kubernetes

mengqiy:
Removed BUILD.


mengqiy:
PTAL at the first move to check if this is what it should look like.

spxtr:
Save the git history! Here is an example PR that preserved git history: https://github.com/kubernetes/repo-infra/pull/20

PR #12, opened 2017-07-21
Clarify criteria for adding code.

monopole:
Wanted to have clearer criteria for adding code, stopping well short
of all the [policy](https://github.com/kubernetes/community/blob/master/contributors/devel/pull-requests.md) in the core repo.

Provide more history about why this repo exists and how it fits with the rest.

Underscore that this repo must not depend on other k8s repos.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,29 +1,38 @@
 # Utils
 
-[![Build Status](https://travis-ci.org/kubernetes/utils.svg?branch=master)](https://travis-ci.org/kubernetes/utils)
+[![Build Status]](https://travis-ci.org/kubernetes/utils)
 
-Utils is a set of golang libraries that are not specific to
-Kubernetes. They should be available and useful to any other Go project
-out there.
+A set of Go libraries that provide low-level,
+kubernetes-independent packages supplementing the [Go
+standard libs].
 
 ## Purpose
 
-The Kubernetes project uses a lot of Golang patterns that are re-used in many
-difference places, and as the project is further split into different
-repositories, this repository is the perfect place to hold this common code.
+As Kubernetes grows and spins functionality out of its
+[core] and into cooperating repositories like
+[apiserver], [kubectl], [kubeadm], etc., the need
+arises for leaf repositories to house shared code and
+avoid cycles in repository relationships.
 
-## Goals
+This repository is intended to hold shared utilities
+with no Kubernetes dependence that may be of interest
+to any Go project.  See these [instructions for moving]
+an existing package to this repository.
 
-Go libraries in this repository must be:
 
-- Generic enough that they are useful for external/non-kubernetes
-  projects,
-- Well factored, well tested and reliable,
-- Be completely go compliant (go get/build/test/etc)
-- Have enough complexity to be shared,
-- Have stable APIs, or backward compatible.
+## Criteria for adding code here
 
-The goal is to keep libraries organized in logical entities.
+- Used by multiple Kubernetes repositories.
+
+- Full unit test coverage.
+
+- Go tools compliant (`go get`, `go test`, etc.).
+
+- Complex enough to be worth vendoring, rather than copying.
+
+- Stable, or backward compatible, API.
+
+- _No dependence on any Kubernetes repository_.
 
 ## Libraries
 
@@ -31,6 +40,12 @@ The goal is to keep libraries organized in logical entities.
   to mock and replace in tests, especially with
   the [FakeExec](exec/testing/fake_exec.go) struct.
 
-## How to move a pkg from other repo
-
-Please look at [HOWTOMOVE.md](./HOWTOMOVE.md).
+[Build Status]: https://travis-ci.org/kubernetes/utils.svg?branch=master
+[Go standard libs]: https://golang.org/pkg/#stdlib
+[api]: https://github.com/kubernetes/api
+[apiserver]: https://github.com/kubernetes/apiserver
+[core]: https://github.com/kubernetes/kubernetes
+[ingress]: https://github.com/kubernetes/ingress
+[kubeadm]: https://github.com/kubernetes/kubeadm
+[kubectl]: https://github.com/kubernetes/kubectl
+[instructions for moving]: ./HOWTOMOVE.md


README.md
```diff
@@ -1,36 +1,59 @@
...
 - [Exec](/exec) provides an interface for `os/exec`. It makes it easier
   to mock and replace in tests, especially with
   the [FakeExec](exec/testing/fake_exec.go) struct.
 
-## How to move a pkg from other repo
-
-Please look at [HOWTOMOVE.md](./HOWTOMOVE.md).
+[Build Status]: https://travis-ci.org/kubernetes/utils.svg?branch=master
+[Go standard libs]: https://golang.org/pkg/#stdlib
+[api]: https://github.com/kubernetes/api
```
mengqiy: It seems that these are not visible in preview.
May need to change the format.

apelisse: These work for me.

README.md
```diff
@@ -1,36 +1,59 @@
...
 
 ## Purpose
 
-The Kubernetes project uses a lot of Golang patterns that are re-used in many
-difference places, and as the project is further split into different
-repositories, this repository is the perfect place to hold this common code.
+As Kubernetes grows and spins functionality out of its
+[core] and into cooperating repositories
+like [apiserver], [kubectl], [kubeadm], etc.,
+the need arises for leaf repos to house shared code
```
apelisse: I'm not a super fan of using "repo(s)" rather than "repository(ies)". Also, while saying that, I notice that there are ... 8 occurrences of "repo" in this section :-)

README.md
```diff
@@ -1,36 +1,59 @@
...
-Go libraries in this repository must be:
+## Criteria for adding code here
 
-- Generic enough that they are useful for external/non-kubernetes
-  projects,
-- Well factored, well tested and reliable,
-- Be completely go compliant (go get/build/test/etc)
-- Have enough complexity to be shared,
-- Have stable APIs, or backward compatible.
+- Used by multiple Kubernetes repos.
```
apelisse: I don't know if I want this to be the first/most important criteria though. I feel like being generic is more important.

README.md
```diff
@@ -1,36 +1,59 @@
...
 
-The goal is to keep libraries organized in logical entities.
+- Full unit test coverage.
```
apelisse: What do you mean by that? 100% coverage? that'd be an ambitious statement, that we don't really have a good way to verify :-)

README.md
```diff
@@ -1,36 +1,59 @@
...
+
+- Go tools compliant (`go get`, `go test`, etc.).
+
+- Complex enough to be worth vendoring, rather than copying.
```
apelisse: I couldn't find a good way to say that, thank you

README.md
```diff
@@ -1,36 +1,59 @@
...
   the [FakeExec](exec/testing/fake_exec.go) struct.
 
-## How to move a pkg from other repo
-
-Please look at [HOWTOMOVE.md](./HOWTOMOVE.md).
+[Build Status]: https://travis-ci.org/kubernetes/utils.svg?branch=master
+[Go standard libs]: https://golang.org/pkg/#stdlib
+[api]: https://github.com/kubernetes/api
+[apiserver]: https://github.com/kubernetes/apiserver
+[common]: https://github.com/kubernetes/common
```
apelisse: That common repo doesn't exist yet, and I already see it being named differently by various people. Can we avoid mention it altogether here?

README.md
monopole: changed to full word

README.md
monopole: i don't know how to quantify thinks like generic or well factored.

i'm thinking :  used by multiple repos

can be used as a means to help reviewers prevent people from gaming this repo as a means to add.change functionality to their own projects surreptitiously by changing code here.   seen it before :)

README.md
monopole: I'm thinking the code here should be simple enough to
cover
https://docs.travis-ci.com/user/coveralls/

README.md
monopole: We all agree it has to come, because there has to be a place to put shared utilities with type awareness (dependence on the api repo).

I think we should mention it here, and explain how the two repos differ, since obviously their names are interchangeable and intentionally communicate no information.   The README for the common repo should also mention this repo as a place to put things that don't have k8s dependence.

README.md
monopole: yeah these work

README.md
mengqiy: We can announce both repoes after they are both setup.

apelisse:
@mengqiy Do you have an opinion on that before I merge?

mengqiy:
I'm fine merging this, but we should not announce this PR until the `kubernetes/common` has been created. We really should work a bit on the other repo, at least ask Tim or Daniel to create it.

monopole:
Text updated to point out that the "common" repo is _proposed_ (i.e. it doesn't exist yet).

Mentioning it here helps define this repo, and answer the implicit question of where to put shared code that does depend on k8s.  Once this is in, we can file an issue to get the other repo created and refer to this repo's README.

README.md
```diff
@@ -1,36 +1,59 @@
...
-## Goals
+This repository is intended to hold shared utilities
+with no Kubernetes dependence that may be of interest
+to any Go project.  See these [instructions for moving]
+an existing package to this repository.
 
-Go libraries in this repository must be:
+A proposed sibling repository, with the proposed name [common],
+will hold shared utilities that may depend on both this
+repository and other near-leaf repositories like [api].
```
apelisse: My understanding was that `common` would only depend on `utils`, with the only difference being that it would be "kubernetes specific"

monopole: Imagine the code review debating whether something is kubernetes specific and not generally useful.  Defining these repos in terms of (objective) deps avoids subjective debates.

apelisse: Agreed, but `common` should still be more common than API shouldn't it?

monopole: One could argue (I certainly don't want to) that `api` is by definition common and thus should be a subdirectory in the common repo.  

So - one could add some more words here to say "The common repo will have common code that isn't already in some other already-existing repo that's already intended for vendoring." but meh.

monopole:
Hi folks, all references to the common repo have been deleted.
Can we submit this?

apelisse:
Looks good to me. @mengqiy waiting for your :+1:

PR #13, opened 2017-07-21
Add goverall

apelisse:
No idea if this works, or what it does.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,10 @@ go:
   - 1.7.x
   - 1.8.x
 go_import_path: k8s.io/utils
+before_install:
+  - go get github.com/mattn/goveralls
 script:
   - diff -u <(echo -n) <(gofmt -d .)
   - go tool vet .
   - go test -v -race ./...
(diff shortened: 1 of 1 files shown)


monopole:
sgtm !

mengqiy:
@apelisse travis failed.

mengqiy:
It seems travis-ci expect an test file for `k8s.io/utils/exec/testing`, which doesn't have a test file.

Unfortunately, there is not test for `k8s.io/kubernetes/pkg/util/interrupt`.

apelisse:
No, I think the problem is that we need an account on coveralls.io. And I'm too busy to look at it right now, so I'm just leaving that there for the moment.

I'm not sure about your comment about `k8s.io/kubernetes/pkg/util/interrupt`. Indeed, there is no test-file, I don't think that is a problem for travis, but I really believe that we should add it as we try to merge it in this repo.

apelisse:
I'll actually close this for now to remove confusion. Will re-open when I'm ready for it.

PR #14, opened 2017-08-09
Add useful test wrapper functions to FakeExec

danwinship:
Migrated from https://github.com/kubernetes/kubernetes/pull/46537...

----

The fake implementation of pkg/util/exec for test programs is kind of really annoying to use. This PR adds some wrapper functions to make it easier, AND to automatically check that the passed-in command was what you expected (which some tests were doing but most weren't).

So:

    fcmd := fakeexec.FakeCmd{
            CombinedOutputScript: []fakeexec.FakeCombinedOutputAction{
                    // iptables version check
                    func() ([]byte, error) { return []byte("iptables v1.9.22"), nil },
                    // iptables-restore version check
                    func() ([]byte, error) { return []byte("iptables-restore v1.9.22"), nil },
                    // Success.
                    func() ([]byte, error) { return []byte{}, nil },
            },
    }
    fexec := fakeexec.FakeExec{
            CommandScript: []fakeexec.FakeCommandAction{
                    func(cmd string, args ...string) exec.Cmd { return fakeexec.InitFakeCmd(&fcmd, cmd, args...) },
                    func(cmd string, args ...string) exec.Cmd { return fakeexec.InitFakeCmd(&fcmd, cmd, args...) },
                    func(cmd string, args ...string) exec.Cmd { return fakeexec.InitFakeCmd(&fcmd, cmd, args...) },
            },
    }

becomes

    fexec := fakeexec.NewFakeExec(t, nil)
    fexec.AddCommand("iptables", "--version").
    	SetCombinedOutput("iptables v1.9.22", nil)
    fexec.AddCommand("iptables-restore", "--version").
    	SetCombinedOutput("iptables-restore v1.9.22", nil)
    fexec.AddCommand("iptables", "-w2", "-N", "FOOBAR", "-t", "nat").
    	SetCombinedOutput("", nil)

There's some discussion on the original kubernetes bug (which I'm going to rebase on top of this to cover only the "porting pkg/util to the new interface" part). The June 27 comment about it not passing "make verify" no longer seems to be a problem with the migration of fakeexec into a separate subdirectory.

### DIFF
--- a/exec/testing/fake_exec.go
+++ b/exec/testing/fake_exec.go
@@ -19,6 +19,11 @@ package testingexec
 import (
 	"fmt"
 	"io"
+	osexec "os/exec"
+	"reflect"
+	"runtime"
+	"strings"
+	"testing"
 
 	"k8s.io/utils/exec"
 )
@@ -28,13 +33,34 @@ type FakeExec struct {
 	CommandScript []FakeCommandAction
 	CommandCalls  int
 	LookPathFunc  func(string) (string, error)
+
+	T *testing.T
+}
+
+// NewFakeExec creates a FakeExec. You can pass nil for t if you are not running from a
+// test program. (This will cause certain failures to result in a panic rather than a
+// t.Fatal.) cmdsInPath is a list of command names for which LookPath will return a match
+// ("/fake-bin/commandname"); if your code does not use LookPath you can pass nil. (If you
+// need more complex behavior, you can override LookPathFunc on the returned FakeExec.)
+func NewFakeExec(t *testing.T, cmdsInPath []string) *FakeExec {
+	return &FakeExec{
+		T: t,
+		LookPathFunc: func(cmd string) (string, error) {
+			for _, pathCmd := range cmdsInPath {
+				if pathCmd == cmd {
+					return "/fake-bin/" + cmd, nil
+				}
+			}
+			return "", &osexec.Error{Name: cmd, Err: osexec.ErrNotFound}
+		},
+	}
 }
 
 type FakeCommandAction func(cmd string, args ...string) exec.Cmd
 
 func (fake *FakeExec) Command(cmd string, args ...string) exec.Cmd {
 	if fake.CommandCalls > len(fake.CommandScript)-1 {
-		panic(fmt.Sprintf("ran out of Command() actions. Could not handle command [%d]: %s args: %v", fake.CommandCalls, cmd, args))
+		fake.fail("ran out of Command() actions at %s executing %v", getCaller(1), append([]string{cmd}, args...))
 	}
 	i := fake.CommandCalls
 	fake.CommandCalls++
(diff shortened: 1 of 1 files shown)


exec/testing/fake_exec.go
```diff
@@ -45,6 +71,82 @@ func (fake *FakeExec) LookPath(file string) (string, error) {
...
+	// Don't re-fail from a "defer fexec.AssertExpectedCommands()" if we already failed once
+	if fake.T.Failed() {
+		return
+	}
+	fake.T.Fatal(msg)
+}
+
+// AddCommand adds a Cmd to be returned by a future fake.Command(...) call. Call
+// SetCombinedOutput or SetRunOutput on the result to specify its result.
+func (fake *FakeExec) AddCommand(cmd string, args ...string) *FakeCmd {
```
apelisse: I like that this makes things easier, but I'd decouple the command creation a little bit.

Can we have:
```
// Hides the logic of appending commands
func (fake *FakeExec) AddCommand(cmd *FakeCmd) { ... }
```

And something else to create the command
```
// Hides the pain of creating commands
func NewMatchingArgsCommand(cmd string, args ...string) *FakeCmd {
    return func(cmd string, args ...string) exec.Cmd {
        // Some of the logic you have in your command
    }
}
```

Then it means that:
1. I can add command easily, but they don't have to follow your logic
2. I can invent new types of commands

exec/testing/fake_exec.go
```diff
@@ -28,13 +33,34 @@ type FakeExec struct {
...
+		},
+	}
 }
 
 type FakeCommandAction func(cmd string, args ...string) exec.Cmd
 
 func (fake *FakeExec) Command(cmd string, args ...string) exec.Cmd {
 	if fake.CommandCalls > len(fake.CommandScript)-1 {
-		panic(fmt.Sprintf("ran out of Command() actions. Could not handle command [%d]: %s args: %v", fake.CommandCalls, cmd, args))
+		fake.fail("ran out of Command() actions at %s executing %v", getCaller(1), append([]string{cmd}, args...))
```
monopole: +1  prefer leaving the "fail" and the error reporting to the client.

exec/testing/fake_exec.go
monopole: again +1

apelisse:
Also, this is complicated enough that we may want to test it?

exec/testing/fake_exec.go
danwinship: Well, but the Command still has to do *something* when the code being tested calls `CombinedOutput()`/`Run()` on it. So you could have it return an error to the caller, but then you're testing how your code responds when it gets an error that doesn't correspond to anything it could actually get in the real world.

If you hit this point (or any of the other panic/fail points), then there is a bug in your test case, not in the code that it's testing. So the only reasonable behavior is to abort the test case.

exec/testing/fake_exec.go
danwinship: If you want to add a command in a different way, you can just continue to use the existing FakeExec API. There is currently only 1 test program in all of kubernetes that wants to do something other than what the API above supports. (`pkg/volume/git_repo/git_repo_test.go`)

exec/testing/fake_exec.go
```diff
@@ -45,6 +71,82 @@ func (fake *FakeExec) LookPath(file string) (string, error) {
...
+// stdout and stderr can be either []byte or string (which will be converted to []byte).
+func (fcmd *FakeCmd) SetRunOutput(stdout, stderr interface{}, err error) {
+	fcmd.RunScript = []FakeRunAction{
+		func() ([]byte, []byte, error) { return outputAsBytes(stdout), outputAsBytes(stderr), err },
+	}
+}
+
+// SetCombinedOutput provides the result of calling CombinedOutput() on a FakeCmd.
+// output can be either a []byte or a string (which will be converted to a []byte).
+func (fcmd *FakeCmd) SetCombinedOutput(output interface{}, err error) {
```
danwinship: Literally every existing caller in kubernetes wants to pass a string rather than a []byte, so if you don't want it to accept both then I'd change the API to just take a string instead, and say people can use the lower-level API if they need to use []byte

exec/testing/fake_exec.go
```diff
@@ -45,6 +71,82 @@ func (fake *FakeExec) LookPath(file string) (string, error) {
...
+	fcmd.CombinedOutputScript = []FakeCombinedOutputAction{
+		func() ([]byte, error) { return outputAsBytes(output), err },
+	}
+}
+
+// AssertExpectedCommands ensures that all of the commands added to fake via AddCommand were actually executed
+func (fake *FakeExec) AssertExpectedCommands() {
```
danwinship: Convenience. That's all any of this is about; making FakeExec more complicated so that test programs can be more simple.

exec/testing/fake_exec.go
apelisse: I don't disagree with that, but then it should just probably just panic

exec/testing/fake_exec.go
apelisse: Yeah, I just don't really like the `interface{}`. string is fine.

exec/testing/fake_exec.go
```diff
@@ -28,13 +33,34 @@ type FakeExec struct {
...
+
+	T *testing.T
+}
+
+// NewFakeExec creates a FakeExec. You can pass nil for t if you are not running from a
+// test program. (This will cause certain failures to result in a panic rather than a
+// t.Fatal.) cmdsInPath is a list of command names for which LookPath will return a match
+// ("/fake-bin/commandname"); if your code does not use LookPath you can pass nil. (If you
+// need more complex behavior, you can override LookPathFunc on the returned FakeExec.)
+func NewFakeExec(t *testing.T, cmdsInPath []string) *FakeExec {
```
danwinship: > Why would I call this function in the first place if I want to change LookPathFunc, considering that's pretty much the only added value of this function?

I guess I like having New methods rather than manually constructing somebody-else's-struct-type by hand. I had added `func NewFakeExec(t *testing.T) *FakeExec` before adding the `cmdsInPath` arg.

But yes, I could do it the way you suggested instead.

exec/testing/fake_exec.go
apelisse: Yeah, then it should probably panic instead, and the name could be better

danwinship:
> Also, this is complicated enough that we may want to test it?

Yes, although ironically, the code using `testing.T` can't really be tested. Though the `panic`-ing versions of those codepaths could be.

apelisse:
> Yes, although ironically, the code using testing.T can't really be tested.

;-)

exec/testing/fake_exec.go
danwinship: (If using `testing.T` is absolutely unacceptable, then we can go back to using `panic`... I just preferred the `t.Fatalf` solution because it doesn't spew a gigantic pile of stack traces at you like `panic` does.)

danwinship: (fwiw, wrote that last comment before seeing your response)

exec/testing/fake_exec.go
apelisse: Yeah, but I think this would make it much easier to compose and re-use the code

thockin:
fejta-bot is coming after PRs now!

thockin:
@danwinship is this still alive?

PR #16, opened 2017-08-30
Update HOWTOMOVE doc

mengqiy:
We may encounter some error when moving pkg.
I remember this when I was working on https://github.com/kubernetes/kube-openapi/pull/9.


HOWTOMOVE.md
```diff
@@ -8,6 +8,7 @@ It has 2 steps to move a pkg from other Kubernetes repos to `k8s.io/utils` repo:
 
 Copying should preserve all the git history associated with it.
 [Here](http://gbayer.com/development/moving-files-from-one-git-repository-to-another-preserving-history/) is a working approach.
+Note: You may need to use `--allow-unrelated-histories` if you get error when running `git pull` following the post above.
```
apelisse: Have you used that so far? I'd really like not to use that as it can be really confusing ...

mengqiy: I used it twice: 1 month ago and yesterday.

PR #24, opened 2017-11-21
Add clock from apimachinery 

dashpole:
My end goal is to use the clock in cAdvisor.  Because of the way that staging works in kubernetes/kubernetes, the import loop kubernetes/apimachinery -> google/cadvisor -> kubernetes/kubernetes is not allowed.
I would copy the clock code to cadvisor, but cadvisor is currently google-owned, and I dont believe copying it from CNCF to google is permitted.

cc @tallclair


dashpole:
I brought over history from apimachinery

mengqiy:
I think this LG.
It seems this code is copied from `k8s.io/apimachinery` instead of `/staging` dir. So each commit has [`k8s-publish-bot`](https://github.com/k8s-publish-robot)'s commit. IMO it will be better to not have the bot commits.

And that publish bot has been down for weeks, so please double check what you copied is synced with the latest code in `/staging`.

@apelisse WDYT?

dashpole:
oh, I should copy directly from staging

dashpole:
give me a minute

apelisse:
Good catch Mengqi, I would never have noticed that

dashpole:
commits are now from kubernetes/kubernetes

mengqiy:
/lgtm
Waiting for travis to be green.

apelisse:
Yeah, I'd be happy to see more examples and better documentation, but won't delay any work on that of course.

apelisse:
One thing that I'd actually like to see change now (sooner rather than later) is:
- Move testing code to its own testing package, so that it's not included in production code.

dashpole:
when you import with godep, it doesnt bring in testing code, right?

dashpole:
or you mean just splitting the mock and real implementations?

apelisse:
> when you import with godep, it doesnt bring in testing code, right?

But `FakeClock` and `IntervalClock` live in `clock.go`, next to `RealClock`.

dashpole:
yes, agreed

clock/clock.go
```diff
@@ -0,0 +1,327 @@
...
+func (RealClock) Tick(d time.Duration) <-chan time.Time {
+	return time.Tick(d)
+}
+
+func (RealClock) Sleep(d time.Duration) {
+	time.Sleep(d)
+}
+
+// FakeClock implements Clock, but returns an arbitrary time.
+type FakeClock struct {
```
thockin: We should put fakes and test-related things in sub-packages, typically called "testing", so they don't get linked into prod binaries.

@dashpole

dashpole: done https://github.com/kubernetes/utils/pull/25/

apelisse: Yeah, we noticed that: #25

PR #25, opened 2017-11-21
Split real and fake clocks

dashpole:
All testing was for the fake clock, so that file is now in the testing sub-dir

/assign @apelisse 
cc @thockin

### DIFF
--- a/clock/testing/fake_clock.go
+++ b/clock/testing/fake_clock.go
@@ -1,0 +1,254 @@
+/*
+Copyright 2014 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


clock/testing/fake_clock_test.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package clock
+package testing
```
apelisse: we don't have any test for the real clock? I guess it's not that easy to test ...

clock/testing/fake_clock.go
```diff
@@ -0,0 +1,254 @@
...
+limitations under the License.
+*/
+
+package testing
+
+import (
+	"sync"
+	"time"
+
+	clockapi "k8s.io/utils/clock"
```
apelisse: Why not just clock?

clock/testing/fake_clock_test.go
dashpole: yeah, we are just using golang's time library

apelisse:
Thank you for taking care of that that quickly

clock/testing/fake_clock.go
dashpole: changed

thockin:
Thanks!

PR #26, opened 2017-11-29
exec supports context (timeout/cancel)

anfernee:
- Add function CommandContext as os/exec does
- Clean up the code a little bit to make it more readable
- Fix panic when calling Stop before Start

Fixes: https://github.com/kubernetes/kubernetes/issues/5644

### DIFF
--- a/exec/exec.go
+++ b/exec/exec.go
@@ -17,6 +17,7 @@ limitations under the License.
 package exec
 
 import (
+	"context"
 	"io"
 	osexec "os/exec"
 	"syscall"
@@ -32,6 +33,13 @@ type Interface interface {
 	// Command returns a Cmd instance which can be used to run a single command.
 	// This follows the pattern of package os/exec.
 	Command(cmd string, args ...string) Cmd
+
+	// CommandContext returns a Cmd instance which can be used to run a single command.
+	//
+	// The provided context is used to kill the process if the context becomes done
+	// before the command completes on its own. For example, a timeout can be set in
+	// the context.
+	CommandContext(ctx context.Context, cmd string, args ...string) Cmd
 
 	// LookPath wraps os/exec.LookPath
 	LookPath(file string) (string, error)
@@ -79,7 +87,7 @@ func New() Interface {
 
 // Command is part of the Interface interface.
 func (executor *executor) Command(cmd string, args ...string) Cmd {
-	return (*cmdWrapper)(osexec.Command(cmd, args...))
+	return &cmdWrapper{osexec.Command(cmd, args...)}
 }
 
 // LookPath is part of the Interface interface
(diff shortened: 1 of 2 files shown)


anfernee:
cc @apelisse

exec/exec.go
```diff
@@ -110,52 +125,51 @@ func (cmd *cmdWrapper) SetStderr(out io.Writer) {
...
+	if c.Process == nil {
 		return
 	}
+
 	c.Process.Signal(syscall.SIGTERM)
+
 	time.AfterFunc(10*time.Second, func() {
-		if c.ProcessState.Exited() {
-			return
+		if c.ProcessState != nil && !c.ProcessState.Exited() {
```
thockin: How could this be nil if c.Process was not?  If possible, it deserves a comment

exec/exec.go
```diff
@@ -79,16 +87,23 @@ func New() Interface {
...
 }
 
+// CommandContext is part of the Interface interface.
+func (executor *executor) CommandContext(ctx context.Context, cmd string, args ...string) Cmd {
+	return &cmdWrapper{osexec.CommandContext(ctx, cmd, args...)}
+}
+
 // Wraps exec.Cmd so we can capture errors.
-type cmdWrapper osexec.Cmd
+type cmdWrapper struct {
```
thockin: Why? This was defined explicitly NOT to inherit methods, and to force us to declare methods we want to pass-through.

exec/exec.go
```diff
@@ -165,10 +179,10 @@ type ExitErrorWrapper struct {
 	*osexec.ExitError
 }
 
-var _ ExitError = ExitErrorWrapper{}
+var _ ExitError = &ExitErrorWrapper{}
 
 // ExitStatus is part of the ExitError interface.
-func (eew ExitErrorWrapper) ExitStatus() int {
+func (eew *ExitErrorWrapper) ExitStatus() int {
```
apelisse: why is this needed?

exec/exec.go
```diff
@@ -33,6 +34,13 @@ type Interface interface {
 	// This follows the pattern of package os/exec.
 	Command(cmd string, args ...string) Cmd
 
+	// CommandContext returns a Cmd instance which can be used to run a single command.
```
apelisse: Thank you for documenting this!

apelisse:
`Context` didn't exist in go 1.6. I think we can remove that though. Would you mind opening another PR for that? Thanks.

anfernee:
thanks guys for quick review. I will break it into different commits. also I will reply your concerns here.

exec/exec.go
anfernee: because I feel this casting `(*osexec.Cmd)(cmd).CombinedOutput()` looks weird. if it's defined as embeded type then the casting can be avoided. it's basically the same as `ExitErrorWrapper` below. anyway, I don't have strong opinion on this. it could be as is.

exec/exec.go
anfernee: just to be cautious. there is no documented connection between `c.Process` and `c.ProcessState`. But docs do say it is available after calling Wait or Run. It should be safe. I am okay to leave it as is.

exec/exec.go
anfernee: because the pointer `*ExitErrorWrapper` is the `ExitError`. It passes by reference. previous code also works, but it de-references the pointer and passes into the function by value. Again, not a must-fix. L182 is required though.

anfernee:
close this one, splitted to #27 #28  other improvements are pending for decision.

thockin:
Regarding the cast - it was done that way to avoid inheriting methods.  It
could be done as a non-embedded, but I don't see a point.

PR #28, opened 2017-11-29
Fix panic calling stop before run

anfernee:


exec/exec.go
```diff
@@ -133,15 +133,17 @@ func (cmd *cmdWrapper) Output() ([]byte, error) {
 // Stop is part of the Cmd interface.
```
apelisse: I think that comment should explain what happens if it's called before `Run()` (and probably more). IOW, We should know what to expect from that function without having to read the code, since `godoc` shows you the comments without the code.

apelisse: https://godoc.org/k8s.io/utils/exec#Cmd

Actually, you might want to change the `Cmd` interface documentation, since this object is hidden/private.

PR #29, opened 2017-11-29
Fix cmd.Run() which should handle error but didn't

anfernee:


exec/exec.go
```diff
@@ -148,16 +143,19 @@ func (cmd *cmdWrapper) Stop() {
...
-		return x
+	if err == nil {
+		return nil
 	}
-	if ee, ok := err.(*osexec.Error); ok {
-		if ee.Err == osexec.ErrNotFound {
+
+	switch e := err.(type) {
+	case *osexec.ExitError:
+		return &ExitErrorWrapper{e}
```
thockin: We should probably change this to a non-pointer - it doesn't need top be heap, I think.

anfernee: I just make it minimal. I think adding `&` or not doesn't change where it is allocated. anyway, thanks for the quick review.

thockin:
It forces it to be on the heap, as far as I can tell. And yeah, this is
fine for now, but maybe a followup? :)

PR #34, opened 2018-01-30
Unstruct

apelisse:
Replaces #32. I've changed, improved and simplified the interface. The documentation looks pretty neat. I can build on top of that.

This is building the basic infrastructure to create generic tools/clients that work with no compiled-in knowledge of the types.

### DIFF
--- a/unstruct/example_test.go
+++ b/unstruct/example_test.go
@@ -1,0 +1,61 @@
+package unstruct_test
+
+import (
+	"fmt"
+
+	"k8s.io/utils/unstruct"
+
+	"gopkg.in/yaml.v2"
+)
+
+func Example() {
+	u := unstruct.New(map[string]interface{}{
+		"int": 1,
+		"list": []interface{}{
+			"A string.",
+			2,
+		},
+	})
+
+	u.Map().Field("int").Set("Number one")
+	u.Map().Field("string").Set("my string")
+	s := u.Map().Field("list").Slice()
+	s.Append("Another string.")
+	s.Append([]interface{}{})
+	s.At(3).Slice().Append(4)
+
+	y, _ := yaml.Marshal(u.Data())
+	fmt.Println(string(y))
+	// Output: int: Number one
+	// list:
+	// - A string.
+	// - 2
+	// - Another string.
+	// - - 4
+	// string: my string
+}
+
+// This example demonstrates how the interface lets you chain calls
+// to easily set and change values in a unstruct type.
+func Example_chaining() {
+	u := unstruct.New(nil)
+
+	u.Map().
+		Field("int").Set("Number one").Parent().Map().
+		Field("string").Set("my string").Parent().Map().
+		Field("list").Slice().
+		Append("A string.").Parent().Slice().
+		Append(2).Parent().Slice().
+		Append("Another string.").Parent().Slice().
+		Append(nil).Slice().Append(4)
+
+	y, _ := yaml.Marshal(u.Data())
+	fmt.Println(string(y))
+	// Output: int: Number one
+	// list:
+	// - A string.
+	// - 2
+	// - Another string.
+	// - - 4
+	// string: my string
+}
--- a/unstruct/interface.go
+++ b/unstruct/interface.go
@@ -86,3 +86,11 @@ type Value interface {
 func New(value interface{}) Value {
 	return &baseValue{data: value}
 }
+
+// Root returns the root for a given Value.
+func Root(v Value) Value {
+	for v.Parent() != nil {
+		v = v.Parent()
+	}
+	return v
+}
--- a/unstruct/interface.go
+++ b/unstruct/interface.go
@@ -84,5 +84,5 @@ type Value interface {
 
 // New creates a representation of the given Value.
 func New(value interface{}) Value {
-	nil
+	return &baseValue{data: value}
 }
--- a/unstruct/interface.go
+++ b/unstruct/interface.go
@@ -1,0 +1,88 @@
+package unstruct
+
+// Map lets you manipulate an object.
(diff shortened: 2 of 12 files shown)


thockin:
Can you include in the PR comment why we need this and who will use it?

unstruct/interface.go
```diff
@@ -0,0 +1,96 @@
...
+	// a slice.
+	//
+	// Returns a Value pointing to that item.
+	Append(value interface{}) Value
+
+	// InsertAt inserts at the index-th position in the slice.
+	//
+	// Returns the slice itself, so that the calls can be chained,
+	// or nil if the index is out-of-bound.
+	InsertAt(index int, value interface{}) Slice
```
thockin: Insert implies preserving the subsequent contents and shifting them one.  Is that what this is really doing?

apelisse:
Updated.

unstruct/interface.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// Value is the most generic representation of the data.
+type Value interface {
+	// Data returns the underlying data provided by the object.
+	Data() interface{}
+
+	// Parent returns the parent interface or nil if no parent is
+	// present (root).
+	Parent() Value
```
thockin: Shouldn't parent be defined in some base type that everything embeds?

unstruct/example_test.go
```diff
@@ -0,0 +1,61 @@
...
+func Example() {
+	u := unstruct.New(map[string]interface{}{
+		"int": 1,
+		"list": []interface{}{
+			"A string.",
+			2,
+		},
+	})
+
+	u.Map().Field("int").Set("Number one")
```
thockin: These examples are deceptively clean because they omit all error handling.  Once you add errors you get something more like:

```
m := u.Map()
if m == nil {
  return fmt.Errorf("expected a map")
}
f := m.Field("int")
if f == nil {
  return fmt.Errorf("expected a field 'int'")
}
f.Set("Number one")
```

and so on. Is that really any better than:

```
m, ok := u.(map[string]interface{})
if !ok {
  return fmt.Errorf("expected a map")
}
if _, found := m["int"]; !found {
  return fmt.Errorf("expected a field 'int'")
}
f["int"] = "Number one"
```

One way is idiomatic and understandable by any Go programmer (and net 1 LOC shorter).  The other way is a couple hundred LOC library, brand new, that people will have to learn.

Convince me it is actually worth doing?  I am not seeing it...

apelisse:
```
m := u.Map()
if m == nil {
  return fmt.Errorf("expected a map")
}
f := m.Field("int")
if f == nil {
  return fmt.Errorf("expected a field 'int'")
}
f.Set("Number one")
```

This can actually be slightly simplified to (4 loc shorter, yay ;-)):
```
m := u.Map()
if m == nil {
  return fmt.Errorf("expected a map")
}
f := m.Field("int").Set("Number one")
```
Since `m.Field("int")` creates the field for you.

The plan is to create this abstraction and tools that can be re-used with various implementations. For example, one of the end goal is to tie this "unstruct" values with the openapi definition. So that you can walk a json and its type definition at the same time. 

I'm also building a "jsonpath" engine that let's you find values and/or specific openapi types. This would probably be the recommended way to walk a json/yaml if you don't know if fields are present or of the proper type, since it simplifies error handling. Also, that engine allows you to use "selectors" as "predicates" such as `$.store.book[?(@.price < 10)]` [1]. Which means that at this point you know the "price" field exists for each "book" value that you have (this removes the need to check that "book" is a map, that has a field "price"). Having to re-cast everything from `interface{}` to maps to ... is kind of unnecessary.

[1]: http://goessner.net/articles/JsonPath/index.html#e2

unstruct/interface.go
apelisse: Yes, should this be clarified in the comment? (but insert implies ...)

unstruct/interface.go
apelisse: Yeah, I thought about that. But then you have an extra interface to remember about, and I don't think it makes anything simpler/cleaner. And I don't think one could use the "Parentable" interface by itself anyway (actually, the `Root` function below could take the `Parentable` instead, but I don't think it's worth the extra interface).

thockin:
> This can actually be slightly simplified to (4 loc shorter, yay ;-)):
>
> m := u.Map()
> if m == nil {
>   return fmt.Errorf("expected a map")
> }
> f := m.Field("int").Set("Number one")
>
> Since m.Field("int") creates the field for you.

Ugh, see, that's a grumpy interface - It's not at all obvious that it
has side-effects, and it's not clear if there is a side-effect-free
form.

> The plan is to create this abstraction and tools that can be re-used with various implementations. For example, one of the end goal is to tie this "unstruct" values with the openapi definition. So that you can walk a json and its type definition at the same time.
>
> I'm also building a "jsonpath" engine that let's you find values and/or specific openapi types. This would probably be the recommended way to walk a json/yaml if you don't know if fields are present or of the proper type, since it simplifies error handling. Also, that engine allows you to use "selectors" as "predicates" such as $.store.book[?(@.price < 10)] 1. Which means that at this point you know the "price" field exists for each "book" value that you have (this removes the need to check that "book" is a map, that has a field "price"). Having to re-cast everything from interface{} to maps to ... is kind of unnecessary.

Those sound cool, but who is using them (or going to be using them)?
I'd suggest starting this code wherever it is called, and then move it
here when it's utility is established.

I don't mean to be a pain, but we should try to keep this as tight as possible.

apelisse:
> I don't mean to be a pain, but we should try to keep this as tight as possible.

I share that goal. I'm with that, thanks!

PR #37, opened 2018-06-07
Fix cancelling timers in FakeClock package

munnerz:
This existing timer implementation used for testing does not properly support cancellation.

This is because of how pointers and copies are handled in the FakeClock structure.

Here, we determine whether to still retain a timer by comparing the registered waiters and the current timer/waiter, and if they do not match, we retain the timer (as it is not the timer being cancelled): https://github.com/kubernetes/utils/blob/master/clock/testing/fake_clock.go#L225-L240

However, we instantiate the `fakeClockWaiter` as a non-pointer within the Timer structure, and record this structure in a slice stored on the FakeClock. You can see that happening here: https://github.com/kubernetes/utils/blob/master/clock/testing/fake_clock.go#L89-L94

The issue is that on L94, we actually create a copy of the `fakeClockWaiter`, which means that the comparison on L232 will never match, and so *all* timers will be retained.

This PR changes the slice in the FakeClock to be a slice of *pointers*, which resolves the issue.
Unit tests still pass with this new adapted version, and my own tests that rely on this package are now passing too.

### DIFF
--- a/clock/testing/fake_clock.go
+++ b/clock/testing/fake_clock.go
@@ -34,7 +34,7 @@ type FakeClock struct {
 	time time.Time
 
 	// waiters are waiting for the fake time to pass their specified time
-	waiters []fakeClockWaiter
+	waiters []*fakeClockWaiter
 }
 
 type fakeClockWaiter struct {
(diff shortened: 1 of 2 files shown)


munnerz:
I've also added a unit test to verify that timer cancellation works 😄

munnerz:
Ping on this - I'd really like to use this package for unit tests in cert-manager.

@apelisse who should I chase to get another review/the merge button hit?

apelisse:
@thockin Do you want to merge that?

thockin:
Seems sane to me.  You are fully empowered to merge changes such as this, @apelisse :)

PR #38, opened 2018-06-11
Move mount code from kubernetes/kubernetes for reusing from CSI plugins

wgliang:
/kind feature
FIx [#64953](https://github.com/kubernetes/kubernetes/issues/64953)
/cc @smarterclayton 

> The mount code (pkg/util/mount) is reasonably well tested but brings in the entire Kubernetes source tree when CSI plugins have to depend on it. With a few exceptions, most of the mount code could be refactored to be a minimal set of utilities that can be layered and make it easier to develop CSI plugins. I noticed that quite a few CSI plugins take a dependency on k/k just to get that one package


mount/mount.go
```diff
@@ -0,0 +1,414 @@
...
+	"path/filepath"
+	"strings"
+)
+
+type FileType string
+
+const (
+	// Default mount command if mounter path is not specified
+	defaultMountCommand           = "mount"
+	MountsInGlobalPDPath          = "mounts"
```
thockin: My assumption is that this referes to how GCE PDs are being managed.   This needs to be generified

thockin: And commented

mount/mount.go
```diff
@@ -0,0 +1,414 @@
...
+	// PrepareSafeSubpath does everything that's necessary to prepare a subPath
+	// that's 1) inside given volumePath and 2) immutable after this call.
+	//
+	// newHostPath - location of prepared subPath. It should be used instead of
+	// hostName when running the container.
+	// cleanupAction - action to run when the container is running or it failed to start.
+	//
+	// CleanupAction must be called immediately after the container with given
+	// subpath starts. On the other hand, Interface.CleanSubPaths must be called
+	// when the pod finishes.
```
thockin: It would be nice to factor this API to not have a concept of pods.  If we need a "transaction" object to carry state, and that happens to map to a pod in general use, that's fine.  The bar for kubernetes/utils/ is higher than k/k wrt genericness and API.

mount/exec.go
```diff
@@ -0,0 +1,50 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mount
+
+import "k8s.io/utils/exec"
+
+func NewOsExec() Exec {
```
thockin: Is this used anywhere?  I don't see it.

wgliang: This is an externally available function. It is actually used in kubernetes.

thockin:
It seems like a silly function to me?

msau42:
/assign 
cc @kubernetes/sig-storage-pr-reviews

mount/mount.go
```diff
@@ -0,0 +1,416 @@
...
+	PrepareSafeSubpath(subPath Subpath) (newHostPath string, cleanupAction func(), err error)
+	// GetMountRefs finds all mount references to the path, returns a
+	// list of paths. Path could be a mountpoint path, device or a normal
+	// directory (for bind mount).
+	GetMountRefs(pathname string) ([]string, error)
+	// GetFSGroup returns FSGroup of the path.
+	GetFSGroup(pathname string) (int64, error)
```
msau42: FSGroup is a kubernetes concept so should not belong here either.

GetSELinuxSupport() and GetMode() are also not used by any plugins either.

msau42:
cc @jsafrane 
do the mount propagation and exec interfaces need to be here too?

msau42:
Maybe it would be easier to restructure the mount interface in Kubernetes core first.  Then pull the generic mount utils out.

PR #41, opened 2018-07-18
cp pointers util from kubernetes/kubernetes

stewart-yu:
We need move the` k8s.io/kubernetes/pkg/util/pointer` package to` k8s.io/utils/` repo, more details please refer to [PR](https://github.com/kubernetes/kubernetes/pull/66284) in `kubernetes` repo.

Following the [instruction](https://github.com/kubernetes/utils/blob/master/HOWTOMOVE.md), this PR is the first step

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ an existing package to this repository.
 
 - [Clock](/clock) provides an interface for time-based operations.  It allows
   mocking time for testing.
+- [Pointers](/pointers) provides some functions for pointer-based operations.
 
 [Build Status]: https://travis-ci.org/kubernetes/utils.svg?branch=master
 [Go standard libs]: https://golang.org/pkg/#stdlib
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ an existing package to this repository.
 
 - [Clock](/clock) provides an interface for time-based operations.  It allows
   mocking time for testing.
+  
 - [Pointers](/pointers) provides some functions for pointer-based operations.
 
 [Build Status]: https://travis-ci.org/kubernetes/utils.svg?branch=master
--- a/pointer/pointer.go
+++ b/pointer/pointer.go
@@ -1,0 +1,76 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package pointer
+
+import (
+	"fmt"
+	"reflect"
+)
+
+// AllPtrFieldsNil tests whether all pointer fields in a struct are nil.  This is useful when,
+// for example, an API struct is handled by plugins which need to distinguish
+// "no plugin accepted this spec" from "this spec is empty".
+//
+// This function is only valid for structs and pointers to structs.  Any other
+// type will cause a panic.  Passing a typed nil pointer will return true.
+func AllPtrFieldsNil(obj interface{}) bool {
+	v := reflect.ValueOf(obj)
+	if !v.IsValid() {
+		panic(fmt.Sprintf("reflect.ValueOf() produced a non-valid Value for %#v", obj))
+	}
+	if v.Kind() == reflect.Ptr {
+		if v.IsNil() {
+			return true
+		}
+		v = v.Elem()
+	}
+	for i := 0; i < v.NumField(); i++ {
+		if v.Field(i).Kind() == reflect.Ptr && !v.Field(i).IsNil() {
+			return false
+		}
+	}
+	return true
+}
+
+// Int32Ptr returns a pointer to an int32
+func Int32Ptr(i int32) *int32 {
+	return &i
+}
+
+// Int64Ptr returns a pointer to an int64
+func Int64Ptr(i int64) *int64 {
+	return &i
+}
+
+// Int32PtrDerefOr dereference the int32 ptr and returns it i not nil,
+// else returns def.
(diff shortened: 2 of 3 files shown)


stewart-yu:
seems no `OWNER` file in this repo
/assign @thockin @luxas @sttts 
PTAL, related kubernetes repo [PR](https://github.com/kubernetes/kubernetes/pull/66284), thanks

README.md
```diff
@@ -1,5 +1,6 @@
 # Utils
 
+
```
sttts: nit: why this empty line?

pointer/pointer.go
```diff
@@ -0,0 +1,71 @@
...
+	if ptr != nil {
+		return *ptr
+	}
+	return def
+}
+
+// BoolPtr returns a pointer to a bool
+func BoolPtr(b bool) *bool {
+	return &b
+}
```
sttts: StringPtr. Don't we have that 1000 times in kube?

stewart-yu: sure,  add it now

stewart-yu:
/cc  @thockin 
for review, thanks

README.md
```diff
@@ -45,6 +45,7 @@ an existing package to this repository.
 
 - [Clock](/clock) provides an interface for time-based operations.  It allows
   mocking time for testing.
+- [Pointers](/pointers) provides some functions for pointer-based operations.
```
apelisse: For some reason, we've kept an empty line between each module (don't ask me why), but maybe we should keep it consistent?

pointer/pointer.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	for i := 0; i < v.NumField(); i++ {
+		if v.Field(i).Kind() == reflect.Ptr && !v.Field(i).IsNil() {
+			return false
+		}
+	}
+	return true
+}
+
+// Int32Ptr returns a pointer to an int32
```
apelisse: AFAICR, Tim gave this as an example of what we shouldn't have in this repo (might remember it wrong, I'll look it up)

apelisse:
https://github.com/kubernetes/kubernetes/issues/48209#issuecomment-312305457

pointer/pointer.go
stewart-yu: seems `Int32Ptr()` had been used, if do not put here,  which file fo put it?

apelisse: The problem is not that it's not being used, the problem is that you're trying to share code with very little complexity.

luxas: It's gonna be used by many packages, and is generally useful, although very simple.

luxas:
On the other hand @apelisse https://github.com/kubernetes/kubernetes/pull/66010#pullrequestreview-135948977 :)

> This seems fine but why apimachinery and not https://github.com/kubernetes/utils ? I might clean up some of the APIs, if you want to put a little polish into it...

apelisse:
I guess I can't argue with that :-)

pointer/pointer_test.go
```diff
@@ -0,0 +1,66 @@
...
+			Bar *int
+		}{nil, new(int)}, false},
+		{&struct {
+			Foo *int
+			Bar *int
+		}{nil, new(int)}, false},
+		{(*struct{})(nil), true},
+	}
+	for i, tc := range testCases {
+		if AllPtrFieldsNil(tc.obj) != tc.expected {
```
apelisse: This could use `t.Run`

luxas:
Can you merge then to unblock us @apelisse :)?

luxas:
Thanks :tada:!

stewart-yu:
Thanks all

sttts:
Finally we have a generic pointer package 🎉🚀

pointer/pointer.go
```diff
@@ -0,0 +1,76 @@
...
+	return def
+}
+
+// BoolPtr returns a pointer to a bool
+func BoolPtr(b bool) *bool {
+	return &b
+}
+
+// StringPtr returns a pointer to the passed string.
+func StringPtr(s string) *string {
```
sttts: float64 is missing 😱

thockin:
This is in so I won't be a pain about it, but this is the sort of dependency that causes very interconnected dep graphs.  Any tool that uses this now has to vendor this whole repo.  Part of the go ethos is that it's better to copy a little code than to take a dependency.  Let's please not let this accumulate tons of random crap.

luxas:
No agree we shouldn't put "random crap" in this repo, but this exact utilpointer code is referenced by most of our API packages that deal with defaulting, so this is kinda-needed unless we want to start doing stuff like this in our defaulting packages (that will now live in `n` different repos)

```go
trueVar := true
obj.SomeBoolPointerField = &trueVar
```
vs

```go
obj.SomeBoolPointerField = utilpointer.BoolVar(true)
```

Well you know this already, but anyway I've made the argument for this now.

For me this could have moved to `k8s.io/apimachinery` equally fine but you said in https://github.com/kubernetes/kubernetes/pull/66010#pullrequestreview-135948977

> This seems fine but why apimachinery and not https://github.com/kubernetes/utils ? I might clean up some of the APIs, if you want to put a little polish into it...

So I'm a bit confused.
In any case I'm of the opinion this needs to live somewhere to not be a pain for the defaulting code for the ComponentConfigs.

> Any tool that uses this now has to vendor this whole repo

hmm I had the idea one can vendor individual packages, not the whole repo? @sttts had also done some experimenting with this I think wrt the ComponentConfig KEP in general

thockin:
Lucas, you should know by now that I never claim to be 100% self-consistent
:)

You can vendor just one package, but you have to download and examine the
whole repo.  Some tools used to not do subdir-vendoring, but I think we're
past that.

luxas:
> Lucas, you should know by now that I never claim to be 100% self-consistent

Hahah, yeah ;). Can't say I'm always super consistent with reviews either lol

> You can vendor just one package, but you have to download and examine the
whole repo.  Some tools used to not do subdir-vendoring, but I think we're
past that.

I don't think this will be a problem as anything should be able to depend on `k8s.io/utils`, this is used by multiple k8s consumers in various repos, the package has a well-defined scope, and all modern vendoring systems I know of supports pulling individual packages.

@thockin If this sounds good to you, go ahead and approve https://github.com/kubernetes/kubernetes/pull/66284 which follows this up in the k/k repo ;)

PR #43, opened 2018-07-27
add FloatPtr to pointer packages

stewart-yu:
1、add FloatPtr to pointer packages, 
2、fix #41 [comments](https://github.com/kubernetes/utils/pull/41#discussion_r205545516)
/cc @sttts  @apelisse

### DIFF
--- a/pointer/pointer.go
+++ b/pointer/pointer.go
@@ -74,3 +74,13 @@ func BoolPtr(b bool) *bool {
 func StringPtr(s string) *string {
 	return &s
 }
+
+// Float32Ptr returns a pointer to the passed float32.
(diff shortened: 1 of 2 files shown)


pointer/pointer.go
```diff
@@ -74,3 +74,13 @@ func BoolPtr(b bool) *bool {
 func StringPtr(s string) *string {
 	return &s
 }
+
+// Float32Ptr returns a pointer to the passed float32.
+func Float32Ptr(i float32) *float32 {
```
thockin: Do we have any users of float32?  In general we discourage it...

thockin: It looks like we do.  It's such a bad idea.

PR #49, opened 2018-10-11
Support os/exec.Cmd.Env

izumin5210:
`exec.Cmd` is missing `os/exec.Cmd.Env`

### DIFF
--- a/exec/exec.go
+++ b/exec/exec.go
@@ -60,6 +60,7 @@ type Cmd interface {
 	SetStdin(in io.Reader)
 	SetStdout(out io.Writer)
 	SetStderr(out io.Writer)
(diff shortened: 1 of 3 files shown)


exec/exec.go
```diff
@@ -121,6 +122,10 @@ func (cmd *cmdWrapper) SetStderr(out io.Writer) {
 	cmd.Stderr = out
 }
 
+func (cmd *cmdWrapper) SetEnv(env []string) {
```
thockin: Can you please add a simple test case?

izumin5210: I added 🙋‍♂️

thockin:
Thanks!

/lgtm
/approve

PR #50, opened 2018-10-16
copy k8s.io/kubernetes/pkg/util/io to k8s.io/utils/io

mcrute:
This change is a precondition of moving `k8s.io/kubernetes/pkg/util/mount` out of the kubernetes package. The whole set of work is related to kubernetes/kubernetes#69585. Once this is merged I'll update the kubernetes code to reference it and remove the package from `kubernetes/kubernetes`

/kind feature
/assign @lavalamp
/cc @andrewsykim

### DIFF
--- a/io/consistentread.go
+++ b/io/consistentread.go
@@ -1,0 +1,45 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


mcrute:
I've definitely signed the CLA, seems to have succeed on one of 3 PRs. Guess we'll try again.

/check-cla

mcrute:
Looks like the CLA check is failing on one of the merged commits. This will probably need overridden. The commits all come from `kubernetes/kubernetes` so they should have already been covered, right?

thockin:
Any thoughts on tests for this?  It's pretty trivial, but I hate untested code.

I'll approve since it's not making life any worse.  Ping me on the followup if you need.

mcrute:
@thockin Thanks! I'll add some tests and submit a PR for them tomorrow.

andrewsykim:
@mcrute also update README here https://github.com/kubernetes/utils#libraries :)

lavalamp:
Thanks for preserving history!

PR #52, opened 2018-10-16
copy k8s.io/kubernetes/pkg/util/nsenter to k8s.io/utils/nsenter

mcrute:
This change is a precondition of moving `k8s.io/kubernetes/pkg/util/mount` out of the kubernetes package. The whole set of work is related to kubernetes/kubernetes#69585. Once this is merged I'll update the kubernetes code to reference it and remove the package from `kubernetes/kubernetes`

/kind feature
/assign @lavalamp
/cc @andrewsykim

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,7 @@ script:
   - diff -u <(echo -n) <(gofmt -d .)
   - go tool vet .
   - go test -v -race ./...
+addons:
+  apt:
+    packages:
+      - coreutils
--- a/nsenter/OWNERS
+++ b/nsenter/OWNERS
@@ -1,0 +1,8 @@
+reviewers:
+  - jsafrane
+  - msau42
+  - cofyc
+approvers:
+  - jsafrane
+  - msau42
+  - cofyc
--- a/nsenter/exec.go
+++ b/nsenter/exec.go
@@ -1,0 +1,67 @@
+// +build linux
+
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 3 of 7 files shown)


mcrute:
I've definitely signed the CLA, seems to have succeed on one of 3 PRs. Guess we'll try again.

/check-cla

mcrute:
Looks like the CLA check is failing on one of the merged commits. This will probably need overridden. The commits all come from `kubernetes/kubernetes` so they should have already been covered, right?

thockin:
@msau42 Can you take a look and if happy I can manually merge.

mcrute:
Travis is running trusty which has a really old version of coreutils that does not contain a `realpath` binary which is required by the nsenter code. Please review the code that's here (migrated from kubernetes) and I'll add another commit to this PR to break the realpath dependency which should make the tests pass.

msau42:
Looks good to me.  There are a few mentions of "kubelet" here and there, but I'm not sure if it's worth the effort to go in and fix the comments and names to be more generic.

msau42:
cc @jsafrane

jsafrane:
lgtm

thockin:
/lgtm
/approve

manual merge coming up

thockin:
I think I jumped the gun.  Failing tests.

Should I revert this PR?

https://travis-ci.org/kubernetes/utils/jobs/442755085

mcrute:
@thockin I'm working on a fix for the tests that should land in PR later today. Feel free to revert it for now if you'd like. Nobody is using this code just yet and it works but it just doesn't test because of an old version of Ubuntu on Travis. Maybe you were thinking #51 instead of this one?

thockin:
maybe?  I'll hang tight and if you fix the tests today-ish we can avoid the revert :)

Thanks!

PR #53, opened 2018-10-17
Install realpath during testing in travis

mcrute:
The nsenter tests require an implementation of realpath but Ubuntu stripped it from coreutils in 8.20-2 because they were already shipping it in a different package. From Xenial onward Ubuntu started shipping realpath with the coreutils package instead of in a different package but Xenial is not yet supported on Travis.

Fixes issues in #52 

/kind bugfix
/assign @thockin

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: go
+dist: xenial
 go:
   - 1.8.x
   - 1.9.x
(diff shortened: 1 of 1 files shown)


thockin:
Should we just leave both there?

mcrute:
Let me rebase out the coreutils commit and then this should be good to go.

thockin:
ping me when ready

mcrute:
@thockin should be good to go. All tests passing and coreutils commit is removed.

PR #54, opened 2018-10-18
add tests for io and update README

mcrute:
These are followup items for #50. This is part of the work related to kubernetes/kubernetes#69585

/kind cleanup
/assign @thockin 
/cc @andrewsykim

### DIFF
--- a/io/consistentread_test.go
+++ b/io/consistentread_test.go
@@ -1,0 +1,94 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


io/consistentread_test.go
```diff
@@ -0,0 +1,84 @@
...
+			outstr = fmt.Sprintf("Foobar %d", i)
+		}
+
+		pipe.Write([]byte(outstr))
+		pipe.Close()
+
+		// Give the reader a moment to process the data and then write to the
+		// pipe again. Without this the pipe buffers all writes and the reader
+		// only gets the first read.
+		time.Sleep(10 * time.Millisecond)
```
thockin: Uggh.  This is guaranteed to fail eventually. 

What if we changed the `ConsistentRead` function to soemthing like:

```
{
    consistentReadSync(file, count, nil)
}
```

and define the new function like:

```
func consistentReadSync(filename string, attempts int, sync func()) ([]byte, error) {
    /// all the same as before
    /// but calls sync() after each read
}
```

And then the test would test the Sync function with a meaningful callback that tells it to write the next loop (e.g. unlock a mutex).  I don't LOVE mutating prod code for testing, but it's not against the rules, and timing tests always fail eventually

mcrute: Yeah I don't love the timing approach either. Here's an alternative that uses an optional sync callback and channels. WDYT?

thockin:
Nice, thanks.

/lgtm
/approve

PR #55, opened 2018-10-19
Move pkg util strings/keymutex from kubernetes kubernetes

dims:
- Move https://github.com/kubernetes/kubernetes/tree/master/pkg/util/strings to k8s.io/utils/strings
- Move https://github.com/kubernetes/kubernetes/tree/master/pkg/util/keymutex to k8s.io/utils/keymutex 

This is to help with the larger effort around breaking up cloud providers and moving them out of tree.

Fixes https://github.com/kubernetes/utils/issues/62


dims:
/assign @thockin 
cc @mcrute @andrewsykim

strings/strings.go
```diff
@@ -0,0 +1,58 @@
...
+// ShortenString returns the first N slice of a string.
+func ShortenString(str string, n int) string {
+	if len(str) <= n {
+		return str
+	}
+	return str[:n]
+}
+
+// isVowel returns true if the rune is a vowel (case insensitive).
+func isVowel(c rune) bool {
```
mcrute: It might be worth exporting this since it could be used to eliminate some [duplicate code in apiserver](https://github.com/kubernetes/kubernetes/search?utf8=%E2%9C%93&q=isVowel&type=)

keymutex/hashed.go
```diff
@@ -0,0 +1,64 @@
...
+		mutexes: make([]sync.Mutex, n),
+	}
+}
+
+type hashedKeyMutex struct {
+	mutexes []sync.Mutex
+}
+
+// Acquires a lock associated with the specified ID.
+func (km *hashedKeyMutex) LockKey(id string) {
```
mcrute: Does golint pass on this package? I thought it complained about comments not starting with function names. New code should be golint clean (he says, realizing now that he forgot to check for his PRs).

dims: travis seems to test at least gofmt and vet. though not go lint. if we want to do golint, we should start by adding it there there and not start on this PR

mcrute: That's fair. I need to do it with my merged code as well so I'll raise a PR with those changes independent of this one.

strings/strings.go
```diff
@@ -0,0 +1,58 @@
...
+// ShortenString returns the first N slice of a string.
+func ShortenString(str string, n int) string {
+	if len(str) <= n {
+		return str
+	}
+	return str[:n]
+}
+
+// IsVowel returns true if the rune is a vowel (case insensitive).
+func IsVowel(c rune) bool {
```
thockin: We have a test for this in staging/src/k8s.io/apiserver/pkg/endpoints/installer_test.go and that pkg seems to be the only user of this function.  Oh, and they have their own copy of this.  Can we nuke this?

strings/escape.go
```diff
@@ -0,0 +1,49 @@
...
+	return strings.Replace(in, "/", "~", -1)
+}
+
+// UnescapePluginName converts a plugin directory name in the format
+// vendor~pluginname into a proper vendor/pluginname.
+func UnescapePluginName(in string) string {
+	return strings.Replace(in, "~", "/", -1)
+}
+
+// EscapeQualifiedNameForDisk converts a plugin name, which might contain a / into a
```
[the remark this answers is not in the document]
dims: @thockin we enforce `vet` in travis CI, but we don't do `golint`. I've filed a PR for that in https://github.com/kubernetes/utils/pull/56

dims:
@thockin this is ready. we will have to bypass the CLA stuff though

thockin:
Why did we close this?

I went looking for it and found it wasn't merged

dims:
re-opened. was not sure what to do about the CLA stuff @thockin

pohly:
Is the problem with the CLA that for some of the authors there's no CLA on record?

Independently today, I proposed keymutex for inclusion into k8s.io/utils (issue #55), so I hope that this can be sorted out.

strings/escape.go
```diff
@@ -0,0 +1,49 @@
...
+package strings
+
+import (
+	"strings"
+)
+
+// EscapePluginName converts a plugin name in the format
+// vendor/pluginname into a proper ondisk vendor~pluginname plugin directory
+// format.
+func EscapePluginName(in string) string {
```
thockin: I *swear* I made this comment before, but can't find it.  The Escape/UnescapePluginName funcs are totally redundant.  Can we get rid of them?

thockin:
We can manually merge once comments are resolved

strings/escape.go
dims: Dropped! (Escape|Unescape)PluginName

dims:
@thockin this is ready, please take a look

thockin:
Manual merge because of CLA of multiple commits

PR #48, opened 2018-10-10
copy k8s.io/kubernetes/pkg/util/netsets to k8s.io/util/netsets

andrewsykim:
Once merged I will update k8s.io/utils vendor and replace k8s.io/kubernetes/pkg/util/net. 

/assign @lavalamp

PR #57, opened 2018-10-22
copy k8s.io/kubernetes/pkg/util/net to k8s.io/util/net

mcrute:
This change is a precondition to migrating cloud providers out of tree. The whole set of work is related to kubernetes/kubernetes#69585. Once this is merged I'll update the kubernetes code to reference it and remove the package from `kubernetes/kubernetes`. This was originally started in #48.

/kind feature
/assign @thockin 
/cc @andrewsykim


mcrute:
I'm in favor of flatting this out into `util/net`, I prefer less nesting in modules wherever possible. However, if that's the approach to take I think I'd rather migrate the entirety of `pkg/util/net` out of `k/k`, the parent package only adds a few more 5 more functions that round out the net interface. If you agree I'll update this PR with that as well as flatten out the packages.

netsets/doc.go
```diff
@@ -0,0 +1,28 @@
...
+//
+// We can't simply use net.IPNet as a map-key in Go (because it contains a
+// []byte).
+//
+// We could use the same workaround we use here (a string representation as the
+// key) to autogenerate sets.  If we do that, or decide on an alternate
+// approach, we should replace the implementations in this package with the
+// autogenerated versions.
+
+package sets
```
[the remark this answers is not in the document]
mcrute: Oops. I'll fix this once we agree on overall approach.

mcrute:
@thockin I'm going to push forward with migrating net and flattening out the package.

mcrute:
All commits except the last one were migrated out of kubernetes/kubernetes. And I've signed the CLA to cover the last one.

/override cla/linuxfoundation

net/ipnet.go
```diff
@@ -0,0 +1,121 @@
...
+
+package net
+
+import (
+	"net"
+	"strings"
+)
+
+// IPNet maps string to net.IPNet.
+type IPNet map[string]*net.IPNet
```
thockin: This is an ambiguous name.  It needs to either be `utils/net.IPNetSet` or `utils/net/sets.IPNet`

net/net.go
```diff
@@ -0,0 +1,59 @@
...
+
+// IsIPv6CIDR returns if cidr is IPv6.
+// This assumes cidr is a valid CIDR.
+func IsIPv6CIDR(cidr string) bool {
+	ip, _, _ := net.ParseCIDR(cidr)
+	return IsIPv6(ip)
+}
+
+// FilterIncorrectIPVersion filters out the incorrect IP version case from a slice of IP strings.
+func FilterIncorrectIPVersion(ipStrings []string, isIPv6Mode bool) ([]string, []string) {
```
thockin: This pair of functions is a weird API.  It seems to only be used in one place - can we just move it there (pkg/proxy/service.go) ?

mcrute: Agree on pushing those down. I've removed them from this PR.

net/ipnet.go
mcrute: Updated to `IPNetSet`

mcrute:
@thockin do you see anything else that needs updated here or is this ready to go?

net/net.go
```diff
@@ -0,0 +1,37 @@
...
+
+// IsIPv6String returns if ip is IPv6.
+func IsIPv6String(ip string) bool {
+	netIP := net.ParseIP(ip)
+	return IsIPv6(netIP)
+}
+
+// IsIPv6CIDR returns if cidr is IPv6.
+// This assumes cidr is a valid CIDR.
+func IsIPv6CIDR(cidr string) bool {
```
thockin: This should probably be `IsIPv6CIDRString()` with possibly a corresponding `IsIPv6CIDR` which takes a net.IPNet ?

thockin:
Just a small point, otherwise LGTM.  Thanks.   Let me know when ready for final

net/net.go
mcrute: Updated and rebased

mcrute:
@thockin bumping this, any other concerns before this is ready to merge?

mcrute:
@spiffxp I think that's just GitHub diffs failing, it looks okay in the [source branch](https://github.com/mcrute/utils/blob/netsets/net/OWNERS)

mcrute:
@spiffxp that might be expected actually since this is a graft from a different repo. It's extracted with all the files at the top level (which would clobber /OWNERS) but then gets [moved to a subfolder](https://github.com/kubernetes/utils/pull/57/commits/09ff3d42e90dce049815a70e07a55cf244e3e943) before it is grafted on. GitHub likely has some issues with that sequence.

mcrute:
@spiffxp no, I'm wrong. Rebased.

dims:
applied `cncf-cla: yes` label by hand

thockin:
Manual merge because of CLA of multiple commits

PR #58, opened 2018-11-01
add @andrewsykim @cheftako and @mcrute as reviewers

andrewsykim:
In the interest of https://github.com/kubernetes/kubernetes/issues/69585, some of us from SIG Cloud Provider would like to help maintain this repo starting as reviewers (and hopefully as approvers in the near future).

/assign @thockin @lavalamp

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -13,3 +13,6 @@ reviewers:
 - lavalamp
 - mengqiy
 - thockin
+- andrewsykim
+- cheftako
+- mcrute


lavalamp:
do you want to own everything? We can also start OWNERs files in subdirs.

andrewsykim:
@lavalamp personally I'm okay with reviewing everything given the repo is pretty small (for now). If we ever get around to defining approvers it might make sense to put ourselves in the subdirs based on the repos we're familiar with.

thockin:
Rock and roll!  My primary request is that we keep the bar HIGH for this repo.  Docs, comments, tests, APIs - this is the most "general purpose" repo we have, let's try to make sure that no k8s-isms sneak in here.

/lgtm
/approve

PR #59, opened 2018-11-07
Add Makefile and build scripts

mcrute:
Adds a `Makefile` and a few hack scripts that do the same things the `.travis.yml` is currently doing. The purpose of this change is to make it easier to test modifications to this package before submitting them.

/kind feature
/assign @thockin

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,37 @@
+# Copyright 2018 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+.PHONY: verify
+verify: depend verify-fmt verify-lint vet
+	go test -v -race ./...
+
+.PHONY: depend
+depend:
+	go get -t -v ./...
+
+.PHONY: verify-fmt
+verify-fmt:
+	./hack/verify-gofmt.sh
+
+.PHONY: verify-lint
+verify-lint:
+	./hack/verify-golint.sh
+
+.PHONY: vet
+vet:
+	go tool vet .
+
+.PHONY: update-fmt
+update-fmt:
+	gofmt -s -w .
--- a/hack/verify-gofmt.sh
+++ b/hack/verify-gofmt.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+
+# Copyright 2017 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+if ! which gofmt > /dev/null; then
+  echo "Can not find gofmt"
+  exit 1
+fi
+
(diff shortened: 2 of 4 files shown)


hack/update-gofmt.sh
```diff
@@ -0,0 +1,21 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+find . -name "*.go" | grep -v "\/vendor\/" | xargs gofmt -s -w
```
andrewsykim: Doesn't seem like we have vendor directory in this repo, do we still need this?

hack/verify-gofmt.sh
```diff
@@ -0,0 +1,32 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+if ! which gofmt > /dev/null; then
+  echo "Can not find gofmt"
+  exit 1
+fi
+
+diff=$(find . -name "*.go" | grep -v "\/vendor\/" | xargs gofmt -s -d 2>&1)
```
andrewsykim: same here, re: `grep -v vendor`

hack/update-gofmt.sh
mcrute: Not really, they're mostly there because I stole these scripts from another kubernetes project :smile:

Makefile
```diff
@@ -0,0 +1,26 @@
+.PHONY: release
+release: depend fmt lint vet test
```
andrewsykim: nit: I think a better target name here is `ci`.

hack/update-gofmt.sh
```diff
@@ -0,0 +1,21 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+find . -name "*.go" | xargs gofmt -s -w
```
andrewsykim: I believe `gofmt` is recursive by default, so just `gofmt -s -w .` might be sufficient here.

hack/verify-gofmt.sh
```diff
@@ -0,0 +1,32 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+if ! which gofmt > /dev/null; then
+  echo "Can not find gofmt"
+  exit 1
+fi
+
+diff=$(find . -name "*.go" | xargs gofmt -s -d 2>&1)
```
andrewsykim: same here with gofmt being recursive

Makefile
```diff
@@ -0,0 +1,26 @@
...
+
+.PHONY: depend
+depend:
+	go get -t -v ./...
```
thockin: Do we need to do this here?  What is it for?  It's hostile to the caller's GOPATH.

Makefile
thockin: or "verify" or "test"

Makefile
```diff
@@ -0,0 +1,26 @@
...
+
+.PHONY: fmt
+fmt:
```
thockin: verify-fmt
verify-lint
etc

Makefile
mcrute: There are several external dependencies in the package that need to be installed before testing. This is what the `.travis.yml` is currently doing. What would be the preferred approach, to use a local `.go` folder within the package itself? I'll need to make the same update in `verify-golint.sh` which attempts to `go get` golint if it's not already on the user's path.

Makefile
mcrute: I went with `verify` as that's what k8s calls it.

Makefile
andrewsykim: I ran a quick test, and seems like it was only to work around dependency to glog
```
$ go get -t -v ./...
github.com/golang/glog (download)
k8s.io/utils/clock
k8s.io/utils/io
github.com/golang/glog
k8s.io/utils/exec
k8s.io/utils/pointer
k8s.io/utils/clock/testing
k8s.io/utils/temp
k8s.io/utils/exec/testing
k8s.io/utils/temp/temptest
k8s.io/utils/nsenter
```

Makefile
```diff
@@ -0,0 +1,41 @@
...
+.PHONY: verify-lint
+verify-lint:
+	./hack/verify-golint.sh
+
+.PHONY: vet
+vet:
+	go tool vet .
+
+.PHONY: verify-uncached
+verify-uncached: depend verify-fmt verify-lint vet
```
thockin: call this "test" ?

PR #60, opened 2018-11-14
[exec] Implement Start, Wait & StdIOPipe methods

hoegaarden:
This change allows users to get a hold of the Std{out,err} streams of processes as `io.Reader`s. This allows e.g. to wrap those readers with a `io.limitedReader` and other standard (and non-standard) mechanisms for generic `io.Reader`s. This is currently not possible, as access to Std{out,err} is only possible via:
- handing in a `io.Writer`s
- inspecting the returned `[]byte`s from the `Output`/`CombinedOutput` methods

Specifically, we were looking into https://github.com/kubernetes/kubernetes/issues/70890 where we would jut employ an `io.limitedReader` to stop reading input after a certain amount of data.

/kind enhancement
/cc @mariantalla

### DIFF
--- a/exec/stdiopipe_test.go
+++ b/exec/stdiopipe_test.go
@@ -1,0 +1,55 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


exec/stdiopipe_test.go
```diff
@@ -0,0 +1,55 @@
...
+package exec_test
+
+import (
+	"fmt"
+	"io/ioutil"
+
+	"k8s.io/utils/exec"
+)
+
+func ExampleCmd_StderrPipe() {
```
thockin: What is this?  It isn't called anywhere,and it doesn't explain what it is for

apelisse: https://blog.golang.org/examples

hoegaarden: It seems the new example does not show up on godoc.org. I will check on that in a separate PR. However, the example still runs as a test case never the less, so it's not completely useless ;)

apelisse: Might be because it's a method of an interface?

PR #61, opened 2018-11-15
[exec] Fix godoc Example

mariantalla:
We changed the Example to document the `New` function, as this seemed to
be a better fitting place.
Using a lower-case suffix does not document a specific method of a type
but marks the example to be a generic example for that type; this also
seemed more fitting for our example.

Signed-off-by: Hannes Hörl <hhorl@pivotal.io>

### DIFF
--- a/exec/stdiopipe_test.go
+++ b/exec/stdiopipe_test.go
@@ -23,7 +23,7 @@ import (
 	"k8s.io/utils/exec"
 )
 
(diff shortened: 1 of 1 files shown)


apelisse:
Looks like it's [working](https://godoc.org/k8s.io/utils/exec#example-New--StderrPipe)! Thanks

PR #65, opened 2019-01-23
Add trace from k/k repo

dims:
per discussion in https://github.com/kubernetes/kubernetes/pull/73209


mcrute:
It would be nice to have some tests for this but since it's just an extraction...

/lgtm

mcrute:
[Looks](https://github.com/kubernetes/test-infra/blob/9f0ac86fd0f4b96f5ca913fb120d2bb6a7c65ff7/prow/plugins/override/override.go#L130) like only an admin can override. Ping @thockin @liggitt

wojtek-t:
Thanks @dims !

@krzysied - FYI

trace/trace.go
```diff
@@ -0,0 +1,96 @@
...
+	tracenum := rand.Int31()
+	endTime := time.Now()
+
+	totalTime := endTime.Sub(t.startTime)
+	buffer.WriteString(fmt.Sprintf("Trace[%d]: %q (started: %v) (total time: %v):\n", tracenum, t.name, t.startTime, totalTime))
+	lastStepTime := t.startTime
+	for _, step := range t.steps {
+		stepDuration := step.stepTime.Sub(lastStepTime)
+		if stepThreshold == 0 || stepDuration > stepThreshold || klog.V(4) {
+			buffer.WriteString(fmt.Sprintf("Trace[%d]: [%v] [%v] %v\n", tracenum, step.stepTime.Sub(t.startTime), stepDuration, step.msg))
```
thockin: @DirectXMan12 This is an "interesting" use of logs - how do we think we would convert this to logr-style?

DirectXMan12: it's not entirely clear to me why it's written to a buffer first, except perhaps performance.   If we remove that constraint, it looks fairly straightforward:

```go
log.Info("trace", "trace id", tracenum, "since start", step.stepTime.Sub(t.startTime), "step duration", stepDuration, "message", step.msg)
```

or better yet

```go
log := log.WithValues("trace id", tracenum)
log.Info("trace", "since start", step.stepTime.Sub(t.startTime), "step duration", stepDuration, "message", step.msg)
```

Individual traces can easily be reconstituted by filtering on `"id" == ${TRACE_ID}` or somesuch.

In the long term?  This is probably a good candidate for actual tracing ;-).

DirectXMan12: It's at least mildly unclear what the motivation for this package is though, except to reduce logging throughput though -- it seems like a decent amount of this could be replaced with "normal" logging plus structured filtering.

dims: @DirectXMan12 this is extremely useful especially in the performance scaling tests. A variation of this helped me find the `nil panic propagation` problem ( see https://github.com/kubernetes/kubernetes/pull/72820/files )

wojtek-t: Amount of logs is one example. If something is happening 100 millions of times but it's only 100 times it's relatively slow, I don't want to see 100 millions of logs - I want to have 100.

Yes - once we have proper tracing, we probably should at least consider replacing those.

PR #68, opened 2019-01-24
k8s.io/utils should use klog

andrewsykim:
Tried to update vendor in k8s.io/kubernetes and ran into error since k8s.io/utils imports glog still.

### DIFF
--- a/nsenter/exec.go
+++ b/nsenter/exec.go
@@ -23,7 +23,7 @@ import (
 	"fmt"
 	"path/filepath"
 
-	"github.com/golang/glog"
+	"k8s.io/klog"
 	"k8s.io/utils/exec"
 )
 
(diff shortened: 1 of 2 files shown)


thockin:
I will be contrarian - itils.should either not use logging at all, or in
the rare cases it REALLY need s to log it should take an interface as
argument.  E.g. logr.

andrewsykim:
@thockin agreed & thought it was weird that utils was using glog in the first place. Will try to follow up on this soon.

dims:
@thockin i was debating that too, but did not want to add `logr` as a new dependency here until we decide on stuff in klog. hence the limited change from glog->klog since we are doing that everywhere already

thockin:
I'll send a sketch-patch.  Maybe someone can carry it over the finish line.

thockin:
https://github.com/kubernetes/utils/pull/70

thockin:
I owe a debt to whomever can fold the two nsenter implementations into one.

PR #70, opened 2019-01-24
A sketch of converting to logr

thockin:
We have 2 other users - another nsenter file and trace.  I don't know
why we have 2 seemingly very similar nsenter executors - maybe we can
fix that?

If this seems reasonable I can check in a klogr implementation (just a
clone of glogr.

It would be nice to get rid of the default and not log at all unless the logger is specified.  Given that the logs are very high V anyway, it seems plausible.

### DIFF
--- a/nsenter/exec.go
+++ b/nsenter/exec.go
@@ -23,9 +23,7 @@ import (
 	"fmt"
 	"path/filepath"
 
-	"github.com/go-logr/klogr"
 	"github.com/go-logr/logr"
-
 	"k8s.io/utils/exec"
 )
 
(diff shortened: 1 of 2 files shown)


thockin:
The way trace uses logs is .. interesting.  Need to think about how it would convert.

thockin:
Also, we should rename Nsenter to NSEnter or something and use a type alias to preserve compat.

andrewsykim:
ah, I see it's pretty important for utils/trace so ignore what I said above haha

dims:
/hold

(need some stuff to get in before we can do this)

dims:
the `klogr` stuff is in! now we need an answer for how trace.go uses klog as @thockin mentioned above before we can proceed with this:
https://github.com/kubernetes/utils/blob/master/trace/trace.go#L71-L81

Issue #69, opened 2019-01-24
Need tests for trace package


dims:
please see comment in https://github.com/kubernetes/utils/pull/65#issuecomment-456990325

ashishranjan738:
I am working on it

ashishranjan738:
@dims can you assign me this issue i would like to work on it

dims:
@ashishranjan738 just go ahead, it's a fresh issue no one is working on it

dims:
see main kubernetes repository on how this code is used for inspiration for the test cases

ashishranjan738:
sure thanks

PR #72, opened 2019-01-25
Adds UT for trace package in utils

ashishranjan738:
Signed-off-by: Ashish Ranjan <ashishranjan738@gmail.com>

This PR adds UT for trace package in utils repository

fixes: #69

### DIFF
--- a/trace/trace_test.go
+++ b/trace/trace_test.go
@@ -1,0 +1,200 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package trace
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+	"time"
+
+	"k8s.io/klog"
+)
+
+func TestStep(t *testing.T) {
+	tests := []struct {
+		name          string
+		inputString   string
+		expectedTrace *Trace
+	}{
+		{
+			name:        "When string is empty",
+			inputString: "",
+			expectedTrace: &Trace{
+				steps: []traceStep{
+					{time.Now(), ""},
+				},
+			},
+		},
+		{
+			name:        "When string is not empty",
+			inputString: "test2",
+			expectedTrace: &Trace{
+				steps: []traceStep{
+					{time.Now(), "test2"},
+				},
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			sampleTrace := &Trace{}
+			sampleTrace.Step(tt.inputString)
+			if sampleTrace.steps[0].msg != tt.expectedTrace.steps[0].msg {
+				t.Errorf("Expected %v \n Got %v \n", tt.expectedTrace, sampleTrace)
+			}
+		})
+	}
+}
+
+func TestTotalTime(t *testing.T) {
+	test := struct {
+		name       string
+		inputTrace *Trace
+	}{
+		name: "Test with current system time",
+		inputTrace: &Trace{
+			startTime: time.Now(),
+		},
+	}
+
+	t.Run(test.name, func(t *testing.T) {
+		got := test.inputTrace.TotalTime()
+		if got == 0 {
+			t.Errorf("Expected total time 0, got %d \n", got)
+		}
+	})
+}
+
+func TestLog(t *testing.T) {
+	test := struct {
+		name             string
+		expectedMessages []string
+		sampleTrace      *Trace
+	}{
+		name: "Check the log dump with 3 msg",
+		expectedMessages: []string{
+			"msg1", "msg2", "msg3",
+		},
+		sampleTrace: &Trace{
+			name: "Sample Trace",
+			steps: []traceStep{
+				{time.Now(), "msg1"},
+				{time.Now(), "msg2"},
+				{time.Now(), "msg3"},
+			},
+		},
+	}
+
+	t.Run(test.name, func(t *testing.T) {
+		var buf bytes.Buffer
+		klog.SetOutput(&buf)
+		test.sampleTrace.Log()
+		for _, msg := range test.expectedMessages {
+			if !strings.Contains(buf.String(), msg) {
+				t.Errorf("\nMsg %q not found in log: \n%v\n", msg, buf.String())
+			}
+		}
+	})
+}
+
+func TestLogIfLong(t *testing.T) {
+	currentTime := time.Now()
+	type mutate struct {
+		delay time.Duration
+		msg   string
+	}
+
+	tests := []*struct {
+		name             string
+		expectedMessages []string
+		sampleTrace      *Trace
+		threshold        time.Duration
+		mutateInfo       []mutate // mutateInfo contains the information to mutate step's time to simulate multiple tests without waiting.
+
+	}{
+		{
+			name: "When threshold is 500 and msg 2 has highest share",
+			expectedMessages: []string{
+				"msg2",
+			},
+			mutateInfo: []mutate{
+				{10, "msg1"},
+				{1000, "msg2"},
+				{0, "msg3"},
+			},
+			threshold: 500,
+		},
+		{
+			name: "When threshold is 10 and msg 3 has highest share",
+			expectedMessages: []string{
+				"msg3",
+			},
+			mutateInfo: []mutate{
+				{0, "msg1"},
+				{0, "msg2"},
+				{50, "msg3"},
+			},
+			threshold: 10,
+		},
+		{
+			name: "When threshold is 0 and all msg have same share",
+			expectedMessages: []string{
+				"msg1", "msg2", "msg3",
+			},
+			mutateInfo: []mutate{
+				{0, "msg1"},
+				{0, "msg2"},
+				{0, "msg3"},
+			},
+			threshold: 0,
+		},
+		{
+			name:             "When threshold is 20 and all msg 1 has highest share",
+			expectedMessages: []string{},
+			mutateInfo: []mutate{
+				{10, "msg1"},
+				{0, "msg2"},
+				{0, "msg3"},
+			},
+			threshold: 20,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var buf bytes.Buffer
+			klog.SetOutput(&buf)
+
+			tt.sampleTrace = New("Test trace")
+
+			for index, mod := range tt.mutateInfo {
+				tt.sampleTrace.Step(mod.msg)
+				tt.sampleTrace.steps[index].stepTime = currentTime.Add(mod.delay)
+			}
+
+			tt.sampleTrace.LogIfLong(tt.threshold)
+
+			for _, msg := range tt.expectedMessages {
+				if msg != "" && !strings.Contains(buf.String(), msg) {
+					t.Errorf("Msg %q expected in trace log: \n%v\n", msg, buf.String())
+				}
+			}
+		})
+	}
+}


ashishranjan738:
@dims have a look in this once free

dims:
@ashishranjan738 
- `LogIfLong()` is not tested ... if we add say 3 steps and the 2nd step takes longer than the other two, then the 2nd step would be logged essentially 
- `Log()` needs to be tested too ... if there are n steps then the logs should have all the steps

trace/trace_test.go
```diff
@@ -0,0 +1,110 @@
...
+			},
+		},
+	}
+
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			sampleTrace.Step(tt.inputString)
+
+			if sampleTrace.steps == nil || len(sampleTrace.steps) != len(tt.expectedTrace.steps) || sampleTrace.steps[len(sampleTrace.steps)-1].msg != tt.expectedTrace.steps[len(tt.expectedTrace.steps)-1].msg {
+				t.Fatalf("Failed %s:\n Expected %v \n Got %v \n", name, tt.expectedTrace, sampleTrace)
```
andrewsykim: This can be `t.Errorf`, generally `t.Fatal` would be used if you want to fail the test early (i.e. some setup method needed for the test case errors).

trace/trace_test.go
```diff
@@ -0,0 +1,110 @@
...
+*/
+
+package trace
+
+import (
+	"testing"
+	"time"
+)
+
+func TestNew(t *testing.T) {
```
andrewsykim: I would argue this unit test is not really needed

trace/trace_test.go
```diff
@@ -0,0 +1,110 @@
...
+				},
+			},
+		},
+	}
+
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			sampleTrace.Step(tt.inputString)
+
+			if sampleTrace.steps == nil || len(sampleTrace.steps) != len(tt.expectedTrace.steps) || sampleTrace.steps[len(sampleTrace.steps)-1].msg != tt.expectedTrace.steps[len(tt.expectedTrace.steps)-1].msg {
```
andrewsykim: Seems like this can just be 

```
if !reflect.DeepEqual(sampleTrace, tt.expectedTrace.steps) {
   t.Errorf(...)
}
```

trace/trace_test.go
```diff
@@ -0,0 +1,110 @@
...
+			got := New(tt.inputString)
+			if got.name != tt.expectedTrace.name {
+				t.Fatalf("Failed %s:\n Expected %v \n Got %v \n", name, tt.expectedTrace, got)
+			}
+		})
+	}
+}
+
+func TestStep(t *testing.T) {
+	sampleTrace := &Trace{}
```
andrewsykim: I would create this variable within the test run `t.Run` otherwise it will be shared state across test cases.

trace/trace_test.go
ashishranjan738: cool

trace/trace_test.go
ashishranjan738: @andrewrynhard I initially tried the way you mentioned but it didn't work as the [stepTime](https://github.com/kubernetes/utils/blob/master/trace/trace.go#L52) of traceSteps contains the value of time.Now()  and this will difficult to predict from the caller function. So I believe its good idea to check for stuff that we have control in our hand like msg.

trace/trace_test.go
ashishranjan738: @andrewrynhard really nice catch !! So I did it for intentionally as the [`Step`](https://github.com/kubernetes/utils/blob/master/trace/trace.go#L47-L53) method actually appends the msg in the steps slice. So by keeping it global i will also able to check whether the length is also increasing when the method is called, as it is a negative test case. Trying to summarise, the result of `Step` method will contain a slice  of struct called `steps` and the last struct of the slice will contain the msg  with which it is called and the length of the steps will be increasing by 1 everytime the method is called so making these two things as base i tried to check whether the last struct contains the valid msg and length of the steps increases or not.

ashishranjan738:
> @ashishranjan738
> 
> * `LogIfLong()` is not tested ... if we add say 3 steps and the 2nd step takes longer than the other two, then the 2nd step would be logged essentially
> * `Log()` needs to be tested too ... if there are n steps then the logs should have all the steps

Added but I think we should not test these methods as these methods don't mutate any state they just logs some info. Ref : https://stackoverflow.com/questions/22606307/how-to-unit-test-a-function-that-does-not-return-anything

trace/trace_test.go
andrewsykim: I would at least break this into separate if statements. Easier to figure out where the diff is in that case.

trace/trace_test.go
andrewsykim: In that case I think the test should take a slice of inputs and run Step on all of them per test case. Sharing state across test cases can make tests really difficult to debug.

ashishranjan738:
@dims added the tests for Log and logic long in second commit

trace/trace_test.go
ashishranjan738: Okay will change the way as you requested thanks for the review

ashishranjan738: addressed :+1:

trace/trace_test.go
dims: +1 to remove it

trace/trace_test.go
```diff
@@ -0,0 +1,207 @@
...
+			for _, msg := range tt.expectedMessages {
+				if !strings.Contains(buf.String(), msg) {
+					t.Errorf("Test: %s \nMsg %q not found in log: \n%v\n", name, msg, buf.String())
+				}
+			}
+		})
+	}
+}
+
+func TestLogIfLong(t *testing.T) {
```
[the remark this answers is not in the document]
ashishranjan738: ok will try to address the same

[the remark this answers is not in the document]
ashishranjan738: @dims made the changes as told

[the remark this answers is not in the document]
ashishranjan738: done

trace/trace_test.go
ashishranjan738: done

trace/trace_test.go
ashishranjan738: removed the check to make it simple

trace/trace_test.go
```diff
@@ -0,0 +1,201 @@
...
+			},
+			mutateInfo: []mutate{
+				{0, "msg1"},
+				{0, "msg2"},
+				{0, "msg3"},
+			},
+			threshold: 0,
+		},
+		"Test 4 :": {
+			notExpectedMessages: []string{
```
dims: Can you please flip this around to expected messages?

ashishranjan738: yeah sure thanks for the review @dims

dims:
@ashishranjan738 thanks! please squash the commits and we can proceed

ashishranjan738:
@dims done

trace/trace_test.go
```diff
@@ -0,0 +1,199 @@
...
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+
+	"k8s.io/klog"
+)
+
+func TestStep(t *testing.T) {
+	tests := map[string]*struct {
```
andrewsykim: IMO, I think something like this makes the test table more readable:
```go
tests := []struct{
    name string
    inputString string
    expectedTrace *Trace
}{
      {
            name: "string is empty",
            inputString: "",
            expectedTrace: &Trace{....}
      } 
}

for _, test := range tests {
    t.Run(test.Name, t.Run(t *testing.T) {
         ....
    })
}
```

With maps it's not immediately clear that the key value is the description of the tests. What do you think?

trace/trace_test.go
```diff
@@ -0,0 +1,199 @@
...
+				startTime: time.Now(),
+			},
+		},
+	}
+
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			got := tt.inputTrace.TotalTime()
+			if got == 0 {
+				t.Errorf("Failed %s:\n Got %v \n", name, got)
```
andrewsykim: The name of the test case doesn't have to be included in the error message. It will be printed as part of `t.Run`. 
```
t.Errorf("expected total time of 0, got %d", got)
```

trace/trace_test.go
```diff
@@ -0,0 +1,199 @@
...
+			},
+		},
+	}
+
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			sampleTrace := &Trace{}
+			sampleTrace.Step(tt.inputString)
+			if sampleTrace.steps[len(sampleTrace.steps)-1].msg != tt.expectedTrace.steps[len(tt.expectedTrace.steps)-1].msg {
+				t.Errorf("Failed %s:\n Expected %v \n Got %v \n", name, tt.expectedTrace, sampleTrace)
```
andrewsykim: The name of the test case doesn't have to be included in the error message. It will be printed as part of t.Run.

trace/trace_test.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			var buf bytes.Buffer
+			klog.SetOutput(&buf)
+			tt.sampleTrace.Log()
+			for _, msg := range tt.expectedMessages {
+				if !strings.Contains(buf.String(), msg) {
+					t.Errorf("Test: %s \nMsg %q not found in log: \n%v\n", name, msg, buf.String())
```
andrewsykim: The name of the test case doesn't have to be included in the error message. It will be printed as part of t.Run.

trace/trace_test.go
```diff
@@ -0,0 +1,199 @@
...
+				tt.sampleTrace.steps[index].stepTime = currentTime.Add(mod.delay)
+			}
+
+			tt.sampleTrace.LogIfLong(tt.threshold)
+
+			fmt.Printf("%s\n\n", buf.String())
+			for _, msg := range tt.expectedMessages {
+				fmt.Printf(buf.String())
+				if msg != "" && !strings.Contains(buf.String(), msg) {
+					t.Errorf("Test: %s \nMsg %q expected in trace log: \n%v\n", name, msg, buf.String())
```
andrewsykim: The name of the test case doesn't have to be included in the error message. It will be printed as part of t.Run.

trace/trace_test.go
```diff
@@ -0,0 +1,199 @@
...
+			tt.sampleTrace = New("Test trace")
+
+			for index, mod := range tt.mutateInfo {
+				tt.sampleTrace.Step(mod.msg)
+				tt.sampleTrace.steps[index].stepTime = currentTime.Add(mod.delay)
+			}
+
+			tt.sampleTrace.LogIfLong(tt.threshold)
+
+			fmt.Printf("%s\n\n", buf.String())
```
andrewsykim: Can we remove this? Guessing it was left behind from debugging the test earlier

trace/trace_test.go
```diff
@@ -0,0 +1,199 @@
...
+			for _, msg := range tt.expectedMessages {
+				fmt.Printf(buf.String())
```
andrewsykim: Can we remove this? Guessing it was left behind from debugging the test earlier

trace/trace_test.go
ashishranjan738: yeah really thanks for pointing out

trace/trace_test.go
ashishranjan738: yeah really thanks for pointing it out

trace/trace_test.go
ashishranjan738: yup

ashishranjan738: updated

trace/trace_test.go
```diff
@@ -0,0 +1,199 @@
...
+			threshold: 20,
+		},
+	}
+
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			var buf bytes.Buffer
+			klog.SetOutput(&buf)
+
+			tt.sampleTrace = New("Test trace")
```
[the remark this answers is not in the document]
ashishranjan738: sure updated the test

trace/trace_test.go
ashishranjan738: sure updated it

trace/trace_test.go
ashishranjan738: nice we can do it

ashishranjan738: updated

trace/trace_test.go
```diff
@@ -0,0 +1,200 @@
...
+)
+
+func TestStep(t *testing.T) {
+	tests := []struct {
+		name          string
+		inputString   string
+		expectedTrace *Trace
+	}{
+		{
+			name:        "Test 1: When string is empty",
```
andrewsykim: nit: can we remove the "Test 1" and just use "when string is empty"?

trace/trace_test.go
```diff
@@ -0,0 +1,200 @@
...
+	tests := []*struct {
+		name             string
+		expectedMessages []string
+		sampleTrace      *Trace
+		threshold        time.Duration
+		mutateInfo       []mutate // mutateInfo contains the information to mutate step's time to simulate multiple tests without waiting.
+
+	}{
+		{
+			name: "Test 1",
```
andrewsykim: name here should be a little more descriptive of what's being tested

trace/trace_test.go
ashishranjan738: yeah done

trace/trace_test.go
ashishranjan738: addressed

trace/trace_test.go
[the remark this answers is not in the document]
andrewsykim: I still see sampleTrace as part of the test table. Instead we could do something like:
```
sampleTrace := New("test trace")
```

trace/trace_test.go
```diff
@@ -0,0 +1,200 @@
...
+	})
+}
+
+func TestLog(t *testing.T) {
+	test := struct {
+		name             string
+		expectedMessages []string
+		sampleTrace      *Trace
+	}{
+		name: "Test1",
```
andrewsykim: same comment around test names

trace/trace_test.go
```diff
@@ -0,0 +1,200 @@
...
+		})
+	}
+}
+
+func TestTotalTime(t *testing.T) {
+	test := struct {
+		name       string
+		inputTrace *Trace
+	}{
+		name: "Test 1: Test with current system time",
```
andrewsykim: same comment around test names

trace/trace_test.go
```diff
@@ -0,0 +1,200 @@
...
+				},
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			sampleTrace := &Trace{}
+			sampleTrace.Step(tt.inputString)
+			if sampleTrace.steps[len(sampleTrace.steps)-1].msg != tt.expectedTrace.steps[len(tt.expectedTrace.steps)-1].msg {
```
andrewsykim: This seems a bit awkward to me. If we know # of steps is always 1 for this test, let's just do:
```go
if sampleTrace.steps[0].msg != tt.expectedTrace.steps[0].msg {
```

andrewsykim:
Few more comments, lgtm otherwise :)

trace/trace_test.go
ashishranjan738: addressed

trace/trace_test.go
ashishranjan738: yeah I agree made the changes

trace/trace_test.go
ashishranjan738: addressed

PR #74, opened 2019-01-28
keymutex: remove glog usage

pohly:
https://github.com/kubernetes/utils/pull/55 was merged shortly after
https://github.com/kubernetes/utils/pull/68 and then brought back
glog, which now breaks vendoring of k8s.io/utils (can't have both glog
and klog defining the same command line flags).

Instead of replacing the logging with klog, logging gets removed
entirely. The rationale is that the log output is likely to be more
useful and/or readable if it is done by the caller. This is also a
first step towards removing the klog dependency from
k8s.io/utils (https://github.com/kubernetes/utils/pull/68#issuecomment-457249328).

/assign @dims

### DIFF
--- a/keymutex/hashed.go
+++ b/keymutex/hashed.go
@@ -20,8 +20,6 @@ import (
 	"hash/fnv"
 	"runtime"
 	"sync"
-
-	"github.com/golang/glog"
 )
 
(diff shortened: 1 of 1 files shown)


andrewsykim:
@dims @thockin blip in k8s bot, can you re-approve this?

PR #78, opened 2019-01-31
Fix #71: Use math.Round in integer.RoundToInt32

### LINKED ISSUE #71: integer.RoundToInt32 fails for 0.49999999999999994
If you add the following test to `integer/integer_test.go`:

```
{
  num: 0.49999999999999994,
  exp: 0,
}
```

And then run the tests, they fail:

```
go test ./integer
--- FAIL: TestRoundToInt32 (0.00s)
	integer_test.go:243: executing scenario 0
	integer_test.go:243: executing scenario 1
	integer_test.go:243: executing scenario 2
	integer_test.go:243: executing scenario 3
	integer_test.go:243: executing scenario 4
	integer_test.go:243: executing scenario 5
	integer_test.go:245: expected 0, got 1
	integer_test.go:243: executing scenario 6
FAIL
FAIL	github.com/kubernetes/utils/integer	0.006s
```

Is this beyond the scope of this function? If not I can submit a PR to fix this, though would it be acceptable to use `math.Round` provided in Go 1.10 and above?

matiasanaya:


matiasanaya:
I've just signed the CLA

matiasanaya:
`math.Round()` is available in `Go1.10` and above, hence the failing CI run on `Go1.9`. I can re-write this PR to be:

```
// RoundToInt32 rounds floats into integer numbers.
func RoundToInt32(a float64) int32 {
	t := math.Trunc(a)
	if math.Abs(a-t) >= 0.5 {
		return int32(t + math.Copysign(1, a))
	}
	return int32(t)
}
```

which works in `Go1.9`.

cc @thockin

thockin:
As far as I can tell this is used in ONE place in k/k.  What if we just got rid of this function entirely, since it is now trivial?

dims:
/hold

@matiasanaya can you please file a PR against k/k as Tim mentioned?

matiasanaya:
@dims, absolutely, though it'll have to wait 5 weeks since I'm traveling at the moment. Cheers!

PR #80, opened 2019-02-06
semantic: add forked reflect based DeepEqual

sttts:
We used to have this in apimachinery and in kube itself. The actual code is kube-independent.

We need this in kube-openapi. Before introducing another fork, we should move it here, for easy consumption for any party.

### DIFF
--- a/semantic/deep_equal.go
+++ b/semantic/deep_equal.go
@@ -1,0 +1,30 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package semantic
+
+import (
+	"k8s.io/utils/third_party/forked/golang/reflect"
+)
+
+// Equalities is a map from type to a function comparing two values of
+// that type.
+type Equalities = reflect.Equalities
+
+// EqualitiesOrDie adds the given funcs and panics on any error.
+func EqualitiesOrDie(funcs ...interface{}) Equalities {
+	return reflect.EqualitiesOrDie(funcs...)
+}
--- a/semantic/deep_equal_test.go
+++ b/semantic/deep_equal_test.go
@@ -1,0 +1,31 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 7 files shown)


dims:
- could we somehow denote what code changed?
- add separate tests for the updated/new behavior?

dims:
never mind this code was picked from k/k and staging ... let's add the third_party/forked in the directory structure please

```
[dims@dims-mac 07:31] ~/go/src/k8s.io/kubernetes ⟩ find . -name deep_equal*.go | xargs ls -altr
-rw-r--r--  1 dims  staff  1058 Nov 16 16:47 ./staging/src/k8s.io/apimachinery/pkg/conversion/deep_equal.go
-rw-r--r--  1 dims  staff  9789 Nov 16 16:47 ./staging/src/k8s.io/apimachinery/third_party/forked/golang/reflect/deep_equal.go
-rw-r--r--  1 dims  staff  3659 Nov 16 16:47 ./staging/src/k8s.io/apimachinery/third_party/forked/golang/reflect/deep_equal_test.go
-rw-r--r--  1 dims  staff  9789 Nov 16 16:47 ./third_party/forked/golang/reflect/deep_equal.go
-rw-r--r--  1 dims  staff  3659 Nov 16 16:47 ./third_party/forked/golang/reflect/deep_equal_test.go
```

dims:
/assign @thockin 

i agree that the code has diverged a lot from what it was in the go repo. Tim, WDYT about the directory names (do we need third_party/forked)?

thockin:
The third_party designation was to satisfy legal.  I would want a legal-ish opinion on this.  @dankohn is there an alias we can tag for consultation?

dankohn:
Yes, @swinslow is happy to help. My read is that it BSD-licensed code and so cannot be part of the main project, but it's fine to keep it in a third-party folder, as BSD is one of the whitelisted licenses (and the code has other users). But Steve can give the definitive thumbs up.

swinslow:
I'd recommend keeping it in a folder that is called "third_party/forked/" or similar, along the lines of what @dankohn described. Longer comments follow below:

The ideal long-term solution would be to upstream the changes where possible, particularly if they'd be usable for others upstream. That way the version used would be "unmodified" from Kubernetes' perspective, and we wouldn't be maintaining a separate fork. But I know there's no guarantee they'd be pulled in, or that we'd want to wait for that to happen.

The CNCF whitelist process (which we'll be posting more visibly in sig-release/licensing) automatically approves dependencies that are under certain non-Apache permissive licenses and meet some other criteria. One of those criteria is that the dependency be located in a third-party folder (e.g., dependencies in /vendor/) so it isn't intermingled with Apache-2.0 project code. Another criteria for the whitelist, though, is that it only applies where the dependency is unmodified. Since the dependency is being modified here, the forked version isn't automatically whitelisted.

However, having the forked dependency in a separate folder anyway _does_ help put people on notice that it might not be Apache-2.0 code. That's part of why the CNCF IP Committee has recommended approving license exceptions for these BSD-3-Clause forked components currently found in the "/third_party/forked/" directory. So I would recommend continuing this practice.

sttts:
I moved the code to `third_party/forked/golang/reflect`. Ptal.

dims:
/approve
/lgtm

Thanks @dankohn @swinslow @thockin

Issue #75, opened 2019-01-28
Move apimachinery/pkg/util/diff to utils


rajathagasthya:
We use that package in unit tests for some of our operators to generate diffs between actual and expected CR objects. If you think it's a good idea to add it to this repo, I'm happy to help!

thockin:
No objections, but note that the bar is higher in this rpo for code
quality, API, and testing.  You might get the opportunity to fix APis on
the way in.

rajathagasthya:
This is actually already implemented in #82. So I'll close this. Thanks.

PR #82, opened 2019-02-07
Move object diff and field path from apimachinery

sttts:
Both packages are not apimachinery specific, but totally generic (and e.g. useful in k/kube-openapi).

### DIFF
--- a/diff/diff.go
+++ b/diff/diff.go
@@ -1,0 +1,314 @@
+/*
+Copyright 2014 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package diff
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"sort"
+	"strings"
+	"text/tabwriter"
+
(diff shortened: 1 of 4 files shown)


diff/diff.go
```diff
@@ -0,0 +1,314 @@
...
+func ObjectGoPrintDiff(a, b interface{}) string {
+	s := spew.ConfigState{DisableMethods: true}
+	return StringDiff(
+		s.Sprintf("%#v", a),
+		s.Sprintf("%#v", b),
+	)
+}
+
+// ObjectReflectDiff returns a diff computed through reflection, without serializing to JSON.
+func ObjectReflectDiff(a, b interface{}) string {
```
liggitt: @smarterclayton will be happy to have this more available

liggitt:
both make sense to me for use outside k/k

diff/diff.go
rajathagasthya: So am I! This fixes my request in #75.

diff/diff_test.go
```diff
@@ -0,0 +1,148 @@
...
+		out  string
+	}{
+		"map": {
+			a: map[string]int{},
+			b: map[string]int{},
+		},
+		"detect nil map": {
+			a: map[string]int(nil),
+			b: map[string]int{},
+			out: `
```
thockin: this is a great place to use dedent and not have wacky indenting in string literals.

dims: @sttts do you want to switch to dedent in this PR or a follow up?

sttts: Would prefer a follow-up. This is a straight move.

sttts:
> The fact that the diff APIs expose their implementation details in their names means we can probably never change these once we publish them. Are we cool with that?

I have my doubts those details should be hidden. It makes a big difference how you extract a diff from objects. JSON is different from reflect.

thockin:
Perhaps the names should be more meaningfu, rather than
implementation-centric?  Or perhaps we should find one-true format that we
can use everywhere.

PR #83, opened 2019-02-08
semantic: document interface and hide implementation

sttts:
As a follow-up from @thockin's comments in https://github.com/kubernetes/utils/pull/81 about documentation of semantic deepequal, I propose this, hiding the forked impementation behind a well documentated interface.

**Note:** the third_party code introduced in the second commit is the one we have been running in k/k since 2015 (copied from current master).

**Note:** the original deepcopy version in the first commit is the current Golang tip. The original commit that added the code to Kubernetes already had the modified version. So we don't know when it was originally copied from Golang.


dims:
let's fix the travis failure @sttts

sttts:
@dims done

semantic/deep_equal.go
```diff
@@ -17,14 +17,48 @@ limitations under the License.
...
+	// DeepDerivative is similar to DeepEqual except that unset fields in a1 are
+	// ignored (not compared). This allows us to focus on the fields that matter to
+	// the semantic comparison.
+	//
+	// The unset fields include a nil pointer and an empty string.
+	DeepDerivative(a1, a2 interface{}) bool
+}
+
+// NewEquality creates a semantic equality object with the given equality funcs.
+func NewEquality(funcs ...interface{}) (Equality, error) {
```
sttts: for certain use-cases we also want to switch on/off the identification of nil and empty maps and slices. We would have a chance to add that now.

sttts: done

sttts:
/hold

some git rebase mess, fixing

semantic/internal/third_party/forked/golang/reflect/deepequal.go
```diff
@@ -2,56 +2,130 @@
...
-// Deep equality test via reflection
-
+// Package reflect is a fork of go's standard library reflection package, which
+// allows for deep equal with equality functions defined.
 package reflect
 
-import "unsafe"
+import (
+	"fmt"
+	"reflect"
```
thockin: should this be aliased?  I was very confused reading the diffs...

semantic/deep_equal.go
thockin: If we turn off the semantic parts isn't that just reflect.DeepEqual?  Or is this providing more differentiation than just that?

sttts:
@thockin does that cmd lib support funcs and semantic maps and slices?

sttts:
And most important: is it as fast? We use deepcopy at so many places. It is pretty critical.

sttts:
@thockin reading the github.com/google/go-cmp/cmp docs, it looks like we have to add an `Equal` method to customize behaviour. That's not what we can do in many cases.

semantic/internal/third_party/forked/golang/reflect/deepequal.go
sttts: fixed

liggitt:
> And most important: is it as fast? We use deepcopy at so many places. It is pretty critical.

benchmarked for the kube-openapi case at https://docs.google.com/spreadsheets/d/1hSSB7OKRsppC-yk4FTn9Ce0wthmS8OeGTwmab_Ugxe8/edit#gid=0

it's within 1% of both reflect.DeepEqual and apiequality.Semantic.DeepEqual in terms of allocations and runtime

> @thockin reading the github.com/google/go-cmp/cmp docs, it looks like we have to add an `Equal` method to customize behaviour. That's not what we can do in many cases.

It can use Equal methods defined on the objects themselves, or custom defined methods like this:

```
return cmp.Equal(a, b, cmp.Options{
  cmp.Comparer(func(a, b spec.Ref) bool { return a.String() == b.String() }),
})
```

liggitt:
that lib looks a lot cleaner, seems to fit the kube-openapi use case, and performance is comparable

dims:
/hold

looks like the other library may be a better candidate?

sttts:
We don't have the immediate kube-openapi need anymore. There `reflect.DeepCopy` now does the job.

thockin:
My concern here is the same it has always been wrt this utils/ repo.  Any package we move here is getting a promotion.  We're explicitly saying we are willing to support it for 3rd parties.  This should be the best of our efforts.  At the time we originally wrote this, go-cmp didn't exist.  Now it does.

If we want to promote this into utils, is it unreasonable to ask whether this is really the API and impl we want?  I don't want to be unreasonable, but this was not intended to just be a dumping ground, either.  It looks like there are not that many users of this, and that most/all could be better encapsulated.

PR #84, opened 2019-02-19
fix negative slice index error in keymutex

danielqsj:
xref: https://github.com/kubernetes/kubernetes/issues/73858

The keymutex's slice index could be negative which should not happened.
This PR aims to make the index always positive.

/assign @thockin @dims

### DIFF
--- a/keymutex/hashed.go
+++ b/keymutex/hashed.go
@@ -42,17 +42,17 @@ type hashedKeyMutex struct {
 
 // Acquires a lock associated with the specified ID.
 func (km *hashedKeyMutex) LockKey(id string) {
-	km.mutexes[km.hash(id)%len(km.mutexes)].Lock()
+	km.mutexes[km.hash(id)%uint32(len(km.mutexes))].Lock()
 }
 
 // Releases the lock associated with the specified ID.
 func (km *hashedKeyMutex) UnlockKey(id string) error {
-	km.mutexes[km.hash(id)%len(km.mutexes)].Unlock()
+	km.mutexes[km.hash(id)%uint32(len(km.mutexes))].Unlock()
 	return nil
 }
 
-func (km *hashedKeyMutex) hash(id string) int {
+func (km *hashedKeyMutex) hash(id string) uint32 {
 	h := fnv.New32a()
 	h.Write([]byte(id))
-	return int(h.Sum32())
+	return h.Sum32()
 }


danielqsj:
weird, why cla check failed ?

dims:
@danielqsj can you please sign the CLA? See https://github.com/kubernetes/community/blob/master/CLA.md for instructions and trouble shooting steps

danielqsj:
@dims I signed the CLA for CNCF a long time ago. And also checked again.
I'm also a member of kubernetes. So is there some issues about the check here because my PRs in k/k can pass CLA check.

dims:
@danielqsj you used a different email id :)
- https://patch-diff.githubusercontent.com/raw/kubernetes/kubernetes/pull/72336.patch
- https://patch-diff.githubusercontent.com/raw/kubernetes/utils/pull/84.patch

danielqsj:
@dims  oops, thank you so much.

dims:
@danielqsj is there some kind of test we could add?

keymutex/hashed.go
```diff
@@ -42,17 +42,17 @@ type hashedKeyMutex struct {
...
 
 // Releases the lock associated with the specified ID.
 func (km *hashedKeyMutex) UnlockKey(id string) error {
-	km.mutexes[km.hash(id)%len(km.mutexes)].Unlock()
+	km.mutexes[km.hash(id)%uint(len(km.mutexes))].Unlock()
 	return nil
 }
 
-func (km *hashedKeyMutex) hash(id string) int {
+func (km *hashedKeyMutex) hash(id string) uint {
```
apelisse: it's fine to cast from uint32 to uint, but I'm not sure why we're doing that.

keymutex/hashed.go
```diff
@@ -42,17 +42,17 @@ type hashedKeyMutex struct {
...
 // Acquires a lock associated with the specified ID.
 func (km *hashedKeyMutex) LockKey(id string) {
-	km.mutexes[km.hash(id)%len(km.mutexes)].Lock()
+	km.mutexes[km.hash(id)%uint(len(km.mutexes))].Lock()
 }
 
 // Releases the lock associated with the specified ID.
 func (km *hashedKeyMutex) UnlockKey(id string) error {
-	km.mutexes[km.hash(id)%len(km.mutexes)].Unlock()
+	km.mutexes[km.hash(id)%uint(len(km.mutexes))].Unlock()
```
apelisse: I also think it's fine to cast the return of `len()` (`int`) into `uint` since `len` is always returning a zero or positive value.

keymutex/hashed.go
danielqsj: hi @apelisse 
The [issue](https://github.com/kubernetes/kubernetes/issues/73858) mentions that on arm machine, the output of `int(h.Sum32())` maybe negative, and cause problems in `LockKey` and `UnlockKey` due to this negative number will be used as the slice index.

But I'm not sure, whether convert to `uint` is safe here.

Otherwise, `math.Abs` could be another option for this case ?

apelisse: Yeah, sorry I made myself unclear. I meant, why don't you just return an uncasted uint32?

danielqsj: aha, that's right. I miss the return type of `h.Sum32()`. Thanks.
Already updated, PTAL.

keymutex/hashed.go
```diff
@@ -42,17 +42,17 @@ type hashedKeyMutex struct {
...
 // Acquires a lock associated with the specified ID.
 func (km *hashedKeyMutex) LockKey(id string) {
-	km.mutexes[km.hash(id)%len(km.mutexes)].Lock()
+	km.mutexes[km.hash(id)%uint32(len(km.mutexes))].Lock()
 }
 
 // Releases the lock associated with the specified ID.
 func (km *hashedKeyMutex) UnlockKey(id string) error {
-	km.mutexes[km.hash(id)%len(km.mutexes)].Unlock()
+	km.mutexes[km.hash(id)%uint32(len(km.mutexes))].Unlock()
```
apelisse: And yes, this is fine.

PR #87, opened 2019-03-29
Merge interrupt history from Kubernetes into k8s.io/utils 

soltysh:
This is picking up the work from https://github.com/kubernetes/utils/pull/47 with comments from @thockin addressed. I've copied the last comments from Tim inline to continue discussion. I would like to merge that so we can continue with kubectl extraction from k/k (https://github.com/kubernetes/kubernetes/pull/75516)

/assign @thockin @dims @smarterclayton 
/cc @seans3

### DIFF
--- a/interrupt/interrupt.go
+++ b/interrupt/interrupt.go
@@ -44,11 +44,12 @@ func Chain(handler *Handler, notify ...func()) *Handler {
 	if handler == nil {
 		return New(nil, notify...)
 	}
-	return New(handler.Signal, append(notify, handler.Close)...)
+	return New(handler.signal, append(notify, handler.close)...)
 }
 
 // New creates a new handler that guarantees all notify functions are run after the critical
-// section exits (or is interrupted by the OS), then invokes the final handler. If no final
+// section exits (or is interrupted by the OS), then invokes the final handler. Notify
+// functions are invoked serially in the order passed to New method. If no final
 // handler is specified, the default final is `os.Exit(1)`. A handler can only be used for
 // one critical section. The final handler is given complete control over reacting to the
 // signal - it may exit the process or ignore the signal as needed. The following signals
@@ -60,8 +61,8 @@ func New(final func(os.Signal), notify ...func()) *Handler {
 	}
 }
 
-// Close executes all the notification handlers if they have not yet been executed.
-func (h *Handler) Close() {
+// close executes all the notification handlers if they have not yet been executed.
+func (h *Handler) close() {
 	h.once.Do(func() {
 		for _, fn := range h.notify {
 			fn()
@@ -69,10 +70,10 @@ func (h *Handler) Close() {
 	})
 }
 
-// Signal is called when an os.Signal is received, and guarantees that all notifications
+// signal is called when an os.Signal is received, and guarantees that all notifications
 // are executed, then the final handler is executed. This function should only be called once
 // per Handler instance.
-func (h *Handler) Signal(s os.Signal) {
+func (h *Handler) signal(s os.Signal) {
 	h.once.Do(func() {
 		for _, fn := range h.notify {
 			fn()
@@ -99,8 +100,8 @@ func (h *Handler) Run(fn func() error) error {
 		if !ok {
 			return
 		}
-		h.Signal(sig)
+		h.signal(sig)
 	}()
-	defer h.Close()
+	defer h.close()
 	return fn()
 }
--- a/interrupt/interrupt.go
+++ b/interrupt/interrupt.go
@@ -23,7 +23,7 @@ import (
 	"syscall"
 )
 
-// terminationSignals are signals that cause the program to exit in the
+// terminationSignals are the default signals that cause the program to exit in the
 // supported platforms (linux, darwin, windows).
 var terminationSignals = []os.Signal{syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT}
 
@@ -50,7 +50,9 @@ func Chain(handler *Handler, notify ...func()) *Handler {
 // New creates a new handler that guarantees all notify functions are run after the critical
 // section exits (or is interrupted by the OS), then invokes the final handler. If no final
 // handler is specified, the default final is `os.Exit(1)`. A handler can only be used for
-// one critical section.
+// one critical section. The final handler is given complete control over reacting to the
+// signal - it may exit the process or ignore the signal as needed. The following signals
+// are watched: SIGHUP, SIGINT, SIGTERM, and SIGQUIT.
 func New(final func(os.Signal), notify ...func()) *Handler {
 	return &Handler{
 		final:  final,
--- a/interrupt/doc.go
+++ b/interrupt/doc.go
@@ -17,5 +17,6 @@ limitations under the License.
 // Package interrupt provides a wrapper for handling signals from
 // the operating system that guarantee cleanup and exit semantics.
 // This allows a command to intercept CTRL+C or SIGTERM and do
-// the right thing before exiting.
+// the right thing before exiting. See `New()` for more information
+// about what signals are caught and how to handle them.
 package interrupt


interrupt/interrupt.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// New creates a new handler that guarantees all notify functions are run after the critical
+// section exits (or is interrupted by the OS), then invokes the final handler. Notify
+// functions are invoked serially in the order passed to New method. If no final
+// handler is specified, the default final is `os.Exit(1)`. A handler can only be used for
+// one critical section. The final handler is given complete control over reacting to the
+// signal - it may exit the process or ignore the signal as needed. The following signals
+// are watched: SIGHUP, SIGINT, SIGTERM, and SIGQUIT.
+func New(final func(os.Signal), notify ...func()) *Handler {
```
soltysh: > We should document the order and parallelism that notify functions are called. Presumably it is in-order and in-serial ?

Yes, it's in-order and serially, updated doc, see last commit.

interrupt/interrupt.go
```diff
@@ -0,0 +1,107 @@
...
+	notify []func()
+	final  func(os.Signal)
+	once   sync.Once
+}
+
+// Chain creates a new handler that invokes all notify functions when the critical section exits
+// and then invokes the optional handler's notifications. This allows critical sections to be
+// nested without losing exactly once invocations. Notify functions can invoke any cleanup needed
+// but should not exit (which is the responsibility of the parent handler).
+func Chain(handler *Handler, notify ...func()) *Handler {
```
soltysh: > Why is this not a method on Handler?

Chain creates a new handler, it's clearly stated in the description, so it shouldn't be a method on handler.

interrupt/interrupt.go
```diff
@@ -0,0 +1,107 @@
...
+type Handler struct {
+	notify []func()
+	final  func(os.Signal)
+	once   sync.Once
+}
+
+// Chain creates a new handler that invokes all notify functions when the critical section exits
+// and then invokes the optional handler's notifications. This allows critical sections to be
+// nested without losing exactly once invocations. Notify functions can invoke any cleanup needed
+// but should not exit (which is the responsibility of the parent handler).
```
soltysh: > If I read this right, the ordering on notify functions is a little weird. We should document it, at least.
> e.g.
> h := New(final, a, b, c)
> h = Chain(h, d, e, f)  
> What order is guaranteed?

The order as documented in the method's description is: first we invoked notify functions,
than handler's notifications. So in your example the order will be: `d, e, f` and then `a, b, c`.

interrupt/interrupt.go
```diff
@@ -0,0 +1,107 @@
...
+// signal - it may exit the process or ignore the signal as needed. The following signals
+// are watched: SIGHUP, SIGINT, SIGTERM, and SIGQUIT.
+func New(final func(os.Signal), notify ...func()) *Handler {
+	return &Handler{
+		final:  final,
+		notify: notify,
+	}
+}
+
+// close executes all the notification handlers if they have not yet been executed.
```
soltysh: > Should this call final? Or document how this is expected to be used, without that? I wouldn't expect this to call the handlers..

This is to ensure that the notifications do get called at least once,
based on the description from New, where we state `New creates a new handler that guarantees all notify functions are run after the critical section exits`

interrupt/interrupt.go
```diff
@@ -0,0 +1,107 @@
...
+
+// Chain creates a new handler that invokes all notify functions when the critical section exits
+// and then invokes the optional handler's notifications. This allows critical sections to be
+// nested without losing exactly once invocations. Notify functions can invoke any cleanup needed
+// but should not exit (which is the responsibility of the parent handler).
+func Chain(handler *Handler, notify ...func()) *Handler {
+	if handler == nil {
+		return New(nil, notify...)
+	}
+	return New(handler.signal, append(notify, handler.close)...)
```
soltysh: > Why do we pass both handler.Signal (which calls all functions and then final) and handler.Close (which calls all functions but NOT final) ? Shouldn't it be enough to do one?

The reason for that is to maintain the guarantees we state in `New` method description.

interrupt/interrupt.go
```diff
@@ -0,0 +1,107 @@
...
+		for _, fn := range h.notify {
+			fn()
+		}
+	})
+}
+
+// signal is called when an os.Signal is received, and guarantees that all notifications
+// are executed, then the final handler is executed. This function should only be called once
+// per Handler instance.
+func (h *Handler) signal(s os.Signal) {
```
soltysh: > does this need to be a public method? Less is more?

Yeah, you're right. Actually both `Signal` and `Close` can be made private. 
I've updated the code accordingly.

soltysh:
WRT CLA this comment https://github.com/kubernetes/utils/pull/47#issuecomment-420314061 applies still, I believe.

dims:
/approve

manually switched the CLA to `yes`

dims:
@smarterclayton @thockin one of you wanna look? LGTM reading all the history/comments

dims:
hmm cla bot does not like me switching things, will manually merge if i don't hear from Tim or Clayton today

interrupt/interrupt.go
thockin: consider:

```
h1 := New(myfinal, foo, bar)
h2 := Chain(h1, bat, qux)
h2.Run(something)
```

In the case of normal completion,  we call `h2.close()` which calls `bat()` and `qux()`, then `h1.close()` which calls `foo()` and `bar()`.  OK

In the case of a signal being delivered, we call `h2.signal()` which calls `bat()` and `qux()`, then `h1.close()` which calls `foo()` and `bar()`.  Then it calls `h1.signal()` which *DOES NOTHING* because the `once` has already been triggered.  So `myfinal() is never called`.

I think you need to pass `handler.final` not `handler.signal` ?

Additionally this does not cleanly handle sub-sections - is it supposed to ?  See trouble example at Run() below.

This whole Chain() API feels fraught to me.  Because of the `once`, calling Chain() is effectively mutating (or invalidating) the original handler.  Can we make that more obvious in the API?

Either say that Chain() "copies" the parent handler's contents (and do `append(notify, handler.notify...)`) or say that Chain() takes modifies the original handler (e.g. (`h := New(myfinal, foo, bar); h.Chain(bat, baz); h.Run()` or something.

interrupt/interrupt.go
```diff
@@ -0,0 +1,107 @@
...
+			os.Exit(1)
+		}
+		h.final(s)
+	})
+}
+
+// Run ensures that any notifications are invoked after the provided fn exits (even if the
+// process is interrupted by an OS termination signal). Notifications are only invoked once
+// per Handler instance, so calling Run more than once will not behave as the user expects.
+func (h *Handler) Run(fn func() error) error {
```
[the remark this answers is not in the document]
thockin: Here's a fully formed demonstration:

```
package main

import (
    "fmt"
    "os"
    "syscall"
    "time"
)

func main() {
    sig()
    nestClean()
    nestSignal()
}
                                                                                                                            
func sig() {
    fmt.Println("TEST SIGNAL")
    fmt.Println("open foo")
    fmt.Println("open bar")
    h1 := New(myfinal, foo, bar)
    h1.Run(func() error {
        raise(syscall.SIGTERM)
        time.Sleep(100 * time.Millisecond)
        return nil 
    })  
    fmt.Println("")
}
func nestClean() {
    fmt.Println("TEST NESTED w/o signal")
    fmt.Println("open foo")
    fmt.Println("open bar")
    h1 := New(myfinal, foo, bar)
    h1.Run(func() error {
        fmt.Println("open qux")
        h2 := Chain(h1, qux)
        return h2.Run(func() error { return nil })
    })  
    fmt.Println("")
}

func nestSignal() {
    fmt.Println("TEST NESTED w/ signal")
    fmt.Println("open foo")
    fmt.Println("open bar")
    h1 := New(myfinal, foo, bar)
    h1.Run(func() error {
        fmt.Println("open qux")
        h2 := Chain(h1, qux)
        return h2.Run(func() error { return raise(syscall.SIGTERM) })
    })  
    fmt.Println("")
}

func myfinal(s os.Signal) {
    fmt.Printf("caught signal %q\n", s)
}
func foo() {
    fmt.Println("close foo")
}
func bar() {
    fmt.Println("close bar")
}
func qux() {
    fmt.Println("close qux")
}

func raise(sig os.Signal) error {
    p, err := os.FindProcess(os.Getpid())
    if err != nil {
        return err
    }
    return p.Signal(sig)
}                                
```

interrupt/interrupt.go
thockin: a fully formed example.  Note that myfinal is never called in the chain case

```
package main

import (
    "fmt"
    "os"
    "syscall"
    "time"
)

func main() {
    sig()
    chain()
}

func sig() {
    fmt.Println("TEST SIGNAL")
    fmt.Println("open foo")
    fmt.Println("open bar")
    h1 := New(myfinal, foo, bar)
    h1.Run(func() error {
        raise(syscall.SIGTERM)
        time.Sleep(100 * time.Millisecond)
        return nil 
    })  
    fmt.Println("")
}

func chain() {
    fmt.Println("TEST CHAIN SIGNAL")
    fmt.Println("open foo")
    fmt.Println("open bar")
    h1 := New(myfinal, foo, bar)
    fmt.Println("open qux")                                                                                                 
    h2 := Chain(h1, qux)
    h2.Run(func() error {
        raise(syscall.SIGTERM)
        time.Sleep(100 * time.Millisecond)
        return nil 
    })  
    fmt.Println("")
}
func myfinal(s os.Signal) {
    fmt.Printf("caught signal %q\n", s)
}
func foo() {
    fmt.Println("close foo")
}
func bar() {
    fmt.Println("close bar")
}
func qux() {
    fmt.Println("close qux")
}

func raise(sig os.Signal) error {
    p, err := os.FindProcess(os.Getpid())
    if err != nil {
        return err
    }
    return p.Signal(sig)
}                                  
```

thockin:
This PR seems to be abandoned, but the problems I cited are real and present in the original, too, right?  I am going to file a new issue, so we don't lose it.

PR #88, opened 2019-04-29
Add a new func `ReadAtMost` and rename the file

haiyanmeng:
Add a new func `ReadAtMost` and rename the file to a more general name

I planned to add `ReadAtMost` into `k8s.io/pkg/util/system` (see https://github.com/kubernetes/kubernetes/pull/76518), and @tallclair  [suggested](https://github.com/kubernetes/kubernetes/pull/76518#discussion_r277087594) that `k8s.io/utils/io` is a btter place for this func.

### DIFF
--- a/io/read.go
+++ b/io/read.go
@@ -18,9 +18,14 @@ package io
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 )
+
+// ErrLimitReached means that the read limit is reached.
+var ErrLimitReached = errors.New("the read limit is reached")
 
 // ConsistentRead repeatedly reads a file until it gets the same content twice.
 // This is useful when reading files in /proc that are larger than page size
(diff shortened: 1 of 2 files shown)


io/read.go
```diff
@@ -53,3 +55,14 @@ func consistentReadSync(filename string, attempts int, sync func(int)) ([]byte,
...
 	return nil, fmt.Errorf("could not get consistent content of %s after %d attempts", filename, attempts)
 }
+
+// ReadAtMost reads up to `limit` bytes from `r`, and reports an error
+// when `limit` bytes are read.
+func ReadAtMost(r io.Reader, limit int64) ([]byte, error) {
+	limitedReader := &io.LimitedReader{R: r, N: limit}
+	data, err := ioutil.ReadAll(limitedReader)
+	if limitedReader.N <= 0 {
+		return data, errors.New("the read limit is reached")
```
tallclair: I think the error should be defined as an exported global - see https://github.com/kubernetes/kubernetes/pull/76518#discussion_r277083231

io/read.go
```diff
@@ -53,3 +55,14 @@ func consistentReadSync(filename string, attempts int, sync func(int)) ([]byte,
 	}
 	return nil, fmt.Errorf("could not get consistent content of %s after %d attempts", filename, attempts)
 }
+
+// ReadAtMost reads up to `limit` bytes from `r`, and reports an error
+// when `limit` bytes are read.
+func ReadAtMost(r io.Reader, limit int64) ([]byte, error) {
+	limitedReader := &io.LimitedReader{R: r, N: limit}
+	data, err := ioutil.ReadAll(limitedReader)
```
tallclair: I think you should check `err` before checking the read limit.

io/read.go
haiyanmeng: Fixed! Thanks!

io/read.go
haiyanmeng: Done!

PR #90, opened 2019-05-06
WIP: Add SinceInMicroseconds() and SinceInSeconds()

oomichi:
kubernetes/kubernetes repo has duplicated these methods in different
pathes. This patch adds these methods and we will replace them in k/k
with the additional methods in this repo.

### DIFF
--- a/clock/clock.go
+++ b/clock/clock.go
@@ -95,3 +95,13 @@ func (r *realTimer) Stop() bool {
 func (r *realTimer) Reset(d time.Duration) bool {
 	return r.timer.Reset(d)
 }
+
+// SinceInMicroseconds gets the time since the specified start in microseconds.
+func SinceInMicroseconds(start time.Time) float64 {
+	return float64(time.Since(start).Nanoseconds() / time.Microsecond.Nanoseconds())
+}
+
+// SinceInSeconds gets the time since the specified start in seconds.
+func SinceInSeconds(start time.Time) float64 {
+	return time.Since(start).Seconds()
+}


thockin:
I am torn on these.  On one hand, repeated code is a waste.  On the other hand so are dependencies.  These are really trivial, but I guess we added the pointer utils, so I have no leg to stand on.

thockin:
Where are the tests?

oomichi:
@thockin Thanks for your review.

> Where are the tests?

A nice point, I will add the corresponding unit tests for them.

oomichi:
I hoped we could move the corresponding unit test code from k/k because they are duplicated code and at least one should have test code. But actually there is not any unit test code, so we need to create test code at this time from scratch :-(

clock/clock.go
```diff
@@ -95,3 +95,13 @@ func (r *realTimer) Stop() bool {
 func (r *realTimer) Reset(d time.Duration) bool {
 	return r.timer.Reset(d)
 }
+
+// SinceInMicroseconds gets the time since the specified start in microseconds.
+func SinceInMicroseconds(start time.Time) float64 {
```
cheftako: This seems to be missing the point of this file. 
This file exists to give us a mechanism to allow us to inject a fake clock.
That fake clock can then be used to control the flow of time in tests to prevent tests from being flaky.
 
I have no issue with having SinceInUnit() methods but they should either be methods of the clock interface (so they can be overriden with the fake clock) or they should take the clock interface as a parameter (so that then call since on with that parameter and hence get the fake implementation when desired).

Some Examples of this include 
https://github.com/kubernetes/apiserver/blob/master/pkg/storage/cacher/watch_cache.go#L305
https://github.com/kubernetes/apiserver/blob/master/pkg/storage/cacher/watch_cache_test.go#L81

PR #93, opened 2019-05-23
Add Trace example and update documentation

misterikkit:
There are users of the trace library who call `Step()` at the
*beginning* of a step rather than the end, resulting in some confusing
logs. This updates examples and docs to clarify.

/kind cleanup
/kind documentation


misterikkit:
@deads2k @dashpole mind taking a look?

PR #94, opened 2019-06-05
add dual stack and multi cidrs funcs

khenidak:
This PR adds all the dual stack funcs and multi cidrs func that are to be used from `kubernetes/kubernetes` and related repos.

### DIFF
--- a/net/net.go
+++ b/net/net.go
@@ -16,7 +16,98 @@ limitations under the License.
 
 package net
 
-import "net"
+import (
+	"fmt"
+	"net"
+)
+
+// ParseCIDRs parses a list of cidrs and return error if any is invalid.
+// order is maintained
+func ParseCIDRs(cidrsString []string) ([]*net.IPNet, error) {
+	cidrs := make([]*net.IPNet, 0, len(cidrsString))
+	for _, cidrString := range cidrsString {
+		_, cidr, err := net.ParseCIDR(cidrString)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse cidr value:%q with error:%v", cidrString, err)
+		}
+		cidrs = append(cidrs, cidr)
+	}
+	return cidrs, nil
+}
+
+// IsDualStackIPs returns if a slice of ips is:
+// - all are valid ips
+// - at least one ip from each family (v4 or v6)
+func IsDualStackIPs(ips []net.IP) (bool, error) {
+	v4Found := false
+	v6Found := false
+	for _, ip := range ips {
+		if ip == nil {
+			return false, fmt.Errorf("ip %v is invalid", ip)
+		}
+
+		if v4Found && v6Found {
+			continue
+		}
+
+		if IsIPv6(ip) {
+			v6Found = true
+			continue
+		}
+
+		v4Found = true
+	}
+
+	return (v4Found && v6Found), nil
+}
+
+// IsDualStackIPStrings returns if
+// - all are valid ips
+// - at least one ip from each family (v4 or v6)
+func IsDualStackIPStrings(ips []string) (bool, error) {
+	parsedIPs := make([]net.IP, 0, len(ips))
+	for _, ip := range ips {
+		parsedIP := net.ParseIP(ip)
+		parsedIPs = append(parsedIPs, parsedIP)
+	}
+	return IsDualStackIPs(parsedIPs)
+}
+
+// IsDualStackCIDRs returns if
+// - all are valid cidrs
+// - at least one cidr from each family (v4 or v6)
+func IsDualStackCIDRs(cidrs []*net.IPNet) (bool, error) {
+	v4Found := false
+	v6Found := false
+	for _, cidr := range cidrs {
+		if cidr == nil {
+			return false, fmt.Errorf("cidr %v is invalid", cidr)
+		}
+
+		if v4Found && v6Found {
+			continue
+		}
+
+		if IsIPv6(cidr.IP) {
+			v6Found = true
+			continue
+		}
+		v4Found = true
+	}
+
+	return v4Found && v6Found, nil
+}
+
+// IsDualStackCIDRStrings returns if
+// - all are valid cidrs
+// - at least one cidr from each family (v4 or v6)
+func IsDualStackCIDRStrings(cidrs []string) (bool, error) {
+	parsedCIDRs, err := ParseCIDRs(cidrs)
+	if err != nil {
+		return false, err
+	}
+	return IsDualStackCIDRs(parsedCIDRs)
+}
 
 // IsIPv6 returns if netIP is IPv6.
 func IsIPv6(netIP net.IP) bool {
@@ -33,5 +124,11 @@ func IsIPv6String(ip string) bool {
 // This assumes cidr is a valid CIDR.
 func IsIPv6CIDRString(cidr string) bool {
 	ip, _, _ := net.ParseCIDR(cidr)
+	return IsIPv6(ip)
+}
+
+// IsIPv6CIDR returns if a cidr is ipv6
+func IsIPv6CIDR(cidr *net.IPNet) bool {
+	ip := cidr.IP
 	return IsIPv6(ip)
 }
--- a/net/net_test.go
+++ b/net/net_test.go
@@ -21,6 +21,254 @@ import (
 	"testing"
 )
 
+func TestParseCIDRs(t *testing.T) {
+	testCases := []struct {
+		cidrs         []string
+		errString     string
+		errorExpected bool
+	}{
+		{
+			cidrs:         []string{},
+			errString:     "should not return an error for an empty slice",
+			errorExpected: false,
+		},
+		{
+			cidrs:         []string{"10.0.0.0/8", "not-a-valid-cidr", "2000::/10"},
+			errString:     "should return error for bad cidr",
+			errorExpected: true,
+		},
+		{
+			cidrs:         []string{"10.0.0.0/8", "2000::/10"},
+			errString:     "should not return error for good  cidrs",
+			errorExpected: false,
+		},
+	}
+
+	for _, tc := range testCases {
+		cidrs, err := ParseCIDRs(tc.cidrs)
+		if tc.errorExpected {
+			if err == nil {
+				t.Errorf("%v", tc.errString)
+			}
+			continue
+		}
+		if err != nil {
+			t.Errorf("%v error:%v", tc.errString, err)
+		}
+
+		// validate lengthes
+		if len(cidrs) != len(tc.cidrs) {
+			t.Errorf("cidrs should be of the same lengthes %v != %v", len(cidrs), len(tc.cidrs))
+		}
+
+	}
+}
+func TestDualStackIPs(t *testing.T) {
+	testCases := []struct {
+		ips            []string
+		errMessage     string
+		expectedResult bool
+		expectError    bool
+	}{
+		{
+			ips:            []string{"1.1.1.1"},
+			errMessage:     "should fail because length is not at least 2",
+			expectedResult: false,
+			expectError:    false,
+		},
+		{
+			ips:            []string{},
+			errMessage:     "should fail because length is not at least 2",
+			expectedResult: false,
+			expectError:    false,
+		},
+		{
+			ips:            []string{"1.1.1.1", "2.2.2.2", "3.3.3.3"},
+			errMessage:     "should fail because all are v4",
+			expectedResult: false,
+			expectError:    false,
+		},
+		{
+			ips:            []string{"fd92:20ba:ca:34f7:ffff:ffff:ffff:ffff", "fd92:20ba:ca:34f7:ffff:ffff:ffff:fff0", "fd92:20ba:ca:34f7:ffff:ffff:ffff:fff1"},
+			errMessage:     "should fail because all are v6",
+			expectedResult: false,
+			expectError:    false,
+		},
+		{
+			ips:            []string{"1.1.1.1", "not-a-valid-ip"},
+			errMessage:     "should fail because 2nd ip is invalid",
+			expectedResult: false,
+			expectError:    true,
+		},
+		{
+			ips:            []string{"not-a-valid-ip", "fd92:20ba:ca:34f7:ffff:ffff:ffff:ffff"},
+			errMessage:     "should fail because 1st ip is invalid",
+			expectedResult: false,
+			expectError:    true,
+		},
+		{
+			ips:            []string{"1.1.1.1", "fd92:20ba:ca:34f7:ffff:ffff:ffff:ffff"},
+			errMessage:     "expected success, but found failure",
+			expectedResult: true,
+			expectError:    false,
+		},
+		{
+			ips:            []string{"fd92:20ba:ca:34f7:ffff:ffff:ffff:ffff", "1.1.1.1", "fd92:20ba:ca:34f7:ffff:ffff:ffff:fff0"},
+			errMessage:     "expected success, but found failure",
+			expectedResult: true,
+			expectError:    false,
+		},
+		{
+			ips:            []string{"1.1.1.1", "fd92:20ba:ca:34f7:ffff:ffff:ffff:ffff", "10.0.0.0"},
+			errMessage:     "expected success, but found failure",
+			expectedResult: true,
+			expectError:    false,
+		},
+		{
+			ips:            []string{"fd92:20ba:ca:34f7:ffff:ffff:ffff:ffff", "1.1.1.1"},
+			errMessage:     "expected success, but found failure",
+			expectedResult: true,
+			expectError:    false,
+		},
+	}
+	// for each test case, test the regular func and the string func
+	for _, tc := range testCases {
+		dualStack, err := IsDualStackIPStrings(tc.ips)
+		if err == nil && tc.expectError {
+			t.Errorf("%s", tc.errMessage)
+			continue
+		}
+		if err != nil && !tc.expectError {
+			t.Errorf("failed to run test case for %v, error: %v", tc.ips, err)
+			continue
+		}
+		if dualStack != tc.expectedResult {
+			t.Errorf("%v for %v", tc.errMessage, tc.ips)
+		}
+	}
+
+	for _, tc := range testCases {
+		ips := make([]net.IP, 0, len(tc.ips))
+		for _, ip := range tc.ips {
+			parsedIP := net.ParseIP(ip)
+			ips = append(ips, parsedIP)
+		}
+		dualStack, err := IsDualStackIPs(ips)
+		if err == nil && tc.expectError {
+			t.Errorf("%s", tc.errMessage)
+			continue
+		}
+		if err != nil && !tc.expectError {
+			t.Errorf("failed to run test case for %v, error: %v", tc.ips, err)
+			continue
+		}
+		if dualStack != tc.expectedResult {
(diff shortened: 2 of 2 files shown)


khenidak:
@thockin @andrewsykim @dims please help review

net/net.go
```diff
@@ -16,7 +16,72 @@ limitations under the License.
...
+import (
+	"fmt"
+	"net"
+)
+
+// ParseCIDRs parses a list of cidrs and return error if any is in valid.
+// order is maintained
+func ParseCIDRs(cidrsString []string) ([]*net.IPNet, error) {
+	cidrs := make([]*net.IPNet, 0)
+	for idx, current := range cidrsString {
```
andrewsykim: nit: can `idx` just be `i`, I feel that `i` is pretty well known for index (in Go at least).

net/net.go
```diff
@@ -16,7 +16,72 @@ limitations under the License.
...
+		_, cidr, err := net.ParseCIDR(current)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse cidr at index:%v with error:%v", idx, err)
```
andrewsykim: I'm not sure if the index is relevant here? Should just be:
```
return nil, fmt.Errorf("error parsing CIDR %q, err: %v", current, err)
```

net/net.go
andrewsykim: nit: current -> cidrString (or cidrStr)

net/net.go
```diff
@@ -16,7 +16,72 @@ limitations under the License.
...
-import "net"
+import (
+	"fmt"
+	"net"
+)
+
+// ParseCIDRs parses a list of cidrs and return error if any is in valid.
+// order is maintained
+func ParseCIDRs(cidrsString []string) ([]*net.IPNet, error) {
+	cidrs := make([]*net.IPNet, 0)
```
thockin: Why not allocate the Len(cidrStrings) size

net/net.go
```diff
@@ -16,7 +16,72 @@ limitations under the License.
...
 
-import "net"
+import (
+	"fmt"
+	"net"
+)
+
+// ParseCIDRs parses a list of cidrs and return error if any is in valid.
+// order is maintained
+func ParseCIDRs(cidrsString []string) ([]*net.IPNet, error) {
```
thockin: cidrStrings

net/net.go
```diff
@@ -16,7 +16,72 @@ limitations under the License.
...
+			return nil, fmt.Errorf("failed to parse cidr at index:%v with error:%v", idx, err)
+		}
+		cidrs = append(cidrs, cidr)
+	}
+
+	return cidrs, nil
+}
+
+// IsDualStack returns if ips are <v4>,<v6> or <v6>,<v4>
+func IsDualStackIPs(ips []net.IP) bool {
```
thockin: Would it make sense to have this as 'IPsByFamily' and return a map of IP family to slice of IP?  That seems like a primitive we could reuse?

thockin: Ditto for string beliw

thockin: And for CIDR and CIDR string

net/net_test.go
```diff
@@ -21,6 +21,185 @@ import (
...
+func TestParseCIDRs(t *testing.T) {
+	testCases := []struct {
+		cidrs         []string
+		errString     string
+		errorExpected bool
+	}{
+		{
+			cidrs:         []string{},
+			errString:     "should not return an error for an empty slice",
+			errorExpected: false,
```
andrewsykim: Remove `errorExpected` and just check the returned err is nil?

net/net_test.go
```diff
@@ -21,6 +21,185 @@ import (
...
+		{
+			cidrs:         []string{"10.0.0.0/8", "2000::/10"},
+			errString:     "should not return error for good  cidrs",
+			errorExpected: false,
+		},
+	}
+
+	for _, tc := range testCases {
+		cidrs, err := ParseCIDRs(tc.cidrs)
+		if tc.errorExpected {
```
andrewsykim: If you make errString an actual error and get rid of `errorExpected` you can do something like this instead:
```go
if reflect.DeepEqual(tc.expectedErr, err) {
   t.Errorf("...")
}

if tc.expectedErr != nil {
     return
}

if len(cidrs) != len(tc.cidrs) {
    t.Errorf(...)
}
```

net/net_test.go
```diff
@@ -21,6 +21,185 @@ import (
...
+			// validate lengthes
+			if len(cidrs) != len(tc.cidrs) {
+				t.Errorf("cidrs should be of the same lengthes %v != %v", len(cidrs), len(tc.cidrs))
+			}
+		}
+	}
+}
+func TestDualStackIPs(t *testing.T) {
+	testCases := []struct {
+		ips       []string
```
andrewsykim: Thoughts on adding a `name` or `desc` field to each test table and wrap with `t.Run`?

net/net.go
khenidak: The map[<family>]<cidr> has no use in k/k code base as far as i can tell. 
`ipam controller`: assign from cidr, no dependency on family (does depend on order of cidrs) 
`route controller`: creates routes on cidrs, no dependency on family
`kubenet`: only care about cidr ranges, only cares about family when it does traffic shaping.

net/net_test.go
khenidak: there are cases where i error is ok, testing bad data.

andrewsykim: can work around that by validating the expected error first though

net/net_test.go
khenidak: I honestly don't think we should test error messages. We can test error types.. but we are not using them here.

net/net_test.go
khenidak: The old tests were using this format.. I didn't want to have some using a testing approach and others using different one. Maybe later i can go and change the entire test file..

net/net_test.go
andrewsykim: Oh I see now.. errString is not the returned error from ParseCIDRsStrings but the error representing how the test would fail. I think that's sort of confusing, seems like that should just be in the test name/description

net/net.go
```diff
@@ -16,7 +16,71 @@ limitations under the License.
...
 
-import "net"
+import (
+	"fmt"
+	"net"
+)
+
+// ParseCIDRsStrings parses a list of cidrs and return error if any is in valid.
+// order is maintained
+func ParseCIDRsStrings(cidrsString []string) ([]*net.IPNet, error) {
```
andrewsykim: Thoughts on `ParseStringCIDRs` instead?

net/net.go
```diff
@@ -16,7 +16,71 @@ limitations under the License.
...
+		return false
+	}
+	if ips[0] == nil || ips[1] == nil {
+		return false
+	}
+	return IsIPv6(ips[0]) != IsIPv6(ips[1])
+}
+
+// IsDualStackIPsStrings returns if ips are (string) are <v4>,<v6> or <v6>,<v4>
+func IsDualStackIPsStrings(ips []string) bool {
```
andrewsykim: Thoughts on `IsDualStackStringIPs`?

net/net.go
```diff
@@ -16,7 +16,71 @@ limitations under the License.
...
+		return false
+	}
+	if cidrs[0] == nil || cidrs[1] == nil {
+		return false
+	}
+	return IsIPv6CIDR(cidrs[0]) != IsIPv6CIDR(cidrs[1])
+}
+
+// IsDualStackCIDRsStrings returns if cidrs(string) are <v4>,<v6> or <v6>,<v4>
+func IsDualStackCIDRsStrings(cidrs []string) bool {
```
andrewsykim: Thoughts on `IsDualStackStringCIDRs`?

net/net.go
```diff
@@ -16,7 +16,71 @@ limitations under the License.
...
-import "net"
+import (
+	"fmt"
+	"net"
+)
+
+// ParseCIDRsStrings parses a list of cidrs and return error if any is in valid.
+// order is maintained
+func ParseCIDRsStrings(cidrsString []string) ([]*net.IPNet, error) {
+	cidrs := make([]*net.IPNet, len(cidrsString))
```
bowei: It is preferred to not do this unless it's really a perf issue, so we don't have to deal with indexing:
```
var cidrs []*net.IPNet
for _, s := range cidrsString {
  ...
  cidrs = append(cidrs, cidr)
}
```

net/net.go
```diff
@@ -16,7 +16,71 @@ limitations under the License.
...
+
+// IsDualStackIPsStrings returns if ips are (string) are <v4>,<v6> or <v6>,<v4>
+func IsDualStackIPsStrings(ips []string) bool {
+	if len(ips) != 2 {
+		return false
+	}
+	ip0 := net.ParseIP(ips[0])
+	ip1 := net.ParseIP(ips[1])
+
+	if ip0 == nil || ip1 == nil {
```
bowei: isn't this redundant given what IsDualStackIPs does?

net/net_test.go
```diff
@@ -21,6 +21,185 @@ import (
 	"testing"
 )
 
+func TestParseCIDRs(t *testing.T) {
+	testCases := []struct {
```
bowei: t.Parallel() (add for other test cases as well)

net/net_test.go
```diff
@@ -21,6 +21,185 @@ import (
...
+			errorExpected: false,
+		},
+	}
+
+	for _, tc := range testCases {
+		cidrs, err := ParseCIDRsStrings(tc.cidrs)
+		if tc.errorExpected {
+			if err == nil {
+				t.Errorf("%v", tc.errString)
+			}
```
bowei: `continue` and then you don't have to use else and indent the rest of the block (fix in other tests as well)

net/net.go
khenidak: it is a bit. But i'd rather have the check done instead of `IsDualStackIPs` changes and we forget to update ..

net/net.go
```diff
@@ -16,7 +16,128 @@ limitations under the License.
...
+	return cidrs, nil
+}
+
+// IsDualStackIPs returns if a slice of ips is:
+// - all are valid ips
+// - at least one ip from each family (v4 or v6)
+// - an error is returned if length is less than 2
+// - an error is returned if duplicate ips are found
+func IsDualStackIPs(ips []net.IP) (bool, error) {
+	if len(ips) < 2 {
```
thockin: Does this need to be an error?  A slice of one element is not dual-stack, why is it an error?

net/net.go
```diff
@@ -16,7 +16,128 @@ limitations under the License.
 
 package net
 
-import "net"
+import (
+	"fmt"
+	"net"
+)
+
+// ParseCIDRs parses a list of cidrs and return error if any is in valid.
```
thockin: nit: "in valid" -> "invalid"

net/net.go
```diff
@@ -16,7 +16,128 @@ limitations under the License.
...
+		cidrs = append(cidrs, cidr)
+	}
+	return cidrs, nil
+}
+
+// IsDualStackIPs returns if a slice of ips is:
+// - all are valid ips
+// - at least one ip from each family (v4 or v6)
+// - an error is returned if length is less than 2
+// - an error is returned if duplicate ips are found
```
thockin: This is low cohesion - duplicateness has nothing to do with dual-stack-ness.

I'm already not super happy with this API as a general-purpose interface, can we make the duplicate-checking a separate function?

net/net.go
```diff
@@ -16,7 +16,128 @@ limitations under the License.
...
+func IsDualStackIPs(ips []net.IP) (bool, error) {
+	if len(ips) < 2 {
+		return false, fmt.Errorf("ips length is less than 2")
+	}
+
+	v4Found := false
+	v6Found := false
+	seen := make(map[string]int)
```
thockin: if you make this map[string]bool you can simply check seen[ip] - the default return for not-found will be false.

net/net.go
thockin: what I am trying to get at is that it feels like there are primitive operations that are plausibly general-purpose and can be composed by caller code into whatever combinations they need.  I'm not going to belabor the overall point because it doesn't matter THAT much, but I envisioned something like:

```
var ips []string // gets initialized somehow.
byFamily := utilnet.IPStringsByFamily(ip)
for k, v := range byFamily {
  dd, err := utilnet.DedupIPs(v)
  if err != nil { ... }
  if len(dd) != len(v) { ... }
}
```

net/net.go
```diff
@@ -16,7 +16,128 @@ limitations under the License.
...
+
+	return (v4Found && v6Found), nil
+}
+
+// IsDualStackIPsStrings returns if
+// - all are valid ips
+// - at least one ip from each family (v4 or v6)
+// - an error is returned if length is less than 2
+// - an error is returned if duplicate ips are found
+func IsDualStackIPsStrings(ips []string) (bool, error) {
```
thockin: Do we need the full set of {IP, CIDR} x {<t>, <t>String}  ?  Can we force callers to canonicalize?

net/net_test.go
```diff
@@ -21,6 +21,268 @@ import (
...
+		if err == nil && tc.expectError {
+			t.Errorf("%s", tc.errMessage)
+			continue
+		}
+		if err != nil && !tc.expectError {
+			t.Errorf("failed to run test case for %v, error: %v", tc.ips, err)
+			continue
+		}
+		if dualStack != tc.expectedResult {
+			t.Errorf("%v for %v", tc.errMessage, tc.ips)
```
thockin: good form to print the index, too, e.g `[%d]` prefix

net/net.go
khenidak: if we treated `[validv4, validv6, invalid]` as an error case, then we should treat `[validv4]` as an error case. Since we always look at the entire slice, no?

net/net_test.go
khenidak: we are printing the entire value that generated the error..

net/net.go
khenidak: same can be said for int. in both cases map value is not used. Because verbosity matter (same perf in both cases). validating if a key exist is has no gotchas and easier to read.

net/net.go
khenidak: I am good with adding a `dedup` general purpose function. But i would argue that `duplicate` cidr or IPs is an error case. checking for duplicates makes this `unit` the smallest possible unit. calling style can be:
```
ips := getIPsSomeHow()
deDupped := dedupIPs() // currently not provided by the util packege
// at this stage ips are valid
if IsDualStackIPs(..){

}
```

as a caller, i know that `IsDualStackIPs` will return true `if and only if` it is dualstack, there is no `true even if it has duplicate` or `true even if has nils`

net/net.go
khenidak: We don't need the full set, forcing the caller to canonicalize would add more pre-call code that the func does really need to operate correctly.

net/net_test.go
thockin: But that doesn't help me fins which testcase it was.  Minor point, I just have debugged enough of these where the error string didn't actually show me which case of the 30 cases was failing.  NOt super critical unless making other changes

net/net.go
thockin: I don't know the callsites, what I means here is that if they all eventually convert `[]string` to `[]net.IPNet` eventually anyway, maybe they should convert first and then call `IsDualStackIPs` and not have a `IsDualStackIPStrings` at all.

net/net.go
thockin: yeah, but with bool I can say `if !seen[val] { seen[val] = true }` which reads really well :)

Minor point, if other changes being made

net/net.go
thockin: I would say either:

a) assert that we transparently handle duplicates by silently de-duping (in which case IsDualStack doesn't care about dups), or...

b) assert that suplicate-ness is an error, make each caller check for dups (in which case IsDualStack doesn't care about dups)

duplicateness is orthogonal to dual-stackness, let's not jam them together just because they tend to both be checked by callers.

net/net.go
thockin: I don't follow.

[validv4] is valid but not dual-stack (not error)

[validv4, validv6, invalid] is not valid (error)

net/net.go
khenidak: the assumption that caller will always eventually work with `[]net.IP` or `[]net.IPNet` in case of cidrs. That assumption is arguable does not hold true for all. case in point api validation. an api-server does not really need to convert `cidrs` or `ip` to their true form, an api server just needs to ensure that they are dual stack. What i mean to say some callers will *only* work string form, others maybe converting..

net/net_test.go
apelisse: Better form is to use [`t.Run`](https://golang.org/pkg/testing/#T.Run)

net/net.go
thockin: OK.  I;d prefer less public functions, but this is not a killer

net/net.go
khenidak: removed the len < 2check

net/net.go
khenidak: I want with a). duplicates are silently ignored.

net/net.go
khenidak: this was removed, because duplicate check was removed.

net/net.go
thockin: comment still says dusp and >= 2

net/net.go
```diff
@@ -16,7 +16,106 @@ limitations under the License.
...
+	}
+
+	return (v4Found && v6Found), nil
+}
+
+// IsDualStackIPsStrings returns if
+// - all are valid ips
+// - at least one ip from each family (v4 or v6)
+// - an error is returned if length is less than 2
+// - an error is returned if duplicate ips are found
```
thockin: comments need to be updated

khenidak: opps.. removed the ref to `less than 2`.

net/net.go
```diff
@@ -35,3 +130,9 @@ func IsIPv6CIDRString(cidr string) bool {
 	ip, _, _ := net.ParseCIDR(cidr)
 	return IsIPv6(ip)
 }
+
+// IsIPv6CIDR returns if a cidr is ipv6
+func IsIPv6CIDR(cidr *net.IPNet) bool {
```
thockin: Do we actually need this, or can the caller simply call `IsIPv6(cidr.IP)`  ?

khenidak: no, I changed it.

net/net.go
khenidak: removed.

net/net.go
```diff
@@ -16,7 +16,98 @@ limitations under the License.
...
+	}
+	return cidrs, nil
+}
+
+// IsDualStackIPs returns if a slice of ips is:
+// - all are valid ips
+// - at least one ip from each family (v4 or v6)
+func IsDualStackIPs(ips []net.IP) (bool, error) {
+	v4Found := false
+	v6Found := false
```
andrewsykim: This can be simplified to
```
var v4Found, v6Found bool
```

net/net.go
```diff
@@ -16,7 +16,98 @@ limitations under the License.
...
+	for _, ip := range ips {
+		if ip == nil {
+			return false, fmt.Errorf("ip %v is invalid", ip)
+		}
+
+		if v4Found && v6Found {
+			continue
```
andrewsykim: break or return?

net/net_test.go
andrewsykim: ^ agreed, `errMessage` should be the test name passed into `t.Run`

PR #95, opened 2019-06-29
Adds support for initializing Trace with output options defined.

connor1989:
This PR allows the `k8s.io/utils/trace` library to be functional for downstream projects by adding an additional initializer that supports defining klog output options. With this new k8s libraries can get prototyped without being explicit with klog.

Example of no output written.

```
var (
	a int32
	b int32
)

func math(t *utiltrace.Trace) {
	t.Step("math func")

	a = 2
	b = 4

	sum := a + b

	t.Step(fmt.Sprintf("This is the sum: %v", sum))
}

func main() {

	trace := utiltrace.New("Init tracing")

	math(trace)

	trace.Log()

}
```

**Does this PR introduce a user-facing change?**:
`NONE`

Test included

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -19,6 +19,7 @@ package trace
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"math/rand"
 	"time"
 
@@ -41,6 +42,13 @@ type Trace struct {
 // New creates a Trace with the specified name
 func New(name string) *Trace {
 	return &Trace{name, time.Now(), nil}
+}
+
+// NewWithOutput creates a Trace with the specified name and klog output options defined
+func NewWithOutput(name string, writer io.Writer) *Trace {
+	// allows libraries to apply tracing and view output without declaring klog
+	klog.SetOutput(writer)
+	return New(name)
 }
 
 // Step adds a new step with a specific message. Call this at the end of an
--- a/trace/trace_test.go
+++ b/trace/trace_test.go
@@ -82,6 +82,34 @@ func TestTotalTime(t *testing.T) {
 	})
 }
 
+func TestForEmptyOutput(t *testing.T) {
+	test := struct {
+		name        string
+		sampleTrace *Trace
+	}{
+		name: "Check there is no log output written",
+	}
+
+	t.Run(test.name, func(t *testing.T) {
+		var buf bytes.Buffer
+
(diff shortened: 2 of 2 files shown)


dims:
/approve

/assign @wojtek-t 

Does this look good to you @wojtek-t ?

Thanks @connor1989 !

trace/trace.go
```diff
@@ -43,6 +44,13 @@ func New(name string) *Trace {
 	return &Trace{name, time.Now(), nil}
 }
 
+// NewWithOutput creates a Trace with the specified name and klog output options defined
+func NewWithOutput(name string, writer io.Writer) *Trace {
+	// allows libraries to apply tracing and view output without declaring klog
+	klog.SetOutput(writer)
```
wojtek-t: The overall idea looks fine.
But I'm not sure if this is the best implementation - if someone will use this function, they will overwrite output of all klog calls? That seems a bit dangerous to me...

Instead of that, I would suggest refactoring it a bit so that we create a writer somehow connected to klog in New method or use the ncoming from NewWithOutput instead of creating, so that the impact is more local.

Thoughts?
@dims ?

connor1989: I'm unclear whether calling `klog.setOutput` actually flushes or clears all thats been [collected](https://github.com/kubernetes/klog/blob/master/klog.go#L738). I didn't want to impose user facing changes with in mucking with `New`. 

Waiting for more direction on this before refactoring 👍

wojtek-t: > I'm unclear whether calling klog.setOutput actually flushes or clears all thats been collected. 

I didn't only mean about things that happened already. I was more afraid of what will happen if I use klog later in some independent code.

> I didn't want to impose user facing changes by mucking with New.

I wasn't saying about user-facing changes - I was only suggesting some implementation change underneath.

connor1989: @wojtek-t yes I see your point. I didn't think that far ahead with the change. Do you have any suggestions for updating the PR? 

@dims thoughts on this?

wojtek-t: It requires understanding it a bit better, so I don't have any specific suggestions TBH.

dims:
@connor1989 if the idea is to be able to check what gets logged, then the following may be better?

https://pasted.tech/pastes/29ff67055b2a9da15b00db3f2404a3aeed3d1735.raw

PR #96, opened 2019-07-12
Copy sigma inotify to k8s.io utils

dims:
Importing https://github.com/sigma/go-inotify into k8s.io/utils with history preserved.

needed to fix up problems mentioned here:
https://github.com/kubernetes/kubernetes/issues/75526


dims:
/assign @sigma @thockin 

cc @cpuguy83

thockin:
Thanks!

/lgtm
/approve

dims:
going to merge this as-is and then @Pothulapati offered to fix up the lint issues in a follow up.

PR #98, opened 2019-07-12
add the version package to utils

yastij:

This moves the version package to utils. it only depends on stdlib packages and is tested.

This would help unblock kubernetes/kubernetes#79526 specifically https://github.com/kubernetes/kubernetes/pull/79526#discussion_r299018383

cc @andrewsykim @rosti @dims

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -1,0 +1,314 @@
+/*
+Copyright 2016 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package version
+
+import (
+	"bytes"
+	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+// Version is an opqaue representation of a version number
+type Version struct {
+	components    []uint
+	semver        bool
+	preRelease    string
+	buildMetadata string
+}
+
+var (
+	// versionMatchRE splits a version string into numeric and "extra" parts
+	versionMatchRE = regexp.MustCompile(`^\s*v?([0-9]+(?:\.[0-9]+)*)(.*)*$`)
+	// extraMatchRE splits the "extra" part of versionMatchRE into semver pre-release and build metadata; it does not validate the "no leading zeroes" constraint for pre-release
+	extraMatchRE = regexp.MustCompile(`^(?:-([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?(?:\+([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?\s*$`)
+)
+
+func parse(str string, semver bool) (*Version, error) {
+	parts := versionMatchRE.FindStringSubmatch(str)
+	if parts == nil {
+		return nil, fmt.Errorf("could not parse %q as version", str)
+	}
+	numbers, extra := parts[1], parts[2]
+
+	components := strings.Split(numbers, ".")
+	if (semver && len(components) != 3) || (!semver && len(components) < 2) {
+		return nil, fmt.Errorf("illegal version string %q", str)
+	}
+
+	v := &Version{
+		components: make([]uint, len(components)),
+		semver:     semver,
+	}
+	for i, comp := range components {
+		if (i == 0 || semver) && strings.HasPrefix(comp, "0") && comp != "0" {
+			return nil, fmt.Errorf("illegal zero-prefixed version component %q in %q", comp, str)
+		}
+		num, err := strconv.ParseUint(comp, 10, 0)
+		if err != nil {
+			return nil, fmt.Errorf("illegal non-numeric version component %q in %q: %v", comp, str, err)
+		}
+		v.components[i] = uint(num)
+	}
+
+	if semver && extra != "" {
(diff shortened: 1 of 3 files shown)


yastij:
just lint error, but I wanted to kick discussions with this PR

yastij:
build fails because of some lint errors, but I wanted to kick discussions with this PR

dims:
yea, ignore the lint errors. someone is working on it.

version/version.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+
+	return v, nil
+}
+
+// ParseGeneric parses a "generic" version string. The version string must consist of two
+// or more dot-separated numeric fields (the first of which can't have leading zeroes),
+// followed by arbitrary uninterpreted data (which need not be separated from the final
+// numeric field by punctuation). For convenience, leading and trailing whitespace is
+// ignored, and the version can be preceded by the letter "v". See also ParseSemantic.
```
thockin: Actual examples of allowed values and disallowed values, highlighting the rules would be very helpful

version/version.go
```diff
@@ -0,0 +1,314 @@
...
+		panic(err)
+	}
+	return v
+}
+
+// ParseSemantic parses a version string that exactly obeys the syntax and semantics of
+// the "Semantic Versioning" specification (http://semver.org/) (although it ignores
+// leading and trailing whitespace, and allows the version to be preceded by "v"). For
+// version strings that are not guaranteed to obey the Semantic Versioning syntax, use
+// ParseGeneric.
```
thockin: actual examples would be helpful

version/version.go
```diff
@@ -0,0 +1,314 @@
...
+
+// Patch returns the patch release number if v is a Semantic Version, or 0
+func (v *Version) Patch() uint {
+	if len(v.components) < 3 {
+		return 0
+	}
+	return v.components[2]
+}
+
+// BuildMetadata returns the build metadata, if v is a Semantic Version, or ""
```
thockin: example of this

version/version.go
```diff
@@ -0,0 +1,314 @@
...
+func (v *Version) Major() uint {
+	return v.components[0]
+}
+
+// Minor returns the minor release number
+func (v *Version) Minor() uint {
+	return v.components[1]
+}
+
+// Patch returns the patch release number if v is a Semantic Version, or 0
```
thockin: What if a value is semver-like (triple) but not strictly semver - do we handle `X.Y.Z%foo`

version/version.go
```diff
@@ -0,0 +1,314 @@
...
+func MustParseSemantic(str string) *Version {
+	v, err := ParseSemantic(str)
+	if err != nil {
+		panic(err)
+	}
+	return v
+}
+
+// Major returns the major release number
+func (v *Version) Major() uint {
```
thockin: I would argue none of these methods should be on pointers.  The struct is value-like, we should treat it as such.  Pointer methods require everyone use pointer (swill comment more below).

https://play.golang.org/p/aDDvEZZ3aqX

version/version.go
```diff
@@ -0,0 +1,314 @@
...
+	return v.preRelease
+}
+
+// Components returns the version number components
+func (v *Version) Components() []uint {
+	return v.components
+}
+
+// WithMajor returns copy of the version object with requested major number
+func (v *Version) WithMajor(major uint) *Version {
```
thockin: I don't think we should be returning pointers.  It defeats some possible optimizations and forces everyone to use pointers.

https://play.golang.org/p/aDDvEZZ3aqX show how a helper function must return pointers to call these methods.  If this is value-like, treat it like a value.

version/version.go
```diff
@@ -0,0 +1,314 @@
...
+// ParseGeneric, this will not include the trailing uninterpreted portion of the version
+// number.
+func (v *Version) String() string {
+	var buffer bytes.Buffer
+
+	for i, comp := range v.components {
+		if i > 0 {
+			buffer.WriteString(".")
+		}
+		buffer.WriteString(fmt.Sprintf("%d", comp))
```
thockin: we sopecifically call components uint, but then you print it as an int here.  Why don't we just store it as int?

thockin: The you could use ItoA

version/version.go
```diff
@@ -0,0 +1,314 @@
...
+type Version struct {
+	components    []uint
+	semver        bool
+	preRelease    string
+	buildMetadata string
+}
+
+var (
+	// versionMatchRE splits a version string into numeric and "extra" parts
+	versionMatchRE = regexp.MustCompile(`^\s*v?([0-9]+(?:\.[0-9]+)*)(.*)*$`)
```
thockin: This seems to allow parsing of numbers that could be bigger than a uint?  Should we bound this to 10 characters (2^32) ?

version/version.go
```diff
@@ -0,0 +1,314 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+// AtLeast tests if a version is at least equal to a given minimum version. If both
+// Versions are Semantic Versions, this will use the Semantic Version comparison
+// algorithm. Otherwise, it will compare only the numeric components, with non-present
+// components being considered "0" (ie, "1.4" is equal to "1.4.0").
```
thockin: Examples of normal and edge cases?

danwinship:
> Don't we have like 4 semver libs in vendor/... already? Why do we have our own?

Well, this one is non just semver, which is why we have it (https://github.com/kubernetes/kubernetes/pull/35436). But it's already in k/apimachinery. Why would it need to be moved to k/utils?

version/version.go
lavalamp: Yeah just because it must be positive is NOT a good enough reason to use a uint.

lavalamp:
I think it'd be OK to move this (this is in apimachinery for lack of a better place; it's more build machinery, though, IMO), but it needs a name other than "version". Like, what's special about this version type? (Special attention to / opinions about all the stuff after the patch version number, it looks like.)

Either way, it'd be nice to address @thockin's comments on the source package. :) :)

PR #100, opened 2019-07-12
Add mount package from k/k/pkg/util/mount

codenrhoden:
Add the `mount` package from k/k, originally located at `pkg/util/mount`.

Git history from k/k is preserved.

fixes kubernetes/kubernetes#64953

/assign @msau42 
/assign @dims

### DIFF
--- a/hack/.golint_failures
+++ b/hack/.golint_failures
@@ -1,1 +1,2 @@
 # Apart from this line, only trailing comments are supported
+mount
--- a/hack/verify-golint.sh
+++ b/hack/verify-golint.sh
@@ -46,7 +46,6 @@ unset IFS
 errors=()
 not_failing=()
 for p in "${all_packages[@]}"; do
-  failing_packages="${failing_packages:-}"
   # Run golint on package/*.go file explicitly to validate all go files
   # and not just the ones for the current platform. This also will ensure that
   # _test.go files are linted.
@@ -90,26 +89,26 @@ else
   exit 1
 fi
 
-# if [[ ${#not_failing[@]} -gt 0 ]]; then
-#   {
-#     echo "Some packages in hack/.golint_failures are passing golint. Please remove them."
-#     echo
-#     for p in "${not_failing[@]}"; do
-#       echo "  $p"
-#     done
-#     echo
-#   } >&2
-#   exit 1
-# fi
-#
-# if [[ ${#gone[@]} -gt 0 ]]; then
-#   {
-#     echo "Some packages in hack/.golint_failures do not exist anymore. Please remove them."
-#     echo
-#     for p in "${gone[@]}"; do
-#       echo "  $p"
-#     done
-#     echo
-#   } >&2
-#   exit 1
-# fi
+if [[ ${#not_failing[@]} -gt 0 ]]; then
+  {
+    echo "Some packages in hack/.golint_failures are passing golint. Please remove them."
+    echo
+    for p in "${not_failing[@]}"; do
+      echo "  $p"
+    done
+    echo
+  } >&2
+  exit 1
+fi
+
+if [[ ${#gone[@]} -gt 0 ]]; then
+  {
+    echo "Some packages in hack/.golint_failures do not exist anymore. Please remove them."
+    echo
+    for p in "${gone[@]}"; do
+      echo "  $p"
+    done
+    echo
+  } >&2
+  exit 1
+fi
--- a/mount/OWNERS
+++ b/mount/OWNERS
@@ -1,0 +1,13 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+reviewers:
+  - jingxu97
+  - saad-ali
+  - jsafrane
+  - msau42
+  - andyzhangx
+approvers:
+  - jingxu97
+  - saad-ali
+  - jsafrane
+
--- a/mount/doc.go
+++ b/mount/doc.go
@@ -1,0 +1,18 @@
+/*
+Copyright 2014 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package mount defines an interface to mounting filesystems.
+package mount // import "k8s.io/kubernetes/pkg/util/mount"
--- a/mount/doc.go
+++ b/mount/doc.go
@@ -15,4 +15,4 @@ limitations under the License.
 */
 
 // Package mount defines an interface to mounting filesystems.
-package mount // import "k8s.io/kubernetes/pkg/util/mount"
+package mount // import "k8s.io/utils/mount"
--- a/mount/fake_mounter.go
+++ b/mount/fake_mounter.go
@@ -1,0 +1,204 @@
+/*
+Copyright 2015 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mount
+
+import (
+	"os"
+	"path/filepath"
+	"sync"
+
+	"k8s.io/klog"
+)
+
+// FakeMounter implements mount.Interface for tests.
+type FakeMounter struct {
+	MountPoints []MountPoint
+	log         []FakeAction
+	// Error to return for a path when calling IsLikelyNotMountPoint
+	MountCheckErrors map[string]error
+	// Some tests run things in parallel, make sure the mounter does not produce
+	// any golang's DATA RACE warnings.
+	mutex       sync.Mutex
+	UnmountFunc UnmountFunc
+}
+
+type UnmountFunc func(path string) error
+
+var _ Interface = &FakeMounter{}
+
+const (
+	// FakeActionMount is the string for specifying mount as FakeAction.Action
+	FakeActionMount = "mount"
+	// FakeActionUnmount is the string for specifying unmount as FakeAction.Action
+	FakeActionUnmount = "unmount"
+)
+
+// FakeAction objects are logged every time a fake mount or unmount is called.
+type FakeAction struct {
+	Action string // "mount" or "unmount"
+	Target string // applies to both mount and unmount actions
+	Source string // applies only to "mount" actions
+	FSType string // applies only to "mount" actions
+}
+
+func NewFakeMounter(mps []MountPoint) *FakeMounter {
+	return &FakeMounter{
+		MountPoints: mps,
+	}
+}
+
+// ResetLog clears all the log entries in FakeMounter
+func (f *FakeMounter) ResetLog() {
+	f.mutex.Lock()
+	defer f.mutex.Unlock()
+
+	f.log = []FakeAction{}
+}
+
+// GetLog returns the slice of FakeActions taken by the mounter
+func (f *FakeMounter) GetLog() []FakeAction {
+	f.mutex.Lock()
+	defer f.mutex.Unlock()
+
+	return f.log
+}
+
+// Mount records the mount event and updates the in-memory mount points for FakeMounter
+func (f *FakeMounter) Mount(source string, target string, fstype string, options []string) error {
+	f.mutex.Lock()
+	defer f.mutex.Unlock()
+
+	opts := []string{}
+
+	for _, option := range options {
+		// find 'bind' option
+		if option == "bind" {
+			// This is a bind-mount. In order to mimic linux behaviour, we must
+			// use the original device of the bind-mount as the real source.
+			// E.g. when mounted /dev/sda like this:
+			//      $ mount /dev/sda /mnt/test
+			//      $ mount -o bind /mnt/test /mnt/bound
+			// then /proc/mount contains:
+			// /dev/sda /mnt/test
+			// /dev/sda /mnt/bound
+			// (and not /mnt/test /mnt/bound)
+			// I.e. we must use /dev/sda as source instead of /mnt/test in the
+			// bind mount.
+			for _, mnt := range f.MountPoints {
+				if source == mnt.Path {
+					source = mnt.Device
+					break
+				}
+			}
+		}
+		// reuse MountPoint.Opts field to mark mount as readonly
+		opts = append(opts, option)
+	}
+
+	// If target is a symlink, get its absolute path
+	absTarget, err := filepath.EvalSymlinks(target)
+	if err != nil {
+		absTarget = target
+	}
+	f.MountPoints = append(f.MountPoints, MountPoint{Device: source, Path: absTarget, Type: fstype, Opts: opts})
+	klog.V(5).Infof("Fake mounter: mounted %s to %s", source, absTarget)
+	f.log = append(f.log, FakeAction{Action: FakeActionMount, Target: absTarget, Source: source, FSType: fstype})
+	return nil
+}
+
+// Unmount records the unmount event and updates the in-memory mount points for FakeMounter
+func (f *FakeMounter) Unmount(target string) error {
+	f.mutex.Lock()
+	defer f.mutex.Unlock()
+
+	// If target is a symlink, get its absolute path
+	absTarget, err := filepath.EvalSymlinks(target)
+	if err != nil {
+		absTarget = target
+	}
+
+	newMountpoints := []MountPoint{}
+	for _, mp := range f.MountPoints {
+		if mp.Path == absTarget {
+			if f.UnmountFunc != nil {
+				err := f.UnmountFunc(absTarget)
+				if err != nil {
+					return err
+				}
+			}
+			klog.V(5).Infof("Fake mounter: unmounted %s from %s", mp.Device, absTarget)
+			// Don't copy it to newMountpoints
+			continue
+		}
+		newMountpoints = append(newMountpoints, MountPoint{Device: mp.Device, Path: mp.Path, Type: mp.Type})
+	}
+	f.MountPoints = newMountpoints
+	f.log = append(f.log, FakeAction{Action: FakeActionUnmount, Target: absTarget})
+	delete(f.MountCheckErrors, target)
+	return nil
+}
+
+// List returns all the in-memory mountpoints for FakeMounter
+func (f *FakeMounter) List() ([]MountPoint, error) {
+	f.mutex.Lock()
+	defer f.mutex.Unlock()
+
+	return f.MountPoints, nil
+}
+
+// IsLikelyNotMountPoint determines whether a path is a mountpoint by checking
+// if the absolute path to file is in the in-memory mountpoints
+func (f *FakeMounter) IsLikelyNotMountPoint(file string) (bool, error) {
+	f.mutex.Lock()
+	defer f.mutex.Unlock()
+
+	err := f.MountCheckErrors[file]
+	if err != nil {
+		return false, err
+	}
+
+	_, err = os.Stat(file)
+	if err != nil {
+		return true, err
+	}
+
+	// If file is a symlink, get its absolute path
+	absFile, err := filepath.EvalSymlinks(file)
+	if err != nil {
+		absFile = file
+	}
+
+	for _, mp := range f.MountPoints {
+		if mp.Path == absFile {
+			klog.V(5).Infof("isLikelyNotMountPoint for %s: mounted %s, false", file, mp.Path)
+			return false, nil
+		}
+	}
+	klog.V(5).Infof("isLikelyNotMountPoint for %s: true", file)
+	return true, nil
+}
+
+// GetMountRefs finds all mount references to the path, returns a
+// list of paths.
+func (f *FakeMounter) GetMountRefs(pathname string) ([]string, error) {
+	realpath, err := filepath.EvalSymlinks(pathname)
+	if err != nil {
+		// Ignore error in FakeMounter, because we actually didn't create files.
+		realpath = pathname
+	}
+	return getMountRefsByDev(f, realpath)
+}
--- a/mount/fake_mounter.go
+++ b/mount/fake_mounter.go
@@ -36,6 +36,7 @@ type FakeMounter struct {
 	UnmountFunc UnmountFunc
 }
 
+// UnmountFunc is a function callback to be executed during the Unmount() call.
 type UnmountFunc func(path string) error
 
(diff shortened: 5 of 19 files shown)


codenrhoden:
Aww man, looks like I'm going to have to fix all the linting erros (mostly exported comments) for CI to pass. Well, better now than never. =)

I'm also surprised to see the Travis-CI job is not testing against Go 1.12.x. I'll add that too.

dims:
/assign @thockin 

Tim, can you please check if the API looks good to you?

codenrhoden:
A couple of updates/concerns on this one...

As far as pleasing the linter goes, I can *mostly* do this. But there is one snag that deserves attention. One of the linter complaints is the following:

> mount/mount.go:128:6: type name will be used as mount.MountPoint by other packages, and that stutters; consider calling this Point

I don't think we want to rename this -- **MountPoint** is the proper name. it is frustrating that there is no way to skip this instance of the complaint. "go lint" in general has taken the stance that they will never allow you to skip anything, as seen in [this comment](https://github.com/golang/lint/issues/263#issuecomment-363652697). So what are the options here?

- Enhance the verify script to skip this error just for mount? Seems like a pain to maintain
- Rename the exported constant? I don't like this either
- Disable linting? Seems like a bad idea. What about just for `mount?`. Also seems like a bad idea
- Switch to a different linter that allows fine-grain control? Maybe.

So that's one hurdle...

Another thing I notice is that the tests, linting, vetting, etc are only happening for Linux when doing CI. When running `make verify` it is only going to use the current OS, so that would be Darwin on my laptop. Which means that I don't see any linting issues for Windows or Linux unless I modify `GOOS` first. The `mount` package we are bringing in is larger than some of the other utilities here, and has full-blown Windows and Linux tests. It would seem problematic to me to not run the Windows tests for them. At a minimum I feel like we should run `go {fmt, lint, vet, build}` for all GOOS={linux, windows, darwin}. Right now you could check in code that doesn't compile on Windows and nothing would check it.

Modifying the CI to also run Windows tests could be an option, as Travis-CI supports Windows. But that would be a bit more work.  Still wanted to point all this out.

codenrhoden:
We could also play games like changing the package name from **mount** to **mnt**, and then importers could use `import mount "k8s.io/utils/mnt"` to minimize code change.

Alternatively change `mount.MountPoint` to `mount.MntPoint`, though this would necessitate many changes to LOC in k/k.

I don't like either option, as they are both band-aids for a silly problem, but just throwing it out there.

msau42:
I believe the cross-build job should run unit tests on all platforms. Maybe we need to add it here. cc @kubernetes/sig-testing-pr-reviews 

Regarding package naming, would naming it to "mounter" work as well? I kind of like that more than "mnt".

BenTheElder:
The cross build job does not run unit tests, it already takes a **ton** of resources to build for all (OS, Arch) pairs, and we're not running on those OSes, we're cross compiling to them.

The main repo has a `typecheck` tool that typechecks go code for all (OS, Arch) we cross compile to which is quicker and cheaper. It would probably be possible to adapt that to golint / go vet if someone were interested.

fejta:
Why is `mount.MountPoint` the proper name?

BenTheElder:
> Right now you could check in code that doesn't compile on Windows and nothing would check it.

Right, this repo has it's own CI setup that is maintained by ... someone? In the main Kubernetes repo when we import this package however we absolutely will ensure that it compiles for Windows etc. with the Prow CI there.

> I don't think we want to rename this -- MountPoint is the proper name. it is frustrating that there is no way to skip this instance of the complaint. "go lint" in general has taken the stance that they will never allow you to skip anything, as seen in this comment. So what are the options here?

go lint is enforced by most projects and does indeed take this tance.

You're normally going to reference it like `mount.Point` from any using packages so that could be reasonable. Everything in the package is mount \<something>.

In the main repo we have the linter set to ignore packages because the code generators produce go code that does not lint, but generally we've tried to avoid that.

codenrhoden:
@fejta
> Why is `mount.MountPoint` (how it will always be declared) the proper name?

I may be assuming too much, but I think it's fair to say that those that work with storage are familiar with the term "mount point", and that the term "point" doesn't mean anything to them. @msau42's suggestion would alleviate this problem as well.

@msau42 
> Regarding package naming, would naming it to "mounter" work as well? I kind of like that more than "mnt".

Yes, that would work too.

@BenTheElder 

> The main repo has a typecheck tool that typechecks go code for all (OS, Arch) we cross compile to which is quicker and cheaper. It would probably be possible to adapt that to golint / go vet if someone were interested.
>
>Right, this repo has it's own CI setup that is maintained by ... someone? In the main Kubernetes repo when we import this package however we absolutely will ensure that it compiles for Windows etc. with the Prow CI there.

Yes, that tool has caught many things for me over time. =) I don't really know what the status of the CI setup on this repo is, or who maintains it. I made some enhancements to it last week when I found missing things (like go 1.12 support), which was the first time I had looked at it.

I'm mostly trying to make sure that we don't take a step backwards in testing with the mount package moving here. While I know that any issues would be picked up on the k/k side, I am trying to look at this from the perspective that these utils are meant for consumption by many others besides k/k, and therefore should be vetted/tested independent of k/k. It may fall upon k/k to pick up bugs with the Windows code, but at the very least we should vet/lint/x-compile things here to make sure things are in good order. I am just pointing out that that isn't happening today.

Put more succinctly, do people see that as a blocker for this package coming in? Should we fix that first? Or take it as a follow-up to add that type of checking to this repo?

stevekuznetsov:
Unless you are operating inside of the utility, though, it is impossible to reference this object without it's module, so you _always_ write `mount.Point`. That doesn't sutter and has all the same context that someone who knows about mount points needs to understand what a `mount.Point` is.

codenrhoden:
> Unless you are operating inside of the utility, though, it is impossible to reference this object without it's module, so you _always_ write `mount.Point`. That doesn't sutter and has all the same context that someone who knows about mount points needs to understand what a `mount.Point` is.

I see your point. Coming at this from a different direction, I'm trying to avoid changing all of [these lines](https://gist.github.com/codenrhoden/61434752051de896503527a80aeabf1b) in k/k. I've taken @msau42's suggestion for now, but that is easily changed back if needed.

msau42:
/lgtm

@dims how do we address the invalid commit messages and cla issues?

dims:
@msau42 we will use the "Merge Pull Request" button to bypass the CLABot (it's broken!)

codenrhoden:
@dims just to follow-up, you are waiting on review from @thockin ?

dims:
@codenrhoden yep! waiting on @thockin

mounter/mount.go
```diff
@@ -0,0 +1,377 @@
...
+	MakeRShared(path string) error
+	// GetFileType checks for file/directory/socket/block/character devices.
+	// Will operate in the host mount namespace if kubelet is running in a container
+	GetFileType(pathname string) (FileType, error)
+	// MakeFile creates an empty file.
+	// Will operate in the host mount namespace if kubelet is running in a container
+	MakeFile(pathname string) error
+	// MakeDir creates a new directory.
+	// Will operate in the host mount namespace if kubelet is running in a container
+	MakeDir(pathname string) error
```
thockin: what if the file exists? What if the path exists but is not a dir?

mounter/mount.go
```diff
@@ -0,0 +1,377 @@
...
+	// Will operate in the host mount namespace if kubelet is running in a container.
```
thockin: Don't talk about kubelet. Maybe "caller" ?

Why this particular restriction and how is that determined?

This is an interface - why is it talking about implementation details?

mounter/mount.go
```diff
@@ -0,0 +1,377 @@
...
+	// Error is returned on any other error than "file not found".
+	ExistsPath(pathname string) (bool, error)
```
thockin: PathExists

mounter/mount.go
```diff
@@ -0,0 +1,377 @@
...
+	// EvalHostSymlinks returns the path name after evaluating symlinks.
+	// Will operate in the host mount namespace if kubelet is running in a container.
```
thockin: Don't talk about kubelet. Maybe "caller" ?

Why this particular restriction and how is that determined?

This is an interface - why is it talking about implementation details?

thockin: What limits on number of expansions (to prevent a loop) ?

mounter/mount.go
```diff
@@ -0,0 +1,377 @@
...
+	EvalHostSymlinks(pathname string) (string, error)
```
thockin: If this whole interface is is "host", do we need Host in the name?

mounter/mount.go
```diff
@@ -0,0 +1,377 @@
...
+	// GetFSGroup returns FSGroup of the path.
```
thockin: FSGroup is a process property.  Group is a file property.  GetGroup.  Or `GetOwner() (uid, gid, error)`

mounter/mount.go
```diff
@@ -0,0 +1,377 @@
...
+
+// SafeFormatAndMount probes a device to see if it is formatted.
+// Namely it checks to see if a file system is present. If so it
+// mounts it otherwise the device is formatted first then mounted.
+type SafeFormatAndMount struct {
+	Interface
+	Exec
+}
+
+// FormatAndMount formats the given disk, if needed, and mounts it.
```
thockin: What details can we provide about how it will be formatted?

mounter/mount.go
```diff
@@ -0,0 +1,377 @@
...
+		}
+	}
+	return notMnt, nil
+}
+
+// IsBind detects whether a bind mount is being requested and makes the remount options to
+// use in case of bind mount, due to the fact that bind mount doesn't respect mount options.
+// The list equals:
+//   options - 'bind' + 'remount' (no duplicate)
+func IsBind(options []string) (bool, []string, []string) {
```
thockin: IsBindRequest ?  It's weird that all the other Is* functions take a path and this is checking a mount command...

mounter/mount.go
```diff
@@ -0,0 +1,377 @@
...
+func checkForNetDev(options []string) bool {
+	for _, option := range options {
+		if option == "_netdev" {
+			return true
+		}
+	}
+	return false
+}
+
+// HasMountRefs checks if the given mountPath has moutRefs
```
thockin: s/moutRefs/mountRefs

mounter/mount.go
```diff
@@ -0,0 +1,377 @@
...
+// TODO: this is a workaround for the unmount device issue caused by gci mounter.
```
thockin: Do we still need this?  We should not add TODOs like this into the util API - we either add it and keep it, or we we don't add it.

mounter/mount_helper_common.go
```diff
@@ -0,0 +1,102 @@
...
+package mounter
+
+import (
+	"fmt"
+	"os"
+
+	"k8s.io/klog"
+)
+
+// CleanupMountPoint unmounts the given path and
```
thockin: weird line wraps here

codenrhoden:
Thanks for the feedback, @thockin. 

In order to address the review comments, I think the best way is for me to implement them back in the `k/k` repo and then attempt this extraction again.

mounter/mount.go
```diff
@@ -0,0 +1,377 @@
+/*
```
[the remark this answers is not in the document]
codenrhoden: Not this one. It's used by all platforms.

mounter/mount.go
```diff
@@ -0,0 +1,377 @@
...
+	// PathIsDevice determines if a path is a device.
+	PathIsDevice(pathname string) (bool, error)
+	// GetDeviceNameFromMount finds the device name by checking the mount path
+	// to get the global mount path within its plugin directory
+	GetDeviceNameFromMount(mounter Interface, mountPath, pluginMountDir string) (string, error)
+	// MakeRShared checks that given path is on a mount with 'rshared' mount
+	// propagation. If not, it bind-mounts the path as rshared.
+	MakeRShared(path string) error
+	// GetFileType checks for file/directory/socket/block/character devices.
+	// Will operate in the host mount namespace if kubelet is running in a container
```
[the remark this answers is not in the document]
codenrhoden: Noted. I'll just remove that part. I think that part of the comment came in when the NSEnter implementation was created. The comment is specific to that implementation and does not apply to the interface.

mounter/exec.go
```diff
@@ -0,0 +1,53 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mounter
+
+import "k8s.io/utils/exec"
+
+// NewOsExec returns a new Exec interface implemenation based on utils/exec
+func NewOsExec() Exec {
```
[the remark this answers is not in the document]
codenrhoden: https://github.com/kubernetes/kubernetes/pull/81187

mounter/doc.go
```diff
@@ -0,0 +1,18 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package mounter defines an interface to mounting filesystems.
+package mounter // import "k8s.io/utils/mounter"
```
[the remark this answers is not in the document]
codenrhoden: Yes. we had chose `mounter` to avoid having to rename `MountPoint`, but I went ahead and did it now...

https://github.com/kubernetes/kubernetes/pull/81186

mounter/exec.go
```diff
@@ -0,0 +1,53 @@
...
+
+var _ Exec = &osExec{}
+
+func (e *osExec) Run(cmd string, args ...string) ([]byte, error) {
+	exe := exec.New()
+	return exe.Command(cmd, args...).CombinedOutput()
+}
+
+// NewFakeExec returns a new FakeExec
+func NewFakeExec(run runHook) *FakeExec {
```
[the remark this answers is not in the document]
codenrhoden: @msau42 I took a stab at creating a `testing` subdir and putting the fakes in there, and immediately hit a snag. I want to get your opinion before I proceed with a solution...

We have three "fakes" -- `FakeMounter`, `FakeHostUtil`, and `FakeExec`. Two have them have implementations that reference types found in `mount`, so it has to be imported. However, a lot of the tests in `mount` use these fakes -- so moving the fakes into a new package creates an import cycle. The normal solution in this case, I believe, would be to create a `mount/types` package, and have them both import that. This will have a pretty large ripple effect, so I want to get your opinion on that before I try it out.

Even once that problem is solved, there is a second one. There are two places in the fakes that call functions found in `mount`, one of which isn't exported. we could export it, but end up with the import cycle again... I think  the way around this is to implement whatever functionality is needed in the fake directly -- it can probably even be simpler.  I think the re-use of functions there was out of convenience. but if we want a real "fake" that is separate, it can't re-use things from `mount`.  (well, types aside, see problem 1).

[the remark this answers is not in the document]
codenrhoden: Oh, and to answer Tim's question on whether `NewFakeExec` has a lot of callers -- it does. It is used by several volume plugins.

mounter/mount.go
codenrhoden: https://github.com/kubernetes/kubernetes/pull/81204

mounter/mount.go
codenrhoden: https://github.com/kubernetes/kubernetes/pull/81230

mounter/exec.go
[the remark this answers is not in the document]
msau42: Just randomly searching around for other fakes, can you see how pkg/kubelet/container/testing.FakeRuntime is solving this import cycle?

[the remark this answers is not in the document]
codenrhoden: Yeah I had looked at a bunch of those for inspiration. It looks like they just took the test that imports the fakes and made it a different package...

https://github.com/kubernetes/kubernetes/blob/acaac181dc8f4c16ed0f5e47f687a6475cae49f3/pkg/kubelet/container/runtime_cache_test.go#L17

That's simple enough. :)

thockin:
Assign back to me once ready.  Feel free to send me the k/k version, too.  I can look at it from an API POV with this extraction in mind.  Would be nice to have more eyes on code review, too :)

mounter/exec.go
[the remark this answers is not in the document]
codenrhoden: https://github.com/kubernetes/kubernetes/pull/81423

hogepodge:
Do we think this patch will be complete before code freeze?

We need to make a decision on if we will keep waiting for this patch or make the move to staging with the code duplication we have in place. https://github.com/kubernetes/kubernetes/pull/80027

codenrhoden:
I'm definitely working to address all the feedback provided by @thockin . 5 PRs so far (4 merged), but there are more to come. Often one little change in a public API in the `mount` package touches 100+ files, so getting approvals can be challenging.

I have no real feeling on whether this will all happen before code freeze or not.

hogepodge:
It's a heroic effort :-) Thank you! If you cc me on the PRs I can offer my eyes for code review.

mounter/mount.go
```diff
@@ -0,0 +1,377 @@
...
+	// Mount mounts source to target as fstype with given options.
+	Mount(source string, target string, fstype string, options []string) error
+	// Unmount unmounts given target.
+	Unmount(target string) error
+	// List returns a list of all mounted filesystems.  This can be large.
+	// On some platforms, reading mounts is not guaranteed consistent (i.e.
+	// it could change between chunked reads). This is guaranteed to be
+	// consistent.
+	List() ([]MountPoint, error)
+	// IsMountPointMatch determines if the mountpoint matches the dir
```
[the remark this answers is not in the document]
codenrhoden: https://github.com/kubernetes/kubernetes/pull/81730

mounter/mount.go
codenrhoden: @msau42 @saad-ali  Any insight on this one? I'm not familiar with what is necessary within GCI...  This method only has one caller, in `pkg/volume/util/operationexecutor/operation_generator.go`.

msau42: I'm a bit confused by this function. I feel like the code is doing the opposite of what the comments say.  @jingxu97 @saad-ali we need to investigate this one.

mounter/mount.go
```diff
@@ -0,0 +1,377 @@
...
+	GetDeviceNameFromMount(mounter Interface, mountPath, pluginMountDir string) (string, error)
+	// MakeRShared checks that given path is on a mount with 'rshared' mount
+	// propagation. If not, it bind-mounts the path as rshared.
+	MakeRShared(path string) error
+	// GetFileType checks for file/directory/socket/block/character devices.
+	// Will operate in the host mount namespace if kubelet is running in a container
+	GetFileType(pathname string) (FileType, error)
+	// MakeFile creates an empty file.
+	// Will operate in the host mount namespace if kubelet is running in a container
+	MakeFile(pathname string) error
```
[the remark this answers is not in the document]
codenrhoden: @msau42 I took a quick glance of whether `MakeFile` and `MakeDir` are needed in this interface at all... They are only used by one volume driver -- host path. But I think they are needed because of nsenter -- they both work differently in the nsenter case...  I'll just try to address the open questions with comments.

mounter/mount.go
msau42: Ok I understand now.  `Contains()` checks anywhere in the string, not just from the beginning, so the duplicated device mount under the shared bind-mounted /var/lib/kubelet (in GCI/COS case) will match.  

Given that this is only used by k8s and is not something that a csi driver would use, can we keep it inside k8s and not export it?

(Also, I think there is a bug using `Contains` but we can open an issue and address that later)

mounter/mount.go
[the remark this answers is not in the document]
msau42: Since it's only used by one plugin, I think it is fine to leave it out of the interface.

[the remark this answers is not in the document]
codenrhoden: One annoying thing is that the answer to "what if the path already exists" is different for `nsenter` vs everything else... But since `nsenter` is internal to k/k only, I'm not going to document that behavior. It won't apply to `k/utils/mount`.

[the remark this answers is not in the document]
codenrhoden: > Since it's only used by one plugin, I think it is fine to leave it out of the interface.

But I think we have to have it, because the implementation is different for `nsenter`. The host path driver has no notion of when the kubelet is containerized or not, that's abstracted into what version of `HostUtils` was created. So until that is gone, I think we are forced to keep this.

I've opened https://github.com/kubernetes/kubernetes/pull/81739 to capture all of this, though.

mounter/mount.go
codenrhoden: Thanks for looking into this one, @msau42.

https://github.com/kubernetes/kubernetes/pull/81745

thockin:
With code freeze looming, and knowing that I will be AFK much of tomorrow, is there anything I can do to help make this fly before code freeze?

codenrhoden:
> With code freeze looming, and knowing that I will be AFK much of tomorrow, is there anything I can do to help make this fly before code freeze?

Thanks @thockin. With the some of the other work that's been identified, and comments I've gotten here: https://github.com/kubernetes/kubernetes/pull/82031#discussion_r318240067, It does not look like this will happen for 1.16. It's *really* close, but as we did additional cleanups, we found even more we can do to **do it right**.

liggitt:
Is there a concrete list of remaining items to address? I'd love to see this land in the 1.17 timeframe to pay down the duplication debt in the k8s.io/legacy-cloud-providers component and for use by [other components](https://github.com/deislabs/secrets-store-csi-driver/blob/master/pkg/secrets-store/nodeserver.go#L40) that are currently having to depend on k8s.io/kubernetes only for the mount utilities.

codenrhoden:
Hi @liggitt,

I'd very much like to see this get finished up as well. The list as I know it today is the following:

- Removing the use of `k/k/pkg/util/mount.Exec` from `k/k` in favor of `k/utils/exec`. This PR has not been opened yet, as I'm also waiting on....
- https://github.com/kubernetes/utils/pull/118  This is needed to be able to do the same type of testing thoughout `k/k`...
- https://github.com/kubernetes/kubernetes/pull/82031 (requires a rebase after the Exec stuff is done
- https://github.com/kubernetes/kubernetes/pull/81423
- https://github.com/kubernetes/utils/pull/100 (this PR, which would need to be redone from scratch once all the above is done.

@msau42 am I missing anything?

codenrhoden:
Wow, this is finally ready for re-review!

The entire package has changed quite a bit since last time. I believe all of Tim's original review comments have been addressed. The package is significantly smaller, too, as we we completely removed the Exec interface.

I know of two problems right now that will pop up from golint:

```
mount/mount.go:70:6: type name will be used as mount.MountPoint by other packages, and that stutters; consider calling this Point
mount/mount_helper_unix.go:59:6: type name will be used as mount.MountInfo by other packages, and that stutters; consider calling this Info
```

Same issue with package name stuttering as before. We considered renaming `MountPoint`, but that was [not well received](https://github.com/kubernetes/kubernetes/pull/81186). So all of the suggestion in my [original comment](https://github.com/kubernetes/utils/pull/100#issuecomment-511518816) still apply:

> - Enhance the verify script to skip this error just for mount? Seems like a pain to maintain
> - Rename the exported constant? I don't like this either
> - Disable linting? Seems like a bad idea. What about just for mount?. Also seems like a bad idea
> - Switch to a different linter that allows fine-grain control? Maybe.

/assign @thockin @msau42 @dims @liggitt

dims:
@codenrhoden many thanks!!! let's port over the .golint_failures / verify-golint.sh over from k/k and ignore the mount package like you mentioned please.

dims:
this code is straight out of k/k with a lot of re-work / feedback there. shipping it!

hogepodge:
Thank you @codenrhoden for this epic effort!

PR #101, opened 2019-07-12
Use IN_CLOEXEC when creating notify watches.

cpuguy83:
This prevents potential inotify FD leaks to sub-processes.

Related to kubernetes/kubernetes#75526


dims:
manual test - https://github.com/kubernetes/kubernetes/pull/80689#issuecomment-518667205

PR #102, opened 2019-07-12
Add go 1.12 to Travis-CI

codenrhoden:
There currently is no testing against Go 1.12, which the latest K8s releases explicitly require. It should be in the testing matrix.

I'm not sure what the policy is for dropping old versions. If it is n-2, we could drop 1.9, but wanted to find out about that first.

/assign @dims

### DIFF
--- a/hack/verify-govet.sh
+++ b/hack/verify-govet.sh
@@ -1,4 +1,6 @@
-# Copyright 2018 The Kubernetes Authors.
+#!/bin/bash
+
+# Copyright 2019 The Kubernetes Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


dims:
@codenrhoden what's the best way to deal with the `vet` command line changes?

```
go tool vet .
vet: invoking "go tool vet" directly is unsupported; use "go vet"
Makefile:33: recipe for target 'vet' failed
```

codenrhoden:
@dims seeing if this new version will work...  From the new commit message, here is what I did

>Since `go tool vet` is deprecated in Go 1.12, add a hack/verify-govet.sh
script that behaves differently for Go 1.{9,10,11}, with the default
behavior supporting the new `go vet` for 1.12+.

codenrhoden:
Woohoo! that did it. I guess the only question now is whether go 1.9 support needs to stick around.

hack/verify-govet.sh
```diff
@@ -0,0 +1,30 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# For use with Go 1.12+
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+TRAVIS_GO_VERSION=${TRAVIS_GO_VERSION:-}
```
dims: Can we please drop down to use `go version` when `TRAVIS_GO_VERSION` is not set?

codenrhoden: Yeah, but then I have to parse it... =)

`go version` gives me `go version go1.12.1 darwin/amd64`, whereas `TRAVIS_GO_VERSION` just gives me `1.12` or `1.12.1` or `1.12.x`.

I was trying to avoid parsing it by making the default behavior be, if that env var is not set, just run the 1.12+ version (which is `go vet`), as most K8s devs should be on 1.12 anyways. But maybe that's a faulty assumption?

codenrhoden: If I need to parse it, I might as well parse it all the time. It wouldn't be that hard... I can do that.

codenrhoden:
@dims okay, I think you'll like this one better. =) Simpler, too.

regex explanation: https://regex101.com/r/tx8naU/1

dims:
much better. thanks @codenrhoden

PR #103, opened 2019-07-15
net: add GetHostname API

bart0sh:
Copied GetHostname API from k/k/pkg/util/node to use in
Kubernetes components outside of Kubernetes core.

This PR is pre-requisite for removing kubeadm dependency to pkg/util/node.
See more details in [this issue](https://github.com/kubernetes/kubeadm/issues/1600)

### DIFF
--- a/net/net.go
+++ b/net/net.go
@@ -17,8 +17,11 @@ limitations under the License.
 package net
 
 import (
+	"errors"
 	"fmt"
 	"net"
+	"os"
+	"strings"
 )
 
 // ParseCIDRs parses a list of cidrs and return error if any is invalid.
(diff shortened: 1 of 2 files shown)


net/net.go
```diff
@@ -132,3 +135,23 @@ func IsIPv6CIDR(cidr *net.IPNet) bool {
 	ip := cidr.IP
 	return IsIPv6(ip)
 }
+
+// GetHostname returns OS's hostname if 'hostnameOverride' is empty; otherwise, return 'hostnameOverride'
+func GetHostname(hostnameOverride string) (string, error) {
```
yastij: @thockin - I'm not sure this should get in `utils` as is. we either should drop the override behaviour or rename the func to show the real intent

bart0sh: @yastij dropping override behaviour would make it almost identical to os.Hostname. Let's rename it. Any name suggestions?

yastij: dering to @thockin on this

neolit123:
@kubernetes/sig-node-pr-reviews 
PTAL.

neolit123:
@thockin 

> Is the trims logic even needed

it was already present in the old code, this PR is just moving the function.

> My inclination is to not do this, unless there is something non-obvious I am missing? 

if kubeadm, kubelet and other tools start deviating in terms of `GetHostname` we might see unexpected results. given this is part of the kubeadm decoupling from k/k, the kubeadm alternative to this PR is to just copy duplicate the function. would that be preferred?

bart0sh:
> I his feels like the sort of thing where de-duping is pointless. My inclination is to not do this, unless there is something non-obvious I am missing? Is the trims logic even needed

I'm generally ok with either way of getting rid of dependency on pkg/util/node. This is my first attempt to do this using duping: https://github.com/kubernetes/kubernetes/pull/79477

Just wanted to point out that GetHostname is used in the kube-proxy, kubeadm, kubelet and pkg/volume/rbd/rbd_util. In extreme case we can end up with 4 copies of this function. Not sure we want that.

PR #104, opened 2019-07-19
add procfs APIs

bart0sh:
This module is copied from k/k/pkg/util/procfs to use in
Kubernetes components outside of Kubernetes core

These changes have been done for the original source code
mainly to pass kubernetes/utils checks and avoid bringing additional
dependencies to the package:

- Added comments to fix golint warnings:
```
  procfs/procfs.go:19:6: exported type ProcFSInterface should have comment or be unexported
  procfs/procfs_fake.go:19:6: exported type FakeProcFS should have comment or be unexported
  procfs/procfs_fake.go:21:1: exported function NewFakeProcFS should have comment or be unexported
  procfs/procfs_linux.go:39:6: exported type ProcFS should have comment or be unexported
  procfs/procfs_linux.go:41:1: exported function NewProcFS should have comment or be unexported
  procfs/procfs_linux.go:56:1: comment on exported method ProcFS.GetFullContainerName should be of the form "GetFullContainerName ..."
  procfs/procfs_linux.go:71:1: comment on exported function PKill should be of the form "PKill ..."
  procfs/procfs_linux.go:94:1: comment on exported function PidOf should be of the form "PidOf ..."
```
- Renamed ProcFSInterface to ProcInterface to fix
  golint warning:
```
    procfs/procfs.go:21:6: type name will be used as procfs.ProcFSInterfaceby other packages, and that stutters; consider calling this Interface
```
- Got rid of dependency to
  k8s.io/apimachinery/pkg/util/errors:NewAggregate API

This PR is pre-requisite for removing kubeadm dependency to pkg/util/procfs.
See more details in [this issue](https://github.com/kubernetes/kubeadm/issues/1600) and [this PR](https://github.com/kubernetes/kubernetes/pull/80296)

### DIFF
--- a/procfs/doc.go
+++ b/procfs/doc.go
@@ -1,0 +1,18 @@
+/*
+Copyright 2015 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package procfs implements utility functions relating to the /proc mount.
+package procfs // import "k8s.io/utils/procfs"
--- a/procfs/example_proc_cgroup
+++ b/procfs/example_proc_cgroup
@@ -1,0 +1,10 @@
+11:name=systemd:/user/1000.user/c1.session
+10:hugetlb:/user/1000.user/c1.session
+9:perf_event:/user/1000.user/c1.session
+8:blkio:/user/1000.user/c1.session
+7:freezer:/user/1000.user/c1.session
+6:devices:/user/1000.user/c1.session
+5:memory:/user/1000.user/c1.session
+4:cpuacct:/user/1000.user/c1.session
+3:cpu:/user/1000.user/c1.session
+2:cpuset:/
--- a/procfs/procfs.go
+++ b/procfs/procfs.go
@@ -1,0 +1,24 @@
+/*
+Copyright 2015 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package procfs
+
+// ProcInterface that presents GetFullContainerName API implemented
+// by ProcFS and FakeProcFS
+type ProcInterface interface {
+	// GetFullContainerName gets the container name given the root process id of the container.
+	GetFullContainerName(pid int) (string, error)
+}
--- a/procfs/procfs_linux.go
+++ b/procfs/procfs_linux.go
@@ -1,0 +1,180 @@
+// +build linux
+
+/*
+Copyright 2015 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package procfs
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+	"syscall"
+	"unicode"
+
+	"k8s.io/klog"
+)
+
+// ProcFS implements ProcInterface
+type ProcFS struct{}
+
+// NewProcFS creates ProcFS object
+func NewProcFS() ProcInterface {
+	return &ProcFS{}
+}
+
+func containerNameFromProcCgroup(content string) (string, error) {
+	lines := strings.Split(content, "\n")
+	for _, line := range lines {
+		entries := strings.SplitN(line, ":", 3)
+		if len(entries) == 3 && entries[1] == "devices" {
+			return strings.TrimSpace(entries[2]), nil
+		}
+	}
+	return "", fmt.Errorf("could not find devices cgroup location")
+}
+
+// GetFullContainerName gets the container name given the root process id of the container.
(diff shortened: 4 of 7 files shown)


procfs/example_proc_cgroup
```diff
@@ -0,0 +1,10 @@
+11:name=systemd:/user/1000.user/c1.session
```
andrewsykim: Why do we need this file here?

procfs/doc.go
```diff
@@ -0,0 +1,18 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package procfs implements utility functions relating to the /proc mount.
+package procfs // import "k8s.io/utils/procfs"
```
andrewsykim: I don't think we've been following this convention in this repo, let's get rid of this file?

procfs/procfs.go
```diff
@@ -0,0 +1,24 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package procfs
+
+// ProcInterface that presents GetFullContainerName API implemented
+// by ProcFS and FakeProcFS
+type ProcInterface interface {
```
andrewsykim: Maybe just `Interface` then if golint is complaining?

procfs/doc.go
bart0sh: I left it here as I saw it in 3 more places:
```
$ find . -name doc.go
./exec/doc.go
./temp/temptest/doc.go
./temp/doc.go
```

procfs/example_proc_cgroup
bart0sh: it was in original sources for a reason I guess. Here is where it was added:
```
commit 6ef3de1d5f1d4b19c154ba15f2bd4686b6036f0c
Author: Ananya Kumar <skywalker94@gmail.com>
Date:   Mon Aug 3 17:28:33 2015 -0700

    Add QoS support on node
```

procfs/procfs.go
bart0sh: golint is not complaining about this one. It was complaining about ProcFSInterface. I just thought that ProcInterface is closer to the original name.

bart0sh:
> I don't see any consumers of `ProcFSInterface` / `GetFullContainerName`, am I missing something? Seems like kubeadm is just consuming the `PidOf` function, can we just move that and delete the rest?

The idea was to move the module as is in order to remove it from k/k. If we move just part of it it would be harder to do I guess.

procfs/procfs_linux.go
```diff
@@ -0,0 +1,180 @@
...
+
+	"k8s.io/klog"
+)
+
+// ProcFS implements ProcInterface
+type ProcFS struct{}
+
+// NewProcFS creates ProcFS object
+func NewProcFS() ProcInterface {
+	return &ProcFS{}
```
thockin: Don't return a pointer - that just wastes heap when it could stay stack-allocated

procfs/procfs_linux.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+	}
+	return "", fmt.Errorf("could not find devices cgroup location")
+}
+
+// GetFullContainerName gets the container name given the root process id of the container.
+// E.g. if the devices cgroup for the container is stored in /sys/fs/cgroup/devices/docker/nginx,
+// return docker/nginx. Assumes that the process is part of exactly one cgroup hierarchy.
+func (pfs *ProcFS) GetFullContainerName(pid int) (string, error) {
+	filePath := path.Join("/proc", strconv.Itoa(pid), "cgroup")
```
thockin: I would not hardcode the name like this.  Instead, something like:

```
const realSlashProcBasePath = "/proc"

func New() ProcFS {
    return newWithBasePath(realSlashProcBasePath)
}

// useful for testing
func newWithBasePath(basePath string) ProcFS {
    return ProcFs{
      basPath: basePath,
    }
}

File: procfs/procfs_linux.go [hunk 6]
```diff
@@ -0,0 +1,180 @@
...
+	content, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		if os.IsNotExist(err) {
```
thockin: how is this different than just returning ` "", err ` in all cases?

procfs/procfs_linux.go
```diff
@@ -0,0 +1,180 @@
...
+	for _, line := range lines {
+		entries := strings.SplitN(line, ":", 3)
+		if len(entries) == 3 && entries[1] == "devices" {
+			return strings.TrimSpace(entries[2]), nil
+		}
+	}
+	return "", fmt.Errorf("could not find devices cgroup location")
+}
+
+// GetFullContainerName gets the container name given the root process id of the container.
```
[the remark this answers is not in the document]
thockin: This function is trying to be abstract, but "container name" is poorly defined.  I think we should abandon the pretense of an Interface and call this `GetCGroupPath(pid)` or similar.

Also, I can't see anyone who calls this, so maybe just nuke it?

procfs/procfs_linux.go
```diff
@@ -0,0 +1,180 @@
...
+// E.g. if the devices cgroup for the container is stored in /sys/fs/cgroup/devices/docker/nginx,
+// return docker/nginx. Assumes that the process is part of exactly one cgroup hierarchy.
+func (pfs *ProcFS) GetFullContainerName(pid int) (string, error) {
+	filePath := path.Join("/proc", strconv.Itoa(pid), "cgroup")
+	content, err := ioutil.ReadFile(filePath)
```
thockin: Should probably read into a fixed buffer of 4KB or 8KB

procfs/procfs_linux_test.go
```diff
@@ -0,0 +1,116 @@
...
+		t.Skipf("not supported on GOOS=%s", runtime.GOOS)
+	}
+	pids, err := PidOf(filepath.Base(os.Args[0]))
+	assert.Empty(t, err)
+	assert.NotZero(t, pids)
+	assert.Contains(t, pids, os.Getpid())
+}
+
+func TestPKill(t *testing.T) {
+	if runtime.GOOS == "darwin" || runtime.GOOS == "windows" {
```
thockin: This is already `+build linux` - all of these are a waste?

procfs/procfs_linux_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+	procCgroupInvalid := "devices:docker/kubelet\ncpuacct:pkg/kubectl"
+	verifyContainerName(procCgroupInvalid, "", true, t)
+}
+
+func TestPidOf(t *testing.T) {
+	if runtime.GOOS == "darwin" || runtime.GOOS == "windows" {
+		t.Skipf("not supported on GOOS=%s", runtime.GOOS)
+	}
+	pids, err := PidOf(filepath.Base(os.Args[0]))
```
thockin: This is a poor test.  A real test would have a phony /proc directory to iterate over.  Or several, to trigger failures.

neolit123:
@bart0sh as you saw, we found a workaround to not use the procfs package in cmd/kubeadm, thus this PR is no longer needed from the kubeadm POV.

but on the wider picture, this is still valid:
> This module is copied from k/k/pkg/util/procfs to use in Kubernetes components outside of Kubernetes core

so feel free to update the PR if you have the time / wish to continue.
thanks.

bart0sh:
> this PR is no longer needed from the kubeadm POV.

OK, let's close it then. The only purpose of this PR was to help kubeadm.

PR #105, opened 2019-07-23
Add fields to trace to allow for consistent paramaterization and to avoid needless string concatenation

jpbetz:
It's common pattern in k8s to create a trace like so:

```
trace := utiltrace.New("Get " + req.URL.Path)
```

Or

```
trace := utiltrace.New(fmt.Sprintf("GuaranteedUpdate etcd3: %s", getTypeName(out)))
```

Or

```
trace.Step(fmt.Sprintf("Writing http response done (%d items)", meta.LenList(result)))
```

Which results in string concatenation even if the trace is never logged.

This adds optional field args to both `New` and `Step`:

```
trace.New("Get", trace.Field{"URL", req.URL.Path}, trace.Field{"Accepts", req.Header.Accepts})
```

Or

```
trace.Step("Writing http response done", trace.Field{"item-count", meta.LenList(result)})
```

cc @liggitt @roycaihw @sttts

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -25,32 +25,55 @@ import (
 	"k8s.io/klog"
 )
 
+// Field is a key value pair that provides additional details about the trace.
+type Field struct {
+	Key   string
+	Value interface{}
+}
+
+func (f Field) format() string {
+	return fmt.Sprintf("%s:%v", f.Key, f.Value)
+}
+
+func writeFields(b *bytes.Buffer, l []Field) {
+	for i, f := range l {
+		b.WriteString(f.format())
+		if i < len(l)-1 {
+			b.WriteString(",")
+		}
+	}
+}
+
 type traceStep struct {
 	stepTime time.Time
 	msg      string
+	fields   []Field
 }
 
 // Trace keeps track of a set of "steps" and allows us to log a specific
 // step if it took longer than its share of the total allowed time
 type Trace struct {
 	name      string
+	fields    []Field
 	startTime time.Time
 	steps     []traceStep
 }
 
-// New creates a Trace with the specified name
-func New(name string) *Trace {
-	return &Trace{name, time.Now(), nil}
+// New creates a Trace with the specified name. The name identifies the operation to be traced. The
+// Fields add key value pairs to provide additional details about the trace, such as operation inputs.
+func New(name string, fields ...Field) *Trace {
+	return &Trace{name: name, startTime: time.Now(), fields: fields}
 }
 
-// Step adds a new step with a specific message. Call this at the end of an
-// execution step to record how long it took.
-func (t *Trace) Step(msg string) {
+// Step adds a new step with a specific message. Call this at the end of an execution step to record
+// how long it took. The Fields add key value pairs to provide additional details about the trace
+// step.
+func (t *Trace) Step(msg string, fields ...Field) {
 	if t.steps == nil {
 		// traces almost always have less than 6 steps, do this to avoid more than a single allocation
 		t.steps = make([]traceStep, 0, 6)
 	}
-	t.steps = append(t.steps, traceStep{time.Now(), msg})
+	t.steps = append(t.steps, traceStep{stepTime: time.Now(), msg: msg, fields: fields})
 }
 
 // Log is used to dump all the steps in the Trace
@@ -65,12 +88,23 @@ func (t *Trace) logWithStepThreshold(stepThreshold time.Duration) {
 	endTime := time.Now()
 
 	totalTime := endTime.Sub(t.startTime)
-	buffer.WriteString(fmt.Sprintf("Trace[%d]: %q (started: %v) (total time: %v):\n", tracenum, t.name, t.startTime, totalTime))
+	buffer.WriteString(fmt.Sprintf("Trace[%d]: %q ", tracenum, t.name))
+	if len(t.fields) > 0 {
+		writeFields(&buffer, t.fields)
+		buffer.WriteString(" ")
+	}
+	buffer.WriteString(fmt.Sprintf("(started: %v) (total time: %v):\n", t.startTime, totalTime))
 	lastStepTime := t.startTime
 	for _, step := range t.steps {
 		stepDuration := step.stepTime.Sub(lastStepTime)
 		if stepThreshold == 0 || stepDuration > stepThreshold || klog.V(4) {
-			buffer.WriteString(fmt.Sprintf("Trace[%d]: [%v] [%v] %v\n", tracenum, step.stepTime.Sub(t.startTime), stepDuration, step.msg))
+			buffer.WriteString(fmt.Sprintf("Trace[%d]: [%v] [%v] ", tracenum, step.stepTime.Sub(t.startTime), stepDuration))
+			buffer.WriteString(step.msg)
+			if len(step.fields) > 0 {
+				buffer.WriteString(" ")
+				writeFields(&buffer, step.fields)
+			}
+			buffer.WriteString("\n")
 		}
 		lastStepTime = step.stepTime
 	}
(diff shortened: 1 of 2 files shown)


trace/trace.go
```diff
@@ -28,29 +28,43 @@ import (
...
 
 // New creates a Trace with the specified name
 func New(name string) *Trace {
-	return &Trace{name, time.Now(), nil}
+	return &Trace{name: name, startTime: time.Now()}
+}
+
+// Newf creates a Trace with the specified name and formatted msg
+func Newf(name string, msg string, args ...interface{}) *Trace {
+	return &Trace{name: name, startTime: time.Now()}
```
andrewsykim: Should this be setting `msgArgs`?

trace/trace.go
```diff
@@ -28,29 +28,43 @@ import (
...
 	stepTime time.Time
 	msg      string
+	msgArgs  []interface{}
 }
 
 // Trace keeps track of a set of "steps" and allows us to log a specific
 // step if it took longer than its share of the total allowed time
 type Trace struct {
 	name      string
+	msg       string
```
roycaihw: Why do we need `msg` and `msgArgs` in addition to `name`? Can we just have `name` + `nameArgs` instead?

trace/trace.go
```diff
@@ -65,12 +79,19 @@ func (t *Trace) logWithStepThreshold(stepThreshold time.Duration) {
 	endTime := time.Now()
 
 	totalTime := endTime.Sub(t.startTime)
-	buffer.WriteString(fmt.Sprintf("Trace[%d]: %q (started: %v) (total time: %v):\n", tracenum, t.name, t.startTime, totalTime))
+	buffer.WriteString(fmt.Sprintf("Trace[%d]: %q ", tracenum, t.name))
```
roycaihw: I wonder if we can use `strings.Builder` to replace `fmt.Sprintf` https://stackoverflow.com/a/47798475

trace/trace.go
jpbetz: That's one option. Looking at the use cases, having two fields: "name" and formatted "details" seems like it might be a good fit. I've updated the PR. Let me know what you think.

trace/trace.go
jpbetz: I'm a bit concerned with the loss of readability for thins like "Trace[%d]:" if we used a string builder. But I have made heavier use of buffer.WriteString where appropriate.

trace/trace.go
jpbetz: You're right. I've replaced `Newf` with a optional `WithDetails` function which hopefully is clearer. I've fixed this along with that change.

trace/trace_test.go
```diff
@@ -83,35 +83,55 @@ func TestTotalTime(t *testing.T) {
...
+			name: "Check formatting",
+			expectedMessages: []string{
+				"URL=/api,count=3", "msg1 arg1 2 false",
+			},
+			sampleTrace: &Trace{
+				name:        "Sample Trace",
+				details:     "URL=%s,count=%d",
+				detailsArgs: []interface{}{"/api", 3},
+				steps: []traceStep{
+					{stepTime: time.Now(), msg: "msg1 %s %d %t", msgArgs: []interface{}{"arg1", 2, false}},
```
andrewsykim: Probably worth adding a testcase with multiple trace steps to test new line strings in the buffer?

jpbetz:
Added.

trace/trace_test.go
```diff
@@ -83,35 +83,56 @@ func TestTotalTime(t *testing.T) {
...
-		sampleTrace: &Trace{
-			name: "Sample Trace",
-			steps: []traceStep{
-				{time.Now(), "msg1"},
-				{time.Now(), "msg2"},
-				{time.Now(), "msg3"},
+		{
+			name: "Check formatting",
+			expectedMessages: []string{
+				"URL=/api,count=3", "msg1 arg1 2 false", "msg2 x",
```
andrewsykim: Probably for a follow-up PR, should probably test against the entire buffer string instead of matching strings like this.

trace/trace.go
```diff
@@ -28,29 +28,46 @@ import (
 type traceStep struct {
 	stepTime time.Time
 	msg      string
+	msgArgs  []interface{}
```
liggitt: existing uses that include a % in msg and do not populate args will now output malformed sprintf output, right? should we only sprintf if `len(msgArgs) > 0`?

liggitt: either way, needs godoc about what msg is expected to be (literal string? format string? format string if msgArgs is non-zero-length? etc)

trace/trace.go
```diff
@@ -28,29 +28,46 @@ import (
...
 
 // Trace keeps track of a set of "steps" and allows us to log a specific
 // step if it took longer than its share of the total allowed time
 type Trace struct {
-	name      string
-	startTime time.Time
-	steps     []traceStep
+	name        string
+	details     string
+	detailsArgs []interface{}
```
liggitt: same comment about godoc for what `details` is expected to be

trace/trace.go
```diff
@@ -28,29 +28,46 @@ import (
...
+func (t *Trace) WithDetailsf(details string, args ...interface{}) *Trace {
+	t.details = details
+	t.detailsArgs = args
+	return t
 }
 
 // Step adds a new step with a specific message. Call this at the end of an
 // execution step to record how long it took.
 func (t *Trace) Step(msg string) {
+	t.Stepf(msg)
```
liggitt: existing calls here that added step messages containing `%` chars should continue working as-is

trace/trace.go
roycaihw: sounds good to me

trace/trace.go
thockin: Exactly my concern on reading this.  You will have to escape all `%` to `%%` in order to do this, or carry a flag all the way through the lifetime that says whether to do format-string expansion later.

trace/trace.go
jpbetz: Oops. I've fixed by organizing the formatted string more clearly and keeping them separate from the literal strings.

Added a comment as well.

trace/trace.go
jpbetz: Thanks for catching this. I've fixed and added a test to ensure Step() continues to work exactly as before with literal strings.

jpbetz:
> How valuable is the formatting vs simpler key/value pairs? I.e. could you instead do something like:
> 
> ```
> trace := utiltrace.New("Get").With("URL", req.URL.Path).With("accepts", req.Header.Accepts)
> ```
> 
> And then carry a `map[string]interface{}` (or a list of pairs if ordering matters) until you need to concatenate it?
> 
> Similary:
> 
> ```
> trace.Stepf("Writing http response done", n_items, meta.LenList(result), "foo", bar)
> ```
> 
> This is similar in form to https://github.com/go-logr/logr/blob/master/logr.go#L111 and https://github.com/go-logr/logr/blob/master/logr.go#L143 which you can see implemented in https://github.com/go-logr/glogr/blob/master/glogr.go (particularly https://github.com/go-logr/glogr/blob/master/glogr.go#L140)
> 
> You lose some flexibility (oh, I wanted that number in hex) but gain a lot of simplicity. And we can probably add format-wrappers if we needed to, e.g. `...With("foo", trace.AsHex(intValue))`.

I like this idea. We'll need to change how the existing trace calls in k8s are output in the logs to adapt them to this, but it does seem like the right approach. I'll try switching to this approach.

jpbetz:
@liggitt @thockin I've switched to an optional list of key/value fields. I went with a varadic param for the fields, but it could easily be switched to a `.With()` builder style if that's preferable. Let me know what you think.

jpbetz:
Friendly ping

liggitt:
I like the simplicity of the variadic... keeps order, avoids repeatedly appending to a slice. this looks reasonable to me.

wojtek-t:
> I like the simplicity of the variadic... keeps order, avoids repeatedly appending to a slice. this looks reasonable to me.

I like that too.

/lgtm

PR #108, opened 2019-08-13
Changing AtoI to uParseInt to handle overflow problems.

vegemitecheesetoast:
Changing AtoI to uParseInt to handle overflow problems with 16 bit integers. Also removing the check for port range because uParseInt provides us with that now.

### DIFF
--- a/net/net.go
+++ b/net/net.go
@@ -17,6 +17,7 @@ limitations under the License.
 package net
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"strconv"
@@ -136,10 +137,13 @@ func IsIPv6CIDR(cidr *net.IPNet) bool {
 
 // ParsePort parses a string representing an IP port.  If the string is not a
 // valid port number, this returns an error.
-func ParsePort(port string) (int, error) {
+func ParsePort(port string, allowZero bool) (int, error) {
 	portInt, err := strconv.ParseUint(port, 10, 16)
 	if err != nil {
 		return 0, err
+	}
+	if int(portInt) == 0 && !allowZero {
+		errors.New("Port 0 is not allowed, provide non 0 port or make allowZero true")
 	}
 	return int(portInt), nil
 }
--- a/net/net.go
+++ b/net/net.go
@@ -137,12 +137,9 @@ func IsIPv6CIDR(cidr *net.IPNet) bool {
 // ParsePort parses a string representing an IP port.  If the string is not a
 // valid port number, this returns an error.
 func ParsePort(port string) (int, error) {
-	portInt, err := strconv.Atoi(port)
+	portInt, err := strconv.ParseUint(port, 10, 16)
 	if err != nil {
 		return 0, err
 	}
-	if !(0 < portInt && portInt <= 65535) {
-		return 0, fmt.Errorf("%d is not a valid port: must be between 1 and 65535, inclusive", portInt)
-	}
-	return portInt, nil
+	return int(portInt), nil
 }
--- a/net/net.go
+++ b/net/net.go
@@ -142,8 +142,8 @@ func ParsePort(port string, allowZero bool) (int, error) {
 	if err != nil {
 		return 0, err
 	}
-	if int(portInt) == 0 && !allowZero {
-		errors.New("Port 0 is not allowed, provide non 0 port or make allowZero true")
+	if portInt == 0 && !allowZero {
+		return 0, errors.New("0 is not a valid port number")
 	}
 	return int(portInt), nil
 }
--- a/net/net_test.go
+++ b/net/net_test.go
@@ -452,7 +452,7 @@ func TestParsePort(t *testing.T) {
 	var tests = []struct {
 		name          string
 		port          string
-		allowZero	  bool
+		allowZero     bool
 		expectedPort  int
 		expectedError bool
 	}{
@@ -475,7 +475,7 @@ func TestParsePort(t *testing.T) {
 			name:          "invalid port: not a number",
 			port:          "a",
 			expectedError: true,
-			allowZero: false,
+			allowZero:     false,
 		},
 		{
 			name:          "invalid port: too small",
@@ -493,10 +493,9 @@ func TestParsePort(t *testing.T) {
 			expectedError: true,
 		},
 		{
-			name:          "zero port: allowed",
-			port:          "0",
-			allowZero:     true,
-			expectedError: false,
+			name:      "zero port: allowed",
+			port:      "0",
+			allowZero: true,
 		},
 		{
 			name:          "zero port: not allowed",
@@ -507,7 +506,7 @@ func TestParsePort(t *testing.T) {
 
 	for _, rt := range tests {
 		t.Run(rt.name, func(t *testing.T) {
-			actualPort, actualError := ParsePort(rt.port, allowZero)
+			actualPort, actualError := ParsePort(rt.port, rt.allowZero)
 
 			if actualError != nil && !rt.expectedError {
 				t.Errorf("%s unexpected failure: %v", rt.name, actualError)
--- a/net/net_test.go
+++ b/net/net_test.go
@@ -452,6 +452,7 @@ func TestParsePort(t *testing.T) {
 	var tests = []struct {
 		name          string
 		port          string
+		allowZero	  bool
 		expectedPort  int
 		expectedError bool
 	}{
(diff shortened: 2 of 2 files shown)


vegemitecheesetoast:
I signed it

thockin:
Travis is weird, got a 502 pulling a package.  I kicked it

vegemitecheesetoast:
I signed it

vegemitecheesetoast:
I signed it

net/net.go
```diff
@@ -137,12 +138,12 @@ func IsIPv6CIDR(cidr *net.IPNet) bool {
...
 func ParsePort(port string) (int, error) {
-	portInt, err := strconv.Atoi(port)
+	portInt, err := strconv.ParseUint(port, 10, 16)
 	if err != nil {
 		return 0, err
 	}
-	if !(0 < portInt && portInt <= 65535) {
-		return 0, fmt.Errorf("%d is not a valid port: must be between 1 and 65535, inclusive", portInt)
+	if port == 0 {
+		return 0, errors.New("Port number can not be 0")
```
bowei: It might be 0 if you want the system to pick the port for you..

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.12
 
 require (
 	github.com/davecgh/go-spew v1.1.1
+	github.com/gorilla/mux v1.6.3-0.20180517173623-c85619274f5d // indirect
```
bowei: weird why it's pulling this stuff in, your change doesn't reference this.

thockin:
We need to make a call on this as an API - does it allow 0 and expect the
caller to special-case disallow 0, or does it disallow 0 and expect the
caller to test for 0 itself?

Or do we add a flag `allowZero bool` and let callers decide that way?  I
actually don't hate that.

go.mod
vegemitecheesetoast: Tim and I fixed it :)

vegemitecheesetoast:
I have two sided thoughts on this one:

We already have this function I found in kubernetes/

```
// IsValidPortNum tests that the argument is a valid, non-zero port number.
func IsValidPortNum(port int) []string {
	if 1 <= port && port <= 65535 {
		return nil
	}
	return []string{InclusiveRangeError(1, 65535)}
}
```

This above code tests it's a valid port number. My thought is this - the purpose of this function isn't to test if the port number is valid, it's just to convert the port number into a 16 bit integer from a string. bowei@'s point is that we may want the system to choose for us. Do we have a separate function that makes that choice? 

My other thought is that the function is called ParsePort(), so maybe it should check if it's a valid port, else it should be called ParseInt().

The bool is the third option, Not sure how I feel about that.

I'm new to this codebase, so I'm not quite sure what kubernetes style would gravitate towards.

thockin:
`IsValidPortNum()` is *specifically* for API validation, where (TTBOMK)
nobody ever allows 0.

> I'm new to this codebase, so I'm not quite sure what kubernetes style
would gravitate towards.

k/utils is ostensibly our best supported and most general-purpose APIs.  I
don't hate the bool flag.

My instinct is that most callers DO NOT want to allow 0, so if we make this
allow 0, we violate POLS.

Maybe we should audit all the call-sites to see how many do or don't allow
0 ?

vegemitecheesetoast:
i signed it

vegemitecheesetoast:
> k/utils is ostensibly our best supported and most general-purpose APIs.  I
> don't hate the bool flag.

If it's general purpose, bool flag it is! Also added tests.

net/net_test.go
```diff
@@ -490,11 +492,22 @@ func TestParsePort(t *testing.T) {
...
+			name:          "zero port: not allowed",
+			port:          "0",
+			expectedError: true,
+		},
 	}
 
 	for _, rt := range tests {
 		t.Run(rt.name, func(t *testing.T) {
-			actualPort, actualError := ParsePort(rt.port)
+			actualPort, actualError := ParsePort(rt.port, allowZero)
```
thockin: I see this part, but I don'tr see the part that updates the function itself?

vegemitecheesetoast:
i signed it

net/net_test.go
vegemitecheesetoast: I committed it but it's gone. I must have done something. Will add again.

vegemitecheesetoast: can you see it now? :)

net/net.go
```diff
@@ -136,13 +137,13 @@ func IsIPv6CIDR(cidr *net.IPNet) bool {
...
-func ParsePort(port string) (int, error) {
-	portInt, err := strconv.Atoi(port)
+func ParsePort(port string, allowZero bool) (int, error) {
+	portInt, err := strconv.ParseUint(port, 10, 16)
 	if err != nil {
 		return 0, err
 	}
-	if !(0 < portInt && portInt <= 65535) {
-		return 0, fmt.Errorf("%d is not a valid port: must be between 1 and 65535, inclusive", portInt)
+	if int(portInt) == 0 && !allowZero {
```
thockin: should not need the cast to int here?

net/net.go
```diff
@@ -136,13 +137,13 @@ func IsIPv6CIDR(cidr *net.IPNet) bool {
...
+		errors.New("Port 0 is not allowed, provide non 0 port or make allowZero true")
```
thockin: From a UX POV, the error is a little impl specific.  simply "0 is not a valid port number" would suffice, I think.

Also Go convention: errors start lower-case.

net/net_test.go
```diff
@@ -474,6 +475,7 @@ func TestParsePort(t *testing.T) {
 			name:          "invalid port: not a number",
 			port:          "a",
 			expectedError: true,
+			allowZero: false,
```
thockin: don't need to specify it if it doesn't matter to the test case

net/net_test.go
thockin: yes

thockin:
Thanks!

/lgtm
/approve

vegemitecheesetoast:
> The travis failure is not your fault.
> 
> @munnerz I told you changing klog,V would hurt. @vegemitecheesetoast do you have a newer klog installed?
> 
> ```
> # k8s.io/utils/trace
> trace/trace.go:100:57: invalid operation: stepThreshold == 0 || stepDuration > stepThreshold || klog.V(4) (mismatched types bool and klog.Verbose)
> trace/trace.go:112:56: invalid operation: stepThreshold == 0 || stepDuration > stepThreshold || klog.V(4) (mismatched types bool and klog.Verbose)
> ```

This is the version I'm using in go.mod in utils: k8s.io/klog v0.3.0

thockin:
Thanks!

/lgtm
/approve

thockin:
go.mod calls for 0.3.0 but go's tooling can sub a newer version.

What does `go mod why` say?  or `go mod graph` ?

Are you building under a GOPATH or not?

thockin:
oh, I guess it's not you, it's travis :)

thockin:
#111 should fix travis

thockin:
You might have to rebase.

`git fetch upstream`
`git rebase upstream/master`
`git push -f`

thockin:
oh Travis.

You need to `gofmt` the file.  Most of us end up doing that automatically on save (e.g. https://github.com/thockin/dotfiles/blob/master/vimrc)

thockin:
Something like 

`gofmt -s -w net/*.go`
`git commit -a --amend`
`git push -f`

vegemitecheesetoast:
OK, I ran the gofmt and did a git diff and saw that it made some fixes. Hope this is ok!

dims:
@vegemitecheesetoast see the next problem in the travis CI job results

```
net/net.go:146:13: result of errors.New call not used
Makefile:29: recipe for target 'vet' failed
```

vegemitecheesetoast:
@dims 
Sorry it's been so long! since last reply back from me. 
net/net.go:146:13: result of errors.New call not used - fixed - sorry don't know what happened there!
Makefile:29: recipe for target 'vet' failed - I'm going to rebase and see if this fixes, because I didn't touch this, so hoping it solves itself :|

vegemitecheesetoast:
Apparently this travis build now passes. Please let me know if there is anything else I can do, otherwise, do I just wait to get this merged?

dims:
/lgtm
/approve

Thanks for your efforts @vegemitecheesetoast !

PR #110, opened 2019-08-21
move slice package from k/k to the strings package

yastij:
this moves the slice package from `pkg/util` of `k/k` as this pretty self-contained. This is part of a larger effort https://github.com/kubernetes/kubernetes/issues/81172

/assign @thockin @andrewsykim 
/cc @liggitt 


yastij:
sgtm

/close

Issue #109, opened 2019-08-15
mismatched types bool and klog.Verbose


BambooAce:
../utils/trace/trace.go:100:57: invalid operation: stepThreshold == 0 || stepDuration > stepThreshold || klog.V(4) (mismatched types bool and klog.Verbose)
../utils/trace/trace.go:112:56: invalid operation: stepThreshold == 0 || stepDuration > stepThreshold || klog.V(4) (mismatched types bool and klog.Verbose)

maht:
I also hit this issue. I have created a fix: [Pull request #113](https://github.com/kubernetes/utils/pull/113)

maht:
Related issue: https://github.com/kubernetes/client-go/issues/659

dove-young:
I also hit this issue. 

```
[root@ argo-events]# make
CGO_ENABLED=0 GOOS=linux GOARCH=amd64 make sensor
make[1]: Entering directory `/root/go/src/github.com/argoproj/argo-events'
go build -v -ldflags '-X github.com/argoproj/argo-events.version=0.10 -X github.com/argoproj/argo-events.buildDate=2019-10-14T03:49:18Z -X github.com/argoproj/argo-events.gitCommit=3b9f64e2f6b8373f27d37d481c72c5a6087cd118 -X github.com/argoproj/argo-events.gitTreeState=dirty' -o /root/go/src/github.com/argoproj/argo-events/dist/sensor ./sensors/cmd/client.go
k8s.io/utils/trace
# k8s.io/utils/trace
../../../k8s.io/utils/trace/trace.go:100:57: invalid operation: stepThreshold == 0 || stepDuration > stepThreshold || klog.V(4) (mismatched types bool and klog.Verbose)
../../../k8s.io/utils/trace/trace.go:112:56: invalid operation: stepThreshold == 0 || stepDuration > stepThreshold || klog.V(4) (mismatched types bool and klog.Verbose)
k8s.io/client-go/transport
# k8s.io/client-go/transport
../../../k8s.io/client-go/transport/round_trippers.go:70:11: cannot convert klog.V(9) (type klog.Verbose) to type bool
../../../k8s.io/client-go/transport/round_trippers.go:72:11: cannot convert klog.V(8) (type klog.Verbose) to type bool
../../../k8s.io/client-go/transport/round_trippers.go:74:11: cannot convert klog.V(7) (type klog.Verbose) to type bool
[...]

PR #113, opened 2019-09-18
Fix incorrect handling of klog.Verbose

maht:
klog.V() is now returning klog.Verbose instead of bool,
so calls to klog.V() need to update to klog.V(...).Enabled()

Signed-off-by: Miguel Herranz <miguel@midokura.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,5 +6,7 @@ require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/spf13/afero v1.2.2
 	github.com/stretchr/testify v1.3.0
-	k8s.io/klog v1.0.0
+	k8s.io/klog/v2 v2.0.0
 )
+
+replace k8s.io/klog/v2 => k8s.io/klog/v2 v2.0.0-20190919134220-6084f89080e3
--- a/go.mod
+++ b/go.mod
@@ -6,5 +6,5 @@ require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/spf13/afero v1.2.2
 	github.com/stretchr/testify v1.3.0
-	k8s.io/klog v0.3.0
+	k8s.io/klog v1.0.0
 )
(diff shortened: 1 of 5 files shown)


maht:
Related to [issue #109 ](https://github.com/kubernetes/utils/issues/109)

apelisse:
I was wondering how this was working before, but it's actually not working now.

Do you need to also update the version of klog?

maht:
> Do you need to also update the version of klog?

I seems they just released the new one yesterday:

https://github.com/kubernetes/klog/releases/tag/v1.0.0

apelisse:
Sure, but this repo pins to [v0.3.0](https://github.com/kubernetes/utils/blob/master/go.mod#L9), that's why your tests are failing.

maht:
> Sure, but this repo pins to [v0.3.0](https://github.com/kubernetes/utils/blob/master/go.mod#L9), that's why your tests are failing.

Ok, I did not fully understood completely at first. You are right.

But then why did I have the compilation problem on the first place? It seems that current version of this repo should with klog v0.3.0 (and v0.4.0 too), but since I tested as part of building other packages, maybe the go tools promoted to v0.4.0 and beyond automatically, reaching the commit breaking the API of klog.

Now that in klog the have created a v1.0.0, we could just fix the problem by using klog v1.0.0.

Additionally, moving to v2 seems like a good idea, but there is no release still, so not sure if is the right moment to do it.

I had modified the pull request to include 2 commits: one to fix current code to klog v1.0.1, and another to move to v2.

I have created another pull request (https://github.com/kubernetes/utils/pull/114) with only the first commit, in case you want to include that one first, and then include this one after it. 

For reference, this is the current status of tags/code in klog:

```
$ pwd
/home/miguel/go/src/github.com/kubernetes/klog
$ for VERSION in v0.3.0 v0.4.0 v1.0.0^1 v1.0.0 master ; do echo ---$VERSION--- ; git show $VERSION:klog.go | grep "^type Verbose" ; echo ; done
---v0.3.0---
type Verbose bool

---v0.4.0---
type Verbose bool

---v1.0.0^1---
type Verbose struct {

---v1.0.0---
type Verbose bool

---master---
type Verbose struct {

$
```

apelisse:
The last tag that I can see on klog is v1.0.0. Why don't you just pin that version and update the code in the same PR?

maht:
> The last tag that I can see on klog is v1.0.0. Why don't you just pin that version and update the code in the same PR?

Because the v1.0.0 is not API compatible with its current master (eventually becoming v2). So one commit is to just fix the compilation issue fixing the dependency version, and another change is in case master/v2 want to be used.

Maybe it is just better to just remain with v1.0.0 until v2 is released (to avoid new potential changes if master is still unstable) . In that case please ignore/close this pull request and just take in account the other one (https://github.com/kubernetes/utils/pull/114).

apelisse:
> Maybe it is just better to just remain with v1.0.0 until v2 is released (to avoid new potential changes if master is still unstable) .

Yeah, that's more reasonable I think.

PR #116, opened 2019-10-02
Simplify growing ring buffer implementation

matte21:
Simplify growing ring buffer implementation by removing unneeded return argument names and if-else statements.

### DIFF
--- a/buffer/ring_growing.go
+++ b/buffer/ring_growing.go
@@ -34,19 +34,14 @@ func NewRingGrowing(initialSize int) *RingGrowing {
 }
 
 // ReadOne reads (consumes) first item from the buffer if it is available, otherwise returns false.
-func (r *RingGrowing) ReadOne() (data interface{}, ok bool) {
+func (r *RingGrowing) ReadOne() (interface{}, bool) {
 	if r.readable == 0 {
 		return nil, false
 	}
 	r.readable--
 	element := r.data[r.beg]
 	r.data[r.beg] = nil // Remove reference to the object to help GC
-	if r.beg == r.n-1 {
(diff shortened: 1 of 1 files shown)


buffer/ring_growing.go
```diff
@@ -34,19 +34,14 @@ func NewRingGrowing(initialSize int) *RingGrowing {
...
 	r.readable--
 	element := r.data[r.beg]
 	r.data[r.beg] = nil // Remove reference to the object to help GC
-	if r.beg == r.n-1 {
-		// Was the last element
-		r.beg = 0
-	} else {
-		r.beg++
-	}
+	r.beg = (r.beg + 1) % r.n
```
deads2k: This change takes a simple adder operation and makes it a divide.  I wouldn't pay the additional cost to make the code smaller.

buffer/ring_growing.go
```diff
@@ -56,13 +51,8 @@ func (r *RingGrowing) WriteOne(data interface{}) {
...
 		newN := r.n * 2
 		newData := make([]interface{}, newN)
-		to := r.beg + r.readable
-		if to <= r.n {
-			copy(newData, r.data[r.beg:to])
-		} else {
-			copied := copy(newData, r.data[r.beg:])
-			copy(newData[copied:], r.data[:(to%r.n)])
-		}
+		copied := copy(newData, r.data[r.beg:])
```
deads2k: This takes a common case only making a single copy and makes it into a function that always does two copies.  I'll admit those copies don't sometimes do a divide.

I'll take this if a benchmark test that shows for reasonable sizes this new implementation is cheaper.

PR #117, opened 2019-10-10
add bound check to avoid overflowing int64

yastij:

This PR adds boundary checks to avoid overlowing the return value.

ref https://github.com/kubernetes/kubernetes/pull/83422

/assign @dims 
/cc @liggitt

### DIFF
--- a/net/net.go
+++ b/net/net.go
@@ -19,6 +19,7 @@ package net
 import (
 	"errors"
 	"fmt"
+	"math"
 	"math/big"
 	"net"
 	"strconv"
@@ -165,10 +166,15 @@ func AddIPOffset(base *big.Int, offset int) net.IP {
 }
 
 // RangeSize returns the size of a range in valid addresses.
+// returns the size of the subnet (or math.MaxInt64 if the range size would overflow int64)
 func RangeSize(subnet *net.IPNet) int64 {
 	ones, bits := subnet.Mask.Size()
 	if bits == 32 && (bits-ones) >= 31 || bits == 128 && (bits-ones) >= 127 {
 		return 0
+	}
+	// this checks that we are not overflowing an int64
+	if bits-ones >= 63 {
+		return math.MaxInt64
 	}
 	return int64(1) << uint(bits-ones)
 }


net/net.go
```diff
@@ -165,11 +165,14 @@ func AddIPOffset(base *big.Int, offset int) net.IP {
 }
 
 // RangeSize returns the size of a range in valid addresses.
-func RangeSize(subnet *net.IPNet) int64 {
+func RangeSize(subnet *net.IPNet, boundary uint) int64 {
 	ones, bits := subnet.Mask.Size()
 	if bits == 32 && (bits-ones) >= 31 || bits == 128 && (bits-ones) >= 127 {
```
liggitt: my order of operations is rusty... was this line already supposed to protect against (bits-ones) being big enough to cause an overflow?

liggitt: or should we add a condition that returns 0 if (bits-ones) > 63?

net/net.go
```diff
@@ -165,11 +165,14 @@ func AddIPOffset(base *big.Int, offset int) net.IP {
...
 		return 0
 	}
+	if bits-ones >= int(boundary) {
```
liggitt: rather than passing in a boundary, should we just do `if (bits-ones) >= 63 { return 0 }`?

liggitt: or `if (bits-ones) >= 63 { return math.MaxInt64 }`

dims:
/approve

let's wait on @liggitt to lgtm

net/net.go
```diff
@@ -165,11 +165,16 @@ func AddIPOffset(base *big.Int, offset int) net.IP {
...
 // RangeSize returns the size of a range in valid addresses.
+// returns the size of the subnet or max value of int64
 func RangeSize(subnet *net.IPNet) int64 {
 	ones, bits := subnet.Mask.Size()
 	if bits == 32 && (bits-ones) >= 31 || bits == 128 && (bits-ones) >= 127 {
 		return 0
 	}
+	// this checks that we are not overflowing an int64
+	if bits-ones >= 63 {
+		return int64(1) << uint(62)
```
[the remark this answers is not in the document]
liggitt: `1 << 62` != MaxInt64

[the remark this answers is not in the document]
yastij: whoops typo, thanks for catching this

yastij:
@liggitt - updated

liggitt:
looks like verify scripts are still failing

net/net.go
```diff
@@ -165,11 +166,16 @@ func AddIPOffset(base *big.Int, offset int) net.IP {
 }
 
 // RangeSize returns the size of a range in valid addresses.
+// returns the size of the subnet or max value of int64
```
liggitt: need to clarify the circumstances in which MaxInt64 is returned, e.g.:
```suggestion
// returns the size of the subnet (or math.MaxInt64 if the range size would overflow int64)
```

liggitt:
I think this is the most reasonable thing we can do in an overflow case without changing the method signature.

/lgtm

PR #118, opened 2019-10-10
Add command and arg order checking to fakeexec

codenrhoden:
This patch introduces the ability to enable strict command and arg order
checking to the FakeExec struct. When enabled, if any commands or args
are called in a different order or with different content than exactly
what was scripted, a panic is thrown.

Additionally, add the ability to not require scripting commands at all,
thereby allowing any calls to return successfully w/o error.

The motivation behind this is trying to use `k8s.io/utils/exec` as a replacement for `k8s.io/kubernetes/pkg/util/mount.Exec`. There is an existing `Exec` interface and implementation of that interface in the k/k mount code that could just be using these utils instead. However, in k/k there is also a "fake mount" implementation that is heavily used to enforce that commands are called in an exact order and with exact arguments. The fake exec found here did not have a way to check that. In fact, it's really only checking that the same *number* of scripted commands are called, not caring what those commands are.

For example, you could script the fake exec with two commands: "cat file1" and "cat file2", then use the fakeexec to call commands "sudo -i" and "reboot" and it would be happy so long as you didn't make a 3rd call to it. The new `ExactOrder` field in this patch takes care of that -- it is off by default for backwards compatibility.

### DIFF
--- a/exec/testing/fake_exec.go
+++ b/exec/testing/fake_exec.go
@@ -29,6 +29,14 @@ type FakeExec struct {
 	CommandScript []FakeCommandAction
 	CommandCalls  int
 	LookPathFunc  func(string) (string, error)
+	// ExactOrder enforces that commands are called in the order they are scripted,
+	// and with the exact same arguments
+	ExactOrder bool
+	// DisableScripts removes the requirement that a slice of FakeCommandAction be
+	// propulated before calling Command(). This makes the fakeexec (and subsequent
+	// calls to Run() or CombinedOutput() always return success and there is no
+	// ability to set their output.
+	DisableScripts bool
 }
 
 var _ exec.Interface = &FakeExec{}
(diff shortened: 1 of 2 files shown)


codenrhoden:
/assign @dims @thockin @msau42 
^Michelle, this is the compatibility I was telling you we needed in order to maintain the same level of testing that is currently found in the mount and volume packages.

exec/testing/fake_exec.go
```diff
@@ -43,7 +46,23 @@ func (fake *FakeExec) Command(cmd string, args ...string) exec.Cmd {
 	}
 	i := fake.CommandCalls
 	fake.CommandCalls++
-	return fake.CommandScript[i](cmd, args...)
+	fakeCmd := fake.CommandScript[i](cmd, args...)
+	if fake.ExactOrder {
```
thockin: Is there ever a time we don't want this?

codenrhoden: I can't really think of one, but I didn't want to suddenly break a bunch of people's tests by turning it on by default. As I am trying to have code in k/k use this package instead of the one found `k/k/pkg/util/mount`, I am already seeing that I am going to have to update a lot of tests because any calls to this fake exec without scripting those calls first results in the error: `panic: ran out of Command() actions. Could not handle command`

With this test module you are *required* to pre-load/script every call that's going to be made. Currently, it doesn't actually care what those calls are, just that there is the right number of them. Or you make less than what was scripted. :) Most of the code in k/k isn't doing that right now. I only found two places (linked below) that are using the `mount.Exec` callback to implement detailed command call verification. Most users just need a fake Exec that satisfies the API. So even without this change, there is going to be a lot of changes to tests to preload the calls that are made. Whether we should make it so that we also *always* check that those commands are called in the exact order given, with the exact arguments? Good question. I was leaving this feature off because it was a major change to how it works and seems like it broke backwards compat. I'll defer to advice from others. Thanks for review @thockin .

Examples of where no checking is done:
[1](https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/fc/fc_test.go#L167) [2](https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/testing/testing.go#L113)

And examples where it *is* done... 
[1](https://github.com/kubernetes/kubernetes/blob/faad5d52bcefa903a0adf4b1dea9a9413944ca33/pkg/util/mount/safe_format_and_mount_test.go#L193-L209) [2](https://github.com/kubernetes/kubernetes/blob/faad5d52bcefa903a0adf4b1dea9a9413944ca33/pkg/volume/iscsi/iscsi_util_test.go#L220-L240)

Note that in these examples, I'm trying to replace the code in k/k to use this library instead. The examples that do check cannot be moved without this equivalent feature without losing testing fidelity.

codenrhoden: A couple more thoughts on this...

Yes there are times where we won't want to validate arguments. Sometimes the arguments are not known ahead of time, because they reference tempdirs. One could make an argument for verifying the command, the command order, but maybe not the args? That would introduce a lot of knobs to turn if you did on a per-command basis.

I also think we need the ability to disable the current "script counting" -- meaning that this fake exec panics if you call a command and it hasn't already been scripted. I'm finding dozens of tests in k/k that just want a fake exec to satisfy an API and updating them all to script them ahead of time is going to be a massive amount of work.

codenrhoden:
As I mentioned in https://github.com/kubernetes/utils/pull/118#discussion_r334098836, trying to use the testing package in k/k without the ability to disable the pre-scripting is an enormous amount of work. I've added a flag to introduce this capability.

I believe this will allow us to preserve the same type of testing that is currently happening in k/k if we use the package instead.

msau42:
/assign @misterikkit 
who has graciously agreed to help review all the mount refactoring work

codenrhoden:
@misterikkit @thockin Can I get eyes on this again?

PR #120, opened 2019-11-14
Add ability to skip golint failures per package

codenrhoden:
This patch modifies hack/verify-golint.sh to be like the one found in
k/k, which reads hack/.golint_failures for a list of packages on which
to ignore golint failures.

Needed in order to pull the mount package in under #100 

/assign @liggitt @dims

I tested that with this in place, if I modified an existing package to fail lint, it still got picked up -- then when I added it to `.golint_failures`, it passed.

`/hack/lib/util.sh` was a straight copy from k/k.

### DIFF
--- a/hack/.golint_failures
+++ b/hack/.golint_failures
@@ -1,0 +1,1 @@
+# Apart from this line, only trailing comments are supported
--- a/hack/lib/util.sh
+++ b/hack/lib/util.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+# Copyright 2014 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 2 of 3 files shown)


dims:
@codenrhoden let's strip down the big file to just what we need here namely and we can merge this
```
kube::util::array_contains
kube::util::check-file-in-alphabetical-order
```

codenrhoden:
@dims, Okay, I can do that...

I have to apologize for shenanigans I'm about to pull. :) The Travis-CI failure is happening because the bash version it's running is too old. It *only* shows up if `hack/.golint_failures` is empty, which it is in this PR.

So I've made a temp change to the script that will make it work, then I will undo that change when we add `mount` to this repo. I tried for a good about of time to make it worth with both old and new versions of bash, but it's not my area and I don't have that time to lose right now.

When I submitted this PR, it was passing on my local machine, but that's because I'm using bash >5. When I manually ran it with an older bash, it fails. :( Anyways, won't be an issue for long.

PR #122, opened 2019-11-18
Update GetDiskFormat() base on new version blkid

FengyunPan2:
Related to: https://github.com/kubernetes/kubernetes/issues/85203
This pr could get disk format base on new version blkid, and it
work fine on old version blkid.

### DIFF
--- a/mount/mount_linux.go
+++ b/mount/mount_linux.go
@@ -362,24 +362,37 @@ func (mounter *SafeFormatAndMount) GetDiskFormat(disk string) (string, error) {
 		return "", err
 	}
 
+	fstype := getDiskType(disk, output, "\n")
+	if len(fstype) == 0 {
+		fstype = getDiskType(disk, output, " ")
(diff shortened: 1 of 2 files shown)


FengyunPan2:
any thought ? @jingxu97

joelrebel:
Any chance this fix could be considered any since this affects alpine linux images with the busybox `blkid` , which does not return its output separated with newlines.

Before I figured this was already reported, I ended up implementing a fix in my fork, https://github.com/joelrebel/utils/commit/4694e926f6fee892e489604157b23bc5d60e2c14

jparklab:
@joelrebel I experienced the same issue with alpine linux. However, this is a bug of the blkid executable in the alphine linux. it does seem to take any options.
e.g. inside an alphine linux image 'blkid -h' does not print help messages. 

I filed a ticket to alpine linux: https://gitlab.alpinelinux.org/alpine/aports/issues/11404

PR #123, opened 2019-11-21
feature: implement Output method for FakeCmd

SataQiu:
`Output` method is useful in some scenarios, and this PR implements it.
Ref: kubernetes/kubeadm#1925

### DIFF
--- a/exec/testing/fake_exec.go
+++ b/exec/testing/fake_exec.go
@@ -87,10 +87,13 @@ func (fake *FakeExec) LookPath(file string) (string, error) {
 // FakeCmd is a simple scripted Cmd type.
 type FakeCmd struct {
 	Argv                 []string
-	CombinedOutputScript []FakeCombinedOutputAction
+	CombinedOutputScript []FakeAction
 	CombinedOutputCalls  int
 	CombinedOutputLog    [][]string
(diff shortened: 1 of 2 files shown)


apelisse:
I'm not strictly opposed to this change since I understand it is correct, BUT, I don't know if this is the best way to do it. Typical tests shouldn't really care if "Output", "Run" or "CombinedOutput" is used, but that the behavior is correct. I suspect that we should have a common "FakeAction" for these three commands to simulate the a running executable.

SataQiu:
Yes, I agree with you @apelisse . 
We really should have a common "FakeAction" for these three commands.

But my concern is that this is a breaking change, there's already a lot of use of "FakeCombinedOutputAction" or "FakeRunAction" in the existing code. In order not to impact existing tests, perhaps the current practice is acceptable. In the long term, we should refactor this code, perhaps in the next release?  For now, I think we should first implement the `Out` method to be consistent with the real `Cmd`.

SataQiu:
/assign @thockin @mengqiy
Waitting for more comments(suggestion)! 
It blocks some other PRs.

SataQiu:
Using a common "FakeAction" for these three commands. @apelisse @thockin

PR #124, opened 2019-12-02
Add test for inotify fd leak fix

### LINKED ISSUE #106: Need a test case for PR #101
We have a manual test here:
https://github.com/kubernetes/kubernetes/pull/80689#issuecomment-518667205

We need to automate that.

avorima:
Fixes #106

### DIFF
--- a/inotify/inotify_linux_test.go
+++ b/inotify/inotify_linux_test.go
@@ -7,9 +7,13 @@
 package inotify
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
+	"os/exec"
+	"reflect"
 	"sync/atomic"
+	"syscall"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


inotify/inotify_linux_test.go
```diff
@@ -105,3 +109,31 @@ func TestInotifyClose(t *testing.T) {
 		t.Fatal("expected error on Watch() after Close(), got nil")
 	}
 }
+
+func TestInotifyFdLeak(t *testing.T) {
+	watcher, _ := NewWatcher()
+	defer watcher.Close()
+
+	child := exec.Command("sleep", "1")
```
dims: is `"`1`" enough? we could bump it up as we are calling SIGTERM right?

avorima: The test run time increase is barely noticeable, but if I can bump the value to be safe. Would 10 be ok?

avorima:
Ok, so I've tried a few things but all the approaches that worked locally failed in the pipeline.

dims:
/close

thanks @avorima

Issue #125, opened 2019-12-03
fsck is not a suitable command to check and repair an XFS filesystem ,use xfs_repair instead


27149chen:
In method `formatAndMount` in file utils/mount/mount_linux.go, command `fsck -a` is used to check and repair all kind of filesystems, including xfs.
But in fact `fsck` does nothing for xfs, we should use `xfs_repair` instead.

See the references:
https://linux.die.net/man/8/xfs_repair
https://linux.die.net/man/8/fsck.xfs

27149chen:
We fixed it in our code by wrapping the Mounter struct, but it is not convenient, so I wonder if you can accept a PR for that?

27149chen:
it is fixed and merged in #126

liggitt:
/reopen

since #126 was reverted

liggitt:
closing since the xfs discussion revealed this is not actually an issue

/close

PR #126, opened 2019-12-06
use xfs_repair to check and repair xfs filesystem in method formatAndMount

27149chen:
This PR tries to fix the issue mentioned in https://github.com/kubernetes/utils/issues/125, changes are as following:
1. move the general check and repair method (using fsck) to an individual function.
2. add a new function "checkAndRepairXfsFilesystem" to check and repair xfs filesystem.
3. in method formatAndMount, add a switch to use different functions for different filesystems, and only run check tool on formatted disks.
4. fix some test cases and add add somre more.

Which are not included in this PR:
1. A clean log on a file system is required for xfs_repair to operate. If the file system was not cleanly unmounted, it should be mounted and unmounted prior to using xfs_repair

References:
1. http://man7.org/linux/man-pages/man8/xfs_repair.8.html
2. https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/storage_administration_guide/fsck-fs-specific#fsck-xfs

### DIFF
--- a/mount/mount_linux.go
+++ b/mount/mount_linux.go
@@ -256,6 +256,54 @@ func (mounter *Mounter) GetMountRefs(pathname string) ([]string, error) {
 	return SearchMountPoints(realpath, procMountInfoPath)
 }
 
+// checkAndRepairFileSystem checks and repairs filesystems using command fsck.
+func (mounter *SafeFormatAndMount) checkAndRepairFilesystem(source string) error {
+	klog.V(4).Infof("Checking for issues with fsck on disk: %s", source)
+	args := []string{"-a", source}
+	out, err := mounter.Exec.Command("fsck", args...).CombinedOutput()
+	if err != nil {
+		ee, isExitError := err.(utilexec.ExitError)
+		switch {
+		case err == utilexec.ErrExecutableNotFound:
+			klog.Warningf("'fsck' not found on system; continuing mount without running 'fsck'.")
+		case isExitError && ee.ExitStatus() == fsckErrorsCorrected:
+			klog.Infof("Device %s has errors which were corrected by fsck.", source)
+		case isExitError && ee.ExitStatus() == fsckErrorsUncorrected:
+			return NewMountError(HasFilesystemErrors, "'fsck' found errors on device %s but could not correct them: %s", source, string(out))
+		case isExitError && ee.ExitStatus() > fsckErrorsUncorrected:
+			klog.Infof("`fsck` error %s", string(out))
+		}
+	}
+	return nil
+}
+
+// checkAndRepairXfsFilesystem checks and repairs xfs filesystem using command xfs_repair.
+func (mounter *SafeFormatAndMount) checkAndRepairXfsFilesystem(source string) error {
+	klog.V(4).Infof("Checking for issues with xfs_repair on disk: %s", source)
+
+	args := []string{source}
+	checkArgs := []string{"-n", source}
+
+	// check-only using "xfs_repair -n", if the exit status is not 0, perform a "xfs_repair"
+	_, err := mounter.Exec.Command("xfs_repair", checkArgs...).CombinedOutput()
+	if err != nil {
+		if err == utilexec.ErrExecutableNotFound {
+			klog.Warningf("'xfs_repair' not found on system; continuing mount without running 'xfs_repair'.")
+			return nil
+		} else {
+			klog.Warningf("Filesystem corruption was detected for %s, running xfs_repair to repair", source)
+			out, err := mounter.Exec.Command("xfs_repair", args...).CombinedOutput()
+			if err != nil {
+				return fmt.Errorf("'xfs_repair' found errors on device %s but could not correct them: %s\n", source, out)
+			} else {
+				klog.Infof("Device %s has errors which were corrected by xfs_repair.", source)
+				return nil
+			}
+		}
+	}
+	return nil
+}
+
 // formatAndMount uses unix utils to format and mount the given disk
 func (mounter *SafeFormatAndMount) formatAndMount(source string, target string, fstype string, options []string) error {
 	readOnly := false
@@ -268,26 +316,6 @@ func (mounter *SafeFormatAndMount) formatAndMount(source string, target string, 
 
 	options = append(options, "defaults")
 	var mountErrorValue MountErrorType
-
-	if !readOnly {
-		// Run fsck on the disk to fix repairable issues, only do this for volumes requested as rw.
-		klog.V(4).Infof("Checking for issues with fsck on disk: %s", source)
-		args := []string{"-a", source}
-		out, err := mounter.Exec.Command("fsck", args...).CombinedOutput()
-		if err != nil {
-			ee, isExitError := err.(utilexec.ExitError)
-			switch {
-			case err == utilexec.ErrExecutableNotFound:
-				klog.Warningf("'fsck' not found on system; continuing mount without running 'fsck'.")
-			case isExitError && ee.ExitStatus() == fsckErrorsCorrected:
-				klog.Infof("Device %s has errors which were corrected by fsck.", source)
-			case isExitError && ee.ExitStatus() == fsckErrorsUncorrected:
-				return NewMountError(HasFilesystemErrors, "'fsck' found errors on device %s but could not correct them: %s", source, string(out))
-			case isExitError && ee.ExitStatus() > fsckErrorsUncorrected:
-				klog.Infof("`fsck` error %s", string(out))
-			}
-		}
-	}
 
 	// Check if the disk is already formatted
 	existingFormat, err := mounter.GetDiskFormat(source)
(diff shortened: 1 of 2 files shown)


27149chen:
@jsafrane , do you have time to have a look of this PR?

mount/mount_linux.go
```diff
@@ -257,6 +258,100 @@ func (mounter *Mounter) GetMountRefs(pathname string) ([]string, error) {
...
+
+// checkAndRepairFileSystem checks and repairs xfs filesystem using command xfs_repair.
+func (mounter *SafeFormatAndMount) checkAndRepairXfsFilesystem(source string) error {
+	klog.V(4).Infof("Checking for issues with xfs_repair on disk: %s", source)
+
+	args := []string{source}
+	checkArgs := []string{"-n", source}
+
+	// check-only using "xfs_repair -n", if the exit status is not 0, perform a "xfs_repair"
+	out, err := mounter.Exec.Command("xfs_repair", checkArgs...).CombinedOutput()
```
jsafrane: `xfs_repair -n`  on an empty device tries to find superblock on the whole device, taking really long time.

```
$ xfs_repair -n /dev/loop0
Phase 1 - find and verify superblock...
bad primary superblock - bad magic number !!!

attempting to find secondary superblock...
..........................................................................................................................................................................................................................................................................................................................................................................................................................................................
```

jsafrane: In this case, running `blkid` first to check the format and `fsck` / `xfs_repair` only when the right filesystem is detected would be better.

See https://github.com/kubernetes/kubernetes/pull/77982 for such attempt.

mount/safe_format_and_mount_test.go
```diff
@@ -181,12 +181,36 @@ func TestSafeFormatAndMount(t *testing.T) {
...
 			expectedError: fmt.Errorf("exit 4"),
 		},
+		{
+			description: "Test that 'xfs_repair' is called twice and repair the filesystem",
+			fstype:      "xfs",
+			mountErrs:   []error{fmt.Errorf("unknown filesystem type '(null)'"), nil},
+			execScripts: []ExecArgs{
+				{"xfs_repair", []string{"-n", "/dev/foo"}, "", &testingexec.FakeExitError{Status: 1}},
+				{"xfs_repair", []string{"/dev/foo"}, "\ndone\n", nil},
+				{"blkid", []string{"-p", "-s", "TYPE", "-s", "PTTYPE", "-o", "export", "/dev/foo"}, "", &testingexec.FakeExitError{Status: 2}},
```
jsafrane: I'd prefer to test case when `blkid` and `mount` succeeds - it's odd to fix xfs only to call `mkfs.xfs` later

mount/mount_linux.go
27149chen: @jsafrane , thank you for pointing out it. But if I add a `GetDiskFormat` here, I think the whole mount flow will be changed, like the order in  kubernetes/kubernetes#77982.
So what is your suggestion here? merge 77982 into my change as one pr, or let 77982 merge first, and continue my work?

mount/safe_format_and_mount_test.go
27149chen: ok, I will add a new one when the above comment has a conclusion

mount/mount_linux.go
jsafrane: I would suggest to contact original author if he's still interested in the PR and re-post it in this repo. With a concrete use case where we really need to call blkid before mount it should be easy to merge.

27149chen: ok, thanks. I asked him in that pr.

27149chen: fixed, run fsck/xfs_repair only when the device is formatted.

27149chen:
@jsafrane , I merged the latest master changes and update the code, please have a look again, thank you very much.

mount/mount_linux.go
```diff
@@ -256,6 +257,100 @@ func (mounter *Mounter) GetMountRefs(pathname string) ([]string, error) {
...
+	if err != nil {
+		if err == utilexec.ErrExecutableNotFound {
+			klog.Warningf("'xfs_repair' not found on system; continuing mount without running 'xfs_repair'.")
+			return nil
+		} else {
+			klog.V(4).Infof("Filesystem corruption was detected for %s, running xfs_repair to repair", source)
```
gnufied: Shouldn't this be logged at higher log level?

mount/mount_linux.go
```diff
@@ -256,6 +257,100 @@ func (mounter *Mounter) GetMountRefs(pathname string) ([]string, error) {
...
+//
+// If xfs_repair is not able to repair the filesystem at a certain phase, it
+// will print some error messages and exit.
+func analyzeXfsRepairOutput(output []byte) error {
+	trimed := bytes.TrimSpace(output)
+	if bytes.HasSuffix(trimed, []byte("done")) {
+		return nil
+	} else {
+		return fmt.Errorf(string(output))
+	}
```
gnufied: I wonder if this could result in breaking existing stuff. I am unsure about searching for "done" string at the end of the message and using combined output from `xfs_repair`. We have been burnt by such checks in past and command output itself may change across versions.

I think may be we should allow different CSI driver authors to opt-in for doing filesystem repair before mounting. We did not call `xfs_repair` before and now we do and we are performing checks based on output of command(which can spew random stuff and provides no guarantees about API) - could this cause breaking behaviour?

27149chen: @gnufied , thank you for pointing it out. I checked several versions of `xfs_repair`, the output format is stable, so I think it is safe to search for "done" here. And I think it won't break existing stuff. Regarding "We have been burnt by such checks in past", do you remember the code and how did you fix it? can you show me so I can learn something.

I don't think we should let the CSI plugins to do the fs check. Because it is a common task, each type of filesystem has a specific check/repair method. Another reason is that `xfs_repair` must be run in a formatted disk, if we let the CSI plugins to run it, they must run `blkid` first, which is redundant.

Yes, for a xfs filesystem, We did not call `xfs_repair` before and now we do. But it won't break anything:
1. If `xfs_repair` is called before, the fs is repaired before enter `mount`,so `xfs_repair -n` will return ok and nothing will be repaireed again.
2. If `xfs_repair` is not called before, we will try to repair it before mount, it will make the `mount` method more robust.

What do you think?

gnufied: It is okay to rely on following outputs of a program called via exec:

1. exit code
2. if program supports some kind of json or other machine readable output.

It is not okay to rely on human readable text printed on terminal for consumption by another program. Because in that case - called program(`xfs_repair` in this case) makes no guarantees about an stable format.

27149chen: @gnufied , after checking the source code `git://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git`, I find that xfs_repair do return a non-zero status if it is failed to repair the filesystem.
The code is in `repair/xfs_repair.c`.
In fact, the problem is the man page, it is wrong until v4.10.0, you can see the changelog of v4.10.0: `https://abi-laboratory.pro/?view=changelog&l=xfsprogs&v=5.0.0` which fixed the description.

So we don't need to analysis the human readable output here, just check the return status

27149chen: attached the diff here to make it more clear.
```
$ git diff v4.9.0 v4.10.0 man/man8/xfs_repair.8
diff --git a/man/man8/xfs_repair.8 b/man/man8/xfs_repair.8
index 1b4d9e3..85e4dc9 100644
--- a/man/man8/xfs_repair.8
+++ b/man/man8/xfs_repair.8
@@ -65,7 +65,9 @@ Forces
 .B xfs_repair
 to zero the log even if it is dirty (contains metadata changes).
 When using this option the filesystem will likely appear to be corrupt,
-and can cause the loss of user files and/or data.
+and can cause the loss of user files and/or data.  See the
+.B "DIRTY LOGS"
+section for more information.
 .TP
 .BI \-l " logdev"
 Specifies the device special file where the filesystem's external
@@ -505,11 +507,38 @@ This message refers to a large directory.
 If the directory were small, the message would read "junking entry ...".
 .SH EXIT STATUS
 .B xfs_repair \-n
-(no modify node)
+(no modify mode)
 will return a status of 1 if filesystem corruption was detected and
 0 if no filesystem corruption was detected.
 .B xfs_repair
-run without the \-n option will always return a status code of 0.
+run without the \-n option will always return a status code of 0 if
+it completes without problems.  If a runtime error is encountered
+during operation, it will return a status of 1.  In this case,
+.B xfs_repair
+should be restarted.  If
+.B xfs_repair is unable
+to proceed due to a dirty log, it will return a status of 2.  See below.
+.SH DIRTY LOGS
+Due to the design of the XFS log, a dirty log can only be replayed
+by the kernel, on a machine having the same CPU architecture as the
+machine which was writing to the log.
+.B xfs_repair
+cannot replay a dirty log and will exit with a status code of 2
+when it detects a dirty log.
+.PP
+In this situation, the log can be replayed by mounting and immediately
+unmounting the filesystem on the same class of machine that crashed.
+Please make sure that the machine's hardware is reliable before
+replaying to avoid compounding the problems.
+.PP
+If mounting fails, the log can be erased by running
+.B xfs_repair
+with the -L option.
+All metadata updates in progress at the time of the crash will be lost,
+which may cause significant filesystem damage.
+This should
+.B only
+be used as a last resort.
 .SH BUGS
 The filesystem to be checked and repaired must have been
 unmounted cleanly using normal system administration procedures
```

27149chen:
@jsafrane , please take another look at the pr, thanks.

jsafrane:
lgtm-ish, can you please rebase?
/approve

27149chen:
@jsafrane , finished the squash and rebase, please approve.

mount/mount_linux.go
```diff
@@ -324,10 +352,27 @@ func (mounter *SafeFormatAndMount) formatAndMount(source string, target string,
...
+			var err error
+			switch existingFormat {
+			case "xfs":
+				err = mounter.checkAndRepairXfsFilesystem(source)
+			default:
+				err = mounter.checkAndRepairFilesystem(source)
+			}
+
+			if err != nil {
+				return err
```
jsafrane: Rebase is not that simple in this case: you should return `NewMountError(mountErrorValue, err.Error())`, to propagate potential `FilesystemMismatch`.

jsafrane: Actually, you should return `FilesystemMismatch`, if the FS type is wrong, or `HasFilesystemErrors` if the FS type is OK, but the filesystem has errors.

27149chen: 1. HasFilesystemErrors is retruned in checkAndRepairXfsFilesystem/checkAndRepairFilesystem, so it is ok to return err here directly.
2. I don't think we should return a FilesystemMismatch here, because the only error which will happen here is a failed_to_repair_filesystem_error, and the fs passed in is the existing fs, not the declear one, so I think there is no filesystem mismatch which can cause this error.
3. I forgot to return a HasFilesystemErrors in checkAndRepairXfsFilesystem, which is fixed now.

jsafrane: ack

jsafrane: ack, I missed checkAndRepairXfsFilesystem/checkAndRepairFilesystem

PR #127, opened 2019-12-13
Validate the existence of filesystem before attempting to mount it (linux)

mvisonneau:
This is a reopen of https://github.com/kubernetes/kubernetes/pull/77982, necessary for https://github.com/kubernetes/utils/pull/126

The current workflow of the mount on linux tries to mount the disk before assessing if a FS actually
exists onto it. This commit review this implementation. Here is what would happen now:
- Validate that the disk is formatted
  - true : mount it
  - false : format it then mount it

This also ensures that the disk FS type is the one we are expecting before attempting to mount it.

What would be nice is to also add some tests for the format and mount functions.

**What type of PR is this?**

/kind bug

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

### DIFF
--- a/mount/mount_linux.go
+++ b/mount/mount_linux.go
@@ -19,7 +19,6 @@ limitations under the License.
 package mount
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -289,55 +288,53 @@ func (mounter *SafeFormatAndMount) formatAndMount(source string, target string, 
 		}
 	}
 
-	// Try to mount the disk
-	klog.V(4).Infof("Attempting to mount disk: %s %s %s", fstype, source, target)
-	mountErr := mounter.Interface.Mount(source, target, fstype, options)
-	if mountErr != nil {
-		// Mount failed. This indicates either that the disk is unformatted or
-		// it contains an unexpected filesystem.
-		existingFormat, err := mounter.GetDiskFormat(source)
-		if err != nil {
-			return err
+	// Check if the disk is already formatted
+	existingFormat, err := mounter.GetDiskFormat(source)
+	if err != nil {
+		return err
+	}
+
+	// Use 'ext4' as the default
+	if len(fstype) == 0 {
+		fstype = "ext4"
+	}
+
+	if existingFormat == "" {
+		// Do not attempt to format the disk if mounting as readonly, return an error to reflect this.
+		if readOnly {
+			return fmt.Errorf("cannot mount unformatted disk %s as we are manipulating it in read-only mode", source)
 		}
-		if existingFormat == "" {
-			if readOnly {
-				// Don't attempt to format if mounting as readonly, return an error to reflect this.
-				return errors.New("failed to mount unformatted volume as read only")
-			}
 
-			// Disk is unformatted so format it.
-			args := []string{source}
-			// Use 'ext4' as the default
-			if len(fstype) == 0 {
-				fstype = "ext4"
+		// Disk is unformatted so format it.
+		args := []string{source}
+		if fstype == "ext4" || fstype == "ext3" {
+			args = []string{
+				"-F",  // Force flag
+				"-m0", // Zero blocks reserved for super-user
+				source,
 			}
+		}
 
-			if fstype == "ext4" || fstype == "ext3" {
-				args = []string{
-					"-F",  // Force flag
-					"-m0", // Zero blocks reserved for super-user
-					source,
-				}
-			}
-			klog.Infof("Disk %q appears to be unformatted, attempting to format as type: %q with options: %v", source, fstype, args)
-			_, err := mounter.Exec.Command("mkfs."+fstype, args...).CombinedOutput()
-			if err == nil {
-				// the disk has been formatted successfully try to mount it again.
-				klog.Infof("Disk successfully formatted (mkfs): %s - %s %s", fstype, source, target)
-				return mounter.Interface.Mount(source, target, fstype, options)
-			}
+		klog.Infof("Disk %q appears to be unformatted, attempting to format as type: %q with options: %v", source, fstype, args)
+		_, err := mounter.Exec.Command("mkfs."+fstype, args...).CombinedOutput()
+		if err != nil {
 			klog.Errorf("format of disk %q failed: type:(%q) target:(%q) options:(%q)error:(%v)", source, fstype, target, options, err)
 			return err
 		}
-		// Disk is already formatted and failed to mount
-		if len(fstype) == 0 || fstype == existingFormat {
-			// This is mount error
-			return mountErr
-		}
-		// Block device is formatted with unexpected filesystem, let the user know
-		return fmt.Errorf("failed to mount the volume as %q, it already contains %s. Mount error: %v", fstype, existingFormat, mountErr)
+
+		klog.Infof("Disk successfully formatted (mkfs): %s - %s %s", fstype, source, target)
+	} else if fstype != existingFormat {
+		// Verify that the disk is formatted with filesystem type we are expecting
+		klog.Warningf("Configured to mount disk %s as %s but current format is %s, things might break", source, existingFormat, fstype)
 	}
-	return mountErr
+
+	// Mount the disk
+	klog.V(4).Infof("Attempting to mount disk %s in %s format at %s", source, fstype, target)
+	if err := mounter.Interface.Mount(source, target, fstype, options); err != nil {
+		return err
+	}
+
+	return nil
 }
 
 // GetDiskFormat uses 'blkid' to see if the given disk is unformatted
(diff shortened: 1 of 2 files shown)


mvisonneau:
I added a couple of tests and fixed some others 👌

mount/mount_linux.go
```diff
@@ -289,55 +288,59 @@ func (mounter *SafeFormatAndMount) formatAndMount(source string, target string,
...
-		return fmt.Errorf("failed to mount the volume as %q, it already contains %s. Mount error: %v", fstype, existingFormat, mountErr)
+
+		klog.Errorf("format of disk %q failed: type:(%q) target:(%q) options:(%q)error:(%v)", source, fstype, target, options, err)
+		return err
+	} else {
+		klog.V(4).Infof("%s is already formatted (%s)", source, existingFormat)
+	}
+
+	// Verify that the disk is formatted with filesystem type we are expecting
+	if fstype != existingFormat {
```
27149chen: I think moving it to above "else" will be better.
```
if existingFormat == "" {
    ...
} else if existingFormat != fstype {
    ...
}
```
And also, setting fstype's default value before this if-else, so that it always has a defalut value.

mount/mount_linux.go
```diff
@@ -289,55 +288,59 @@ func (mounter *SafeFormatAndMount) formatAndMount(source string, target string,
...
-		if len(fstype) == 0 || fstype == existingFormat {
-			// This is mount error
-			return mountErr
+
+		klog.Infof("Disk %q appears to be unformatted, attempting to format as type: %q with options: %v", source, fstype, args)
+		_, err := mounter.Exec.Command("mkfs."+fstype, args...).CombinedOutput()
+		if err == nil {
+			// the disk has been formatted successfully try to mount it again.
+			klog.Infof("Disk successfully formatted (mkfs): %s - %s %s", fstype, source, target)
+			return mounter.Interface.Mount(source, target, fstype, options)
```
27149chen: don't need mount and return here, it is a bit redundant, when you exit this if-else, the remaining code will mount and return no matter it is formatted or not.

mount/mount_linux.go
```diff
@@ -289,55 +288,59 @@ func (mounter *SafeFormatAndMount) formatAndMount(source string, target string,
...
 		}
-		// Block device is formatted with unexpected filesystem, let the user know
-		return fmt.Errorf("failed to mount the volume as %q, it already contains %s. Mount error: %v", fstype, existingFormat, mountErr)
+
+		klog.Errorf("format of disk %q failed: type:(%q) target:(%q) options:(%q)error:(%v)", source, fstype, target, options, err)
+		return err
+	} else {
+		klog.V(4).Infof("%s is already formatted (%s)", source, existingFormat)
```
27149chen: I think this log is needless, it will be printed in every remount. And without it, we already have enough knowledge in log that the device it already formatted before.

mount/mount_linux.go
mvisonneau: Indeed, I amended!

mount/mount_linux.go
mvisonneau: I removed it 👌

mount/mount_linux.go
mvisonneau: agreed, I updated the logic 👍

mount/mount_linux.go
```diff
@@ -289,55 +288,53 @@ func (mounter *SafeFormatAndMount) formatAndMount(source string, target string,
...
-		// Disk is already formatted and failed to mount
-		if len(fstype) == 0 || fstype == existingFormat {
-			// This is mount error
-			return mountErr
-		}
-		// Block device is formatted with unexpected filesystem, let the user know
-		return fmt.Errorf("failed to mount the volume as %q, it already contains %s. Mount error: %v", fstype, existingFormat, mountErr)
+
+		klog.Infof("Disk successfully formatted (mkfs): %s - %s %s", fstype, source, target)
+	} else if fstype != existingFormat {
```
jsafrane: Previously, it was possible to mount `ext3` volume even if its fsType was `ext4`. IMO it's a valid use case and we should not break it. Just try to mount it and throw stdout / stderr at the user, similarly to the previous code.

27149chen: You need to set default for fstype before this if-else, because it might be empty.

mount/mount_linux.go
```diff
@@ -289,55 +288,59 @@ func (mounter *SafeFormatAndMount) formatAndMount(source string, target string,
...
+	// Check if the disk is already formatted
+	existingFormat, err := mounter.GetDiskFormat(source)
+	if err != nil {
+		return err
+	}
+
+	if existingFormat == "" {
+		// Do not attempt to format the disk if mounting as readonly, return an error to reflect this.
+		if readOnly {
+			return fmt.Errorf("Cannot mount unformatted disk %s as we are manipulating it in read-only mode", source)
```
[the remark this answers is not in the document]
mvisonneau: done 👍

mount/mount_linux.go
mvisonneau: interesting, sounds good! I have added a warning instead 👍

mvisonneau: absolutely, thanks for the notice 👀 !

mount/mount_linux.go
```diff
@@ -2,13 +2,10 @@
 
 /*
 Copyright 2014 The Kubernetes Authors.
-
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
-
     http://www.apache.org/licenses/LICENSE-2.0
-
```
27149chen: don't touch the copyright header

mount/mount_linux.go
```diff
@@ -289,55 +285,53 @@ func (mounter *SafeFormatAndMount) formatAndMount(source string, target string,
...
-			// This is mount error
-			return mountErr
-		}
-		// Block device is formatted with unexpected filesystem, let the user know
-		return fmt.Errorf("failed to mount the volume as %q, it already contains %s. Mount error: %v", fstype, existingFormat, mountErr)
+
+		klog.Infof("Disk successfully formatted (mkfs): %s - %s %s", fstype, source, target)
+	} else if fstype != existingFormat {
+		// Verify that the disk is formatted with filesystem type we are expecting
+		klog.Warningf("Configured to mount disk %s as %s but current format is %s, things might break", source, existingFormat, fstype)
```
27149chen: I think ext3/4 is a special case, if you want to mount an xfs device as ext4, it should fail. @jsafrane , am I right?

jsafrane: `mounter.Interface.Mount` below will fail with nozero exit code, in the same way as it fails today.

mount/safe_format_and_mount_test.go
```diff
@@ -1,12 +1,9 @@
 /*
 Copyright 2014 The Kubernetes Authors.
-
 Licensed under the Apache License, Version 2.0 (the "License");
```
jsafrane: Please don't modify copyright headers.

mount/mount_linux.go
27149chen: you are right!

mount/safe_format_and_mount_test.go
mvisonneau: indeed 🤦‍♂ sorry about that

mount/mount_linux.go
mvisonneau: indeed 🤦‍♂ sorry about that

mount/mount_linux.go
```diff
@@ -289,55 +288,53 @@ func (mounter *SafeFormatAndMount) formatAndMount(source string, target string,
...
```
jingxu97: sorry to comment late. here if the required file system does not match the existing one, why not return error since the following mount should fail in this case?

mount/mount_linux.go
```diff
@@ -289,55 +288,53 @@ func (mounter *SafeFormatAndMount) formatAndMount(source string, target string,
...
-			return err
+	// Check if the disk is already formatted
+	existingFormat, err := mounter.GetDiskFormat(source)
+	if err != nil {
+		return err
+	}
+
+	// Use 'ext4' as the default
+	if len(fstype) == 0 {
+		fstype = "ext4"
```
jingxu97: here we are saying if user does not specify a fstype, by default we set it "ext4". But why not we can set it as the current existing file system?

27149chen: fstype is the default value for all. means that if existing format is ext3, and fstype is empty, the device will be mounted as ext4, not ext3.

mount/mount_linux.go
27149chen: I think the first comment of @jsafrane  explained the reason:

> Previously, it was possible to mount ext3 volume even if its fsType was ext4. IMO it's a valid use case and we should not break it. Just try to mount it and throw stdout / stderr at the user, similarly to the previous code.

PR #130, opened 2020-01-08
Add set type for net.IP

andrewsykim:
As discussed in https://github.com/kubernetes/kubernetes/pull/85617#discussion_r359002282, it would be generically useful to have a set type for net.IP. 

/assign @danwinship @thockin

### DIFF
--- a/net/ipnet.go
+++ b/net/ipnet.go
@@ -17,6 +17,7 @@ limitations under the License.
 package net
 
 import (
+	"fmt"
 	"net"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


danwinship:
/lgtm
in terms of correctness and consistency with `IPNetSet`, though it's annoying that `IPNetSet` (and now `IPSet`) is inconsistent with `k8s.io/apimachinery/pkg/util/set`. eg, `ParseIPNets()` rather than `NewIPNet()` and `StringSlice()` rather than `UnsortedList()` and no `Union()`/`Intersection()`/etc

andrewsykim:
> though it's annoying that IPNetSet (and now IPSet) is inconsistent with k8s.io/apimachinery/pkg/util/set. eg, ParseIPNets() rather than NewIPNet() and StringSlice() rather than UnsortedList() and no Union()/Intersection()/etc

:100: agree with this. I can look into this in a follow-up PR but it's quite a bit of work because we have to revendor it back to kubernetes/kubernetes.

PR #132, opened 2020-01-17
Improve the workflow of xfs devices filesystem check and mount

27149chen:
This PR improves the method checkAndRepairXfsFilesystem which will be used before mounting an xfs device.
changes are as following:
1. Retry three times if xfs_repair is failed.
2. Add a new function "replayXfsDirtyLogs" to replay dirty logs by mounting and immediately unmounting the filesystem.
3. Fix some test cases and add add somre more.

Backgroud:
1. If xfs_repair fails to repair the file system successfully, try giving the same xfs_repair command twice more; xfs_repair may be able to make more repairs on successive runs.
2.  Due to the design of the XFS log, a dirty log can only be replayed by the kernel, on a machine having the same CPU architecture as the machine which was writing to the log.  xfs_repair cannot replay a dirty log and will exit with a status code of 2 when it detects a dirty log. In this situation, the log can be replayed by mounting and immediately unmounting the filesystem on the same class of machine that crashed.  

References:
1. http://man7.org/linux/man-pages/man8/xfs_repair.8.html
2. http://fibrevillage.com/storage/666-how-to-repair-a-xfs-filesystem


27149chen:
@jsafrane  @gnufied  anyone who can help to have a look? thanks.

27149chen:
@saad-ali  @jingxu97  could you help to have a look at this pr? thanks.

dims:
approve in principle ... please get a LGTM from sig-storage folks

mount/mount_linux.go
```diff
@@ -292,18 +297,71 @@ func (mounter *SafeFormatAndMount) checkAndRepairXfsFilesystem(source string) er
...
+	out, err := mounter.Exec.Command("xfs_repair", source).CombinedOutput()
+	if err != nil {
+		return NewMountError(HasFilesystemErrors, "'xfs_repair' found errors on device %s but could not correct them: %s\n", source, out)
+	} else {
+		klog.Infof("Device %s has errors which were corrected by xfs_repair.", source)
+		return nil
+	}
+}
+
 // formatAndMount uses unix utils to format and mount the given disk
```
gnufied: How did you test this change btw? Via a CSI driver I assume?

mount/mount_linux.go
```diff
@@ -292,18 +297,71 @@ func (mounter *SafeFormatAndMount) checkAndRepairXfsFilesystem(source string) er
...
+							return mounter.replayXfsDirtyLogs(source)
+						default:
+							return NewMountError(HasFilesystemErrors, "'xfs_repair' found errors on device %s but could not correct them: %s\n", source, string(out))
+						}
+					} else {
+						return NewMountError(HasFilesystemErrors, "failed to run 'xfs_repair' on device %s: %v\n", source, err)
+					}
+				} else {
+					klog.Infof("Device %s has errors which were corrected by xfs_repair.", source)
+					return nil
```
gnufied: nit - Can we flatten some of this nested by returning early?

```golang
out, err := mounter.Exec.Command("xfs_repair", args...).CombinedOutput()
	if err == nil {
		klog.Infof("Device %s has errors which were corrected by xfs_repair.", source)
		return nil
	}
	e, isExitError := err.(utilexec.ExitError)
	if !isExitError {
		return NewMountError(HasFilesystemErrors, "failed to run 'xfs_repair' on device %s: %v\n", source, err)
		
	}
	switch e.ExitStatus() {
	case xfsRepairErrorsUncorrected:
		uncorrectedErr = NewMountError(HasFilesystemErrors, "'xfs_repair' found errors on device %s but could not correct them: %s\n", source, string(out))
		// If the exit status is 2, do not retry, replay the dirty logs instead.
	case xfsRepairErrorsDirtyLogs:
		return mounter.replayXfsDirtyLogs(source)
	default:
		return NewMountError(HasFilesystemErrors, "'xfs_repair' found errors on device %s but could not correct them: %s\n", source, string(out))
	}
```

mount/mount_linux.go
```diff
@@ -292,18 +297,71 @@ func (mounter *SafeFormatAndMount) checkAndRepairXfsFilesystem(source string) er
...
-			if err != nil {
-				return NewMountError(HasFilesystemErrors, "'xfs_repair' found errors on device %s but could not correct them: %s\n", source, out)
-			} else {
-				klog.Infof("Device %s has errors which were corrected by xfs_repair.", source)
-				return nil
+			var uncorrectedErr error
+
+			// If xfs_repair fails to repair the file system successfully, try giving the same xfs_repair
+			// command twice more; xfs_repair may be able to make more repairs on successive runs.
+			for i := 0; i < 3; i++ {
```
gnufied: Move number 3 to a constant and add comment for running `xfs_repair` 3 times there.

mount/mount_linux.go
```diff
@@ -292,18 +297,71 @@ func (mounter *SafeFormatAndMount) checkAndRepairXfsFilesystem(source string) er
...
+	klog.V(4).Infof("Attempting to replay the dirty logs on device %s", source)
+	msg := fmt.Sprintf("failed to replay dirty logs on device %s", source)
+	target, err := ioutil.TempDir("", "")
+	if err != nil {
+		return NewMountError(HasFilesystemErrors, "%s: %v\n", msg, err)
+	}
+	defer os.RemoveAll(target)
+
+	klog.V(4).Infof("Attempting to mount disk %s at %s", source, target)
+	if err := mounter.Interface.Mount(source, target, "", []string{"defaults"}); err != nil {
```
gnufied: Why not use `target` field given in the mount command as mount location rather than tempdir? Could a kubelet/app crash leave the volume mounted in tempdir and never get cleaned up?

mount/mount_linux.go
27149chen: fixed, it looks better, thanks.

mount/mount_linux.go
27149chen: yes, we are supporting xfs volume and snapshot, and find that sometimes mount is failed.

mount/mount_linux.go
27149chen: I think maybe the target is not existing or not available for mounting for some potential reasons. tempdir is more safer to temporarily mount and unmount, and we will make sure it is deleted in the end.
But you are right, it is a risk if the caller crashes between mount and unmount.

27149chen: Think again, it is the caller of formatAndMount's responsibility to make sure the existence and availablity of the target, so I think we can use the `target` field here.

27149chen:
@gnufied I have updated the code according to your review, PTAL. thanks.

mount/mount_linux.go
```diff
@@ -292,18 +298,66 @@ func (mounter *SafeFormatAndMount) checkAndRepairXfsFilesystem(source string) er
...
 			}
 		}
 	}
 	return nil
 }
 
+// replayXfsDirtyLogs tries to replay dirty logs by by mounting and
+// immediately unmounting the filesystem on the same class of machine
+// that crashed.
+func (mounter *SafeFormatAndMount) replayXfsDirtyLogs(source, target string) error {
```
saad-ali: If this operation fails at some point, do we have to worry about mounts it created?

mount/mount_linux.go
```diff
@@ -292,18 +298,66 @@ func (mounter *SafeFormatAndMount) checkAndRepairXfsFilesystem(source string) er
...
-			if err != nil {
-				return NewMountError(HasFilesystemErrors, "'xfs_repair' found errors on device %s but could not correct them: %s\n", source, out)
-			} else {
-				klog.Infof("Device %s has errors which were corrected by xfs_repair.", source)
-				return nil
+			var uncorrectedErr error
+
+			// If xfs_repair fails to repair the file system successfully, try giving the same xfs_repair
+			// command twice more; xfs_repair may be able to make more repairs on successive runs.
+			for i := 0; i < xfsRepairRetries; i++ {
```
saad-ali: Dumb question but I don't understand this much so bear with me: this code seems to retry 3 times no mater what, is it possible that after the first or second try, subsequent retries are unnecessary?

gnufied: Won't this result in calling `xfs_repair` 4 times? We already called one time outside the loop.

mount/mount_linux.go
gnufied: Yeah I was worried about this too, but in context of Kubernetes if mount target is a known directory to kubelet then it will either try to clean it up or complain loudly via orphan-pod function. We should be okay.

mount/mount_linux.go
27149chen: the first attempt is only a check (with -n flag) to see if there is any fs issues, it will not do any repair, and it is faster. So it can save some time if the fs is error free.

gnufied:
mostly lgtm.

@27149chen Do you know why we are seeing filesystem corruption on volumes restored from snapshots? Is that because snapshots were taken in inconsistent state?  Is there work tracked somewhere to fix the root cause of the problem too?

27149chen:
> mostly lgtm.
> 
> @27149chen Do you know why we are seeing filesystem corruption on volumes restored from snapshots? Is that because snapshots were taken in inconsistent state? Is there work tracked somewhere to fix the root cause of the problem too?

@gnufied yes.
When taking snapshot with continuous IO in container, there may be some data still be cached  in filesystem buffer. So there maybe some corruptions between filesystem metadata (superblock) and system log. So we need to repair the corruptions before filesystem mount.
I think the ideal way is freezing the pod before snapshop, we do not support it yet, and there are still some discussions like this one: https://github.com/container-storage-interface/spec/pull/407

mount/mount_linux.go
```diff
@@ -292,18 +298,74 @@ func (mounter *SafeFormatAndMount) checkAndRepairXfsFilesystem(source string) er
...
+	// make sure unmount is always called after mount.
+	err := func() (returnErr error) {
+		defer func() {
+			klog.V(4).Infof("Unmounting %s", target)
+			if err := mounter.Interface.Unmount(target); err != nil {
+				returnErr = NewMountError(HasFilesystemErrors, "%s: %v\n", msg, err)
+			}
+		}()
+		klog.V(4).Infof("Attempting to mount disk %s at %s", source, target)
+		if err := mounter.Interface.Mount(source, target, "", []string{"defaults"}); err != nil {
```
gnufied: From man-page:

```
In this situation, the log can be replayed by mounting and
       immediately unmounting the filesystem on the same class of machine
       that crashed.  Please make sure that the machine's hardware is
       reliable before replaying to avoid compounding the problems.
```

Are we being too aggressive in automatically fixing errors here? Can this make problems worse somehow? Should this be configurable? If we merge this PR as it is - it will become the new default even for in-tree Kubernetes drivers, so we have to be careful.

27149chen: @gnufied do you mean that the node might be bad or not the same class, so there is potential risk?

27149chen:
@gnufied , could you continue your review please? I asked a question, can you reply it? thanks.

gnufied:
@27149chen We are hoping someone who knows XFS more than me has a chance to review this PR before it gets merged. To me it looks good but my XFS knowledge is lacking and it makes big enough that change that could affect all of k8s including in-tree drivers (which are strictly in maintenance mode).

27149chen:
> @27149chen I am hoping someone who knows XFS more than me has a chance to review this PR before it gets merged. To me it looks good but my XFS knowledge is lacking and it makes big enough that change that could affect all of k8s including in-tree drivers (which are strictly in maintenance mode).

@gnufied , please add your lgtm label if it looks good to you.
I will try to find an xfs expert to help review it.

sandeen:
Hi - I'm the upstream xfsprogs maintainer, and have been an XFS developer for nearly 20 years.  The problem you're trying to solve here was recently brought to my attention.

It's my understanding that all of this effort to run `xfs_repair` stems from the fact that non-quiesced filesystems are getting snapshotted.  This inevitably leads to corruption and data loss, which cannot be un-done by `xfs_repair`.  To be honest, the only acceptable solution to the problem that this PR seems to be trying to solve is to go back and fix the snapshotting procedure to ensure that you have properly quiesced the filesystem (by unmounting it, or by issuing the FIFREEZE ioctl to the filesystem either via `xfs_freeze`, or directly), prior to the snapshot.

Anything else, honestly, is simply not going to work, and will _consistently_ lead to data loss.

27149chen:
> Hi - I'm the upstream xfsprogs maintainer, and have been an XFS developer for nearly 20 years. The problem you're trying to solve here was recently brought to my attention.
> 
> It's my understanding that all of this effort to run `xfs_repair` stems from the fact that non-quiesced filesystems are getting snapshotted. This inevitably leads to corruption and data loss, which cannot be un-done by `xfs_repair`. To be honest, the only acceptable solution to the problem that this PR seems to be trying to solve is to go back and fix the snapshotting procedure to ensure that you have properly quiesced the filesystem (by unmounting it, or by issuing the FIFREEZE ioctl to the filesystem either via `xfs_freeze`, or directly), prior to the snapshot.
> 
> Anything else, honestly, is simply not going to work, and will _consistently_ lead to data loss.

@sandeen thank you very much for your professional comment, it really helps a lot.
Here are some questions I want to ask:
1. You can find that we will run `fsck` in every remount, but as you know it is not work for xfs. So I want to know when and in which situation should we run `xfs_repair` in our code.
2. I understand that the best way is to go back and fix the snapshotting procedure. But what if an issue still happens after snapshot (no matter the procedure is right and wrong)? What should we do? If we can not repaire it by running `xfs_repair`, is there any other ways?
3. You may notice that always running `fsck` is not a good practice, like I said in #137, do you know how can we check if the filesystem is running well after mounting? So that we can run `fsck` after we detect that the filesystem has problem, instead of running it every time.

sandeen:
>     1. You can find that we will run `fsck` in every remount, but as you know it is not work for xfs. So I want to know when and in which situation should we run `xfs_repair` in our code.

There is no reason to fsck/repair any metadata journaling filesystem before every mount.  This is what the metadata log is for - it ensures consistency after a crash/power loss/etc.

For journaling filesystems, fsck/repair tools only need to be used after filesystem corruption has been detected, or if for some reason you need to verify filesystem integrity prior to some administrative operation.  (For example, ext4 recommends a full e2fsck to validate the filesystem before doing a resize, because resize can be a very invasive, possibly risky operation if corruption is encountered during the operation.)
 
>     2. I understand that the best way is to go back and fix the snapshotting procedure. But what if an issue still happens after snapshot (no matter the procedure is right or wrong)? What should we do? If we can not repaire it by running `xfs_repair`, is there any other ways?

If you are properly snapshotting the filesystem, xfs_repair won't be needed.  xfs_repair should be _exception_ activity, rare enough that it would be done manually when intervention is required.  i.e. something goes wrong (disk bit flip, admin error, code bug, etc), xfs notices the corruption and shuts down, the administrator notices the error, and runs xfs_repair.
Please note fsck tools like xfs_repair are not data recovery tools.  All they can do is put the filesystem back in a consistent state.  Sometimes this is done by repairing errors, and sometimes it is done by _discarding_ badly corrupted parts of the filesystem.

Please understand the difference: xfs_repair _will_ put a filesystem back into a _consistent_ state.  But it will _not_ fully recover the _prior_ filesystem state if you do something like take a non-atomic snapshot of the device.

>     3. You may notice that always running `fsck` is not a good practice, like I said in #137, do you know how can we check if the filesystem is running well after mounting? So that we can run `fsck` after we detect that the filesystem has problem, instead of running it every time.

XFS has runtime checks - both CRC verification as well as structure integrity validation - on every metadata read and write from disk.  If it finds an error, in most cases it will shut down the fileystem.  In general there should be no need to check the filesystem while it's mounted.
Can I ask why you have a need to frequently "check if the filesystem is running well after mounting?"
If you really do want/need to do this, there are a couple of options:

1. use xfs_scrub - which is currently experimental
2. properly snapshot the device, and run xfs_repair -n on that snapshot

27149chen:
@sandeen , thank you for your explaination. My understanding is that we don't need to run fsck/xfs_repair before mounting in our code, we should mount it directly, and if it succeeds, we don't need to do anything, and if it is failed, we should return error and ask the user fix it manually. Am I right?

sandeen:
> @sandeen , thank you for your explaination. My understanding is that we don't need to run fsck/xfs_repair before mounting in our code, we should mount it directly, and if it succeeds, we don't need to do anything, and if it is failed, we should return error and ask the user fix it manually. Am I right?

Apologies for the late reply, I didn't get a notification about your question.

That would be my suggestion, yes.

PR #133, opened 2020-01-21
Split MajorMinor into two fields

dims:
this reflects what is in the docker/docker/pkg/mount API. makes it
easier to port additional code to out API.

### DIFF
--- a/mount/mount_helper_unix.go
+++ b/mount/mount_helper_unix.go
@@ -61,8 +61,12 @@ type MountInfo struct {
 	ID int
 	// The ID of the parent mount (or of self for the root of this mount namespace's mount tree).
 	ParentID int
-	// The value of `st_dev` for files on this filesystem.
-	MajorMinor string
+	// Major indicates one half of the device ID which identifies the device class
+	// (parsed from `st_dev` for files on this filesystem).
+	Major int
+	// Minor indicates one half of the device ID which identifies a specific
+	// instance of device (parsed from `st_dev` for files on this filesystem).
+	Minor int
 	// The pathname of the directory in the filesystem which forms the root of this mount.
 	Root string
 	// Mount source, filesystem-specific information. e.g. device, tmpfs name.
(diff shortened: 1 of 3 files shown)


dims:
Example of porting cadvisor where this was a bit hard - https://github.com/google/cadvisor/pull/2383/files

dims:
Bonus: it does not affect k/k in any way as we don't use the field - https://cs.k8s.io/?q=MajorMinor&i=nope&files=&repos=kubernetes/kubernetes

mount/mount_helper_unix.go
```diff
@@ -106,10 +110,18 @@ func ParseMountInfo(filename string) ([]MountInfo, error) {
 		if err != nil {
 			return nil, err
 		}
+		mm := strings.Split(fields[2], ":")
+		if len(mm) != 2 {
+			return nil, fmt.Errorf("parsing '%s' failed: unexpected minor:major pair %s", line, mm)
+		}
+		major, _ := strconv.Atoi(mm[0])
```
thockin: Why not actually handle errors?

dims: Mirroring existing behavior that we may be depending on - https://github.com/kubernetes/kubernetes/blob/master/vendor/github.com/docker/docker/pkg/mount/mountinfo_linux.go#L52-L53

i haven't started at enough of these st_dev(s) to say if that's a good idea or a bad idea :)

dims: Also currently we don't throw an error if the string has bad contents ...

thockin:
Past misbehavior does not forgive perpetuating it :)

dims:
+1 @thockin fixing it now.

dims:
DONE @thockin PTAL. Added a bunch of tests for the new behavior.

codenrhoden:
LGTM. Nice work with the added tests. Not adding `/lgtm` since I don't want this to merge before @thockin can take another look.

mount/mount_helper_unix.go
```diff
@@ -106,10 +110,24 @@ func ParseMountInfo(filename string) ([]MountInfo, error) {
 		if err != nil {
 			return nil, err
 		}
+		mm := strings.Split(fields[2], ":")
+		if len(mm) != 2 {
+			return nil, fmt.Errorf("parsing '%s' failed: unexpected minor:major pair %s", line, mm)
+		}
+		major, err := strconv.Atoi(mm[0])
+		if err != nil {
+			return nil, fmt.Errorf("parsing '%s' failed: unable to parse major device id", mm[0])
```
thockin: nit: can you print the error, too?  In the unlikely event this fires....

dims: Done!

dims:
/hold cancel

thanks @thockin

PR #138, opened 2020-02-13
Introduce parameter for sensitive mount options.

saad-ali:
Introduce optional `sensitiveOptions` parameter to allow sensitive mount options to be passed in a separate parameter from the normal mount options and ensures the `sensitiveOptions` are never logged.

Once this is merged, any mount library consumers that pass sensitive mount options should be updated to take advantage of the new parameter.

/assign @msau42 
/assign @liggitt

### DIFF
--- a/mount/fake_mounter.go
+++ b/mount/fake_mounter.go
@@ -82,6 +82,15 @@ func (f *FakeMounter) GetLog() []FakeAction {
 
 // Mount records the mount event and updates the in-memory mount points for FakeMounter
 func (f *FakeMounter) Mount(source string, target string, fstype string, options []string) error {
+	return f.MountSensitive(source, target, fstype, options, nil /* sensitiveOptions */)
+}
+
+// Mount records the mount event and updates the in-memory mount points for FakeMounter
+// sensitiveOptions to be passed in a separate parameter from the normal
+// mount options and ensures the sensitiveOptions are never logged. This
+// method should be used by callers that pass sensitive material (like
+// passwords) as mount options.
+func (f *FakeMounter) MountSensitive(source string, target string, fstype string, options []string, sensitiveOptions []string) error {
 	f.mutex.Lock()
 	defer f.mutex.Unlock()
 
@@ -117,7 +126,7 @@ func (f *FakeMounter) Mount(source string, target string, fstype string, options
 	if err != nil {
 		absTarget = target
 	}
-	f.MountPoints = append(f.MountPoints, MountPoint{Device: source, Path: absTarget, Type: fstype, Opts: opts})
+	f.MountPoints = append(f.MountPoints, MountPoint{Device: source, Path: absTarget, Type: fstype, Opts: append(opts, sensitiveOptions...)})
 	klog.V(5).Infof("Fake mounter: mounted %s to %s", source, absTarget)
 	f.log = append(f.log, FakeAction{Action: FakeActionMount, Target: absTarget, Source: source, FSType: fstype})
 	return nil
--- a/mount/mount.go
+++ b/mount/mount.go
@@ -31,12 +31,21 @@ import (
 const (
 	// Default mount command if mounter path is not specified.
 	defaultMountCommand = "mount"
+	// Log message where sensitive mount options were removed
+	sensitiveOptionsRemoved = "<masked>"
 )
 
 // Interface defines the set of methods to allow for mount operations on a system.
 type Interface interface {
 	// Mount mounts source to target as fstype with given options.
+	// options MUST not contain sensitive material (like passwords).
 	Mount(source string, target string, fstype string, options []string) error
+	// MountSensitive is the same as Mount() but this method allows
+	// sensitiveOptions to be passed in a separate parameter from the normal
+	// mount options and ensures the sensitiveOptions are never logged. This
+	// method should be used by callers that pass sensitive material (like
+	// passwords) as mount options.
+	MountSensitive(source string, target string, fstype string, options []string, sensitiveOptions []string) error
 	// Unmount unmounts given target.
 	Unmount(target string) error
 	// List returns a list of all mounted filesystems.  This can be large.
@@ -72,7 +81,7 @@ type MountPoint struct {
 	Device string
 	Path   string
 	Type   string
-	Opts   []string
+	Opts   []string // Opts may contain sensitive mount options (like passwords) and MUST be treated as such (e.g. not logged).
 	Freq   int
 	Pass   int
 }
@@ -122,8 +131,18 @@ type SafeFormatAndMount struct {
 // read-only it will format it first then mount it. Otherwise, if the
 // disk is already formatted or it is being mounted as read-only, it
 // will be mounted without formatting.
+// options MUST not contain sensitive material (like passwords).
 func (mounter *SafeFormatAndMount) FormatAndMount(source string, target string, fstype string, options []string) error {
-	return mounter.formatAndMount(source, target, fstype, options)
+	return mounter.FormatAndMountSensitive(source, target, fstype, options, nil /* sensitiveOptions */)
+}
+
+// FormatAndMountSensitive is the same as FormatAndMount but this method allows
+// sensitiveOptions to be passed in a separate parameter from the normal mount
+// options and ensures the sensitiveOptions are never logged. This method should
+// be used by callers that pass sensitive material (like passwords) as mount
+// options.
+func (mounter *SafeFormatAndMount) FormatAndMountSensitive(source string, target string, fstype string, options []string, sensitiveOptions []string) error {
+	return mounter.formatAndMountSensitive(source, target, fstype, options, sensitiveOptions)
 }
 
 // getMountRefsByDev finds all references to the device provided
@@ -240,6 +259,16 @@ func IsNotMountPoint(mounter Interface, file string) (bool, error) {
 // The list equals:
 //   options - 'bind' + 'remount' (no duplicate)
 func MakeBindOpts(options []string) (bool, []string, []string) {
+	bind, bindOpts, bindRemountOpts, _ := MakeBindOptsSensitive(options, nil /* sensitiveOptions */)
+	return bind, bindOpts, bindRemountOpts
+}
+
+// MakeBindOptsSensitive is the same as MakeBindOpts but this method allows
+// sensitiveOptions to be passed in a separate parameter from the normal mount
+// options and ensures the sensitiveOptions are never logged. This method should
+// be used by callers that pass sensitive material (like passwords) as mount
+// options.
+func MakeBindOptsSensitive(options []string, sensitiveOptions []string) (bool, []string, []string, []string) {
 	// Because we have an FD opened on the subpath bind mount, the "bind" option
 	// needs to be included, otherwise the mount target will error as busy if you
 	// remount as readonly.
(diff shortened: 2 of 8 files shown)


mount/fake_mounter.go
```diff
@@ -117,7 +126,7 @@ func (f *FakeMounter) Mount(source string, target string, fstype string, options
 	if err != nil {
 		absTarget = target
 	}
-	f.MountPoints = append(f.MountPoints, MountPoint{Device: source, Path: absTarget, Type: fstype, Opts: opts})
+	f.MountPoints = append(f.MountPoints, MountPoint{Device: source, Path: absTarget, Type: fstype, Opts: append(opts, sensitiveOptions...)})
```
liggitt: are options ever order sensitive, or can sensitive options always be grouped together at the end?

saad-ali: Ceph appends them to the end https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/cephfs/cephfs.go#L315

I'll check Azure

saad-ali: AzureFile also appears to tack them on to the end 
https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/azure_file/azure_file.go#L284

mount/mount.go
```diff
@@ -36,7 +36,14 @@ const (
...
 type Interface interface {
 	// Mount mounts source to target as fstype with given options.
+	// options MUST not contain sensitive material (like passwords).
 	Mount(source string, target string, fstype string, options []string) error
+	// MountSensitive is the same as Mount() but this method allows
+	// sensitiveOptions to be passed in a separate parameter from the normal
+	// mount options and ensures the sensitiveOptions are never logged. This
+	// method should be used by callers that pass sensitive material (like
+	// passwords) as mount options.
+	MountSensitive(source string, target string, fstype string, options []string, sensitiveOptions []string) error
```
liggitt: do we expect implementations of `mount.Interface` other than the fake/linux/windows/unsupported mounters provided by this package? if so, adding this method to an existing interface makes this version of k8s.io/utils incompatible with those implementations

mount/fake_mounter.go
liggitt: add a comment to the MountPoint#Opts field that indicates it may contain sensitive values and should not be logged?

mount/mount_linux.go
```diff
@@ -169,31 +183,39 @@ func detectSystemd() bool {
 }
 
 // MakeMountArgs makes the arguments to the mount(8) command.
-// Implementation is shared with NsEnterMounter
-func MakeMountArgs(source, target, fstype string, options []string) []string {
```
liggitt: taking a public method private is not compatible

mount/mount_linux.go
```diff
@@ -169,31 +183,39 @@ func detectSystemd() bool {
...
 	mountArgs = append(mountArgs, target)
+	mountArgsLogStr += " " + target + " "
 
-	return mountArgs
+	return mountArgs, mountArgsLogStr
 }
 
-// AddSystemdScope adds "system-run --scope" to given command line
-// implementation is shared with NsEnterMounter
-func AddSystemdScope(systemdRunPath, mountName, command string, args []string) (string, []string) {
```
liggitt: same comment about a public method

mount/mount_linux.go
```diff
@@ -305,9 +327,10 @@ func (mounter *SafeFormatAndMount) checkAndRepairXfsFilesystem(source string) er
 }
 
 // formatAndMount uses unix utils to format and mount the given disk
-func (mounter *SafeFormatAndMount) formatAndMount(source string, target string, fstype string, options []string) error {
+func (mounter *SafeFormatAndMount) formatAndMountSensitive(source string, target string, fstype string, options []string, sensitiveOptions []string) error {
 	readOnly := false
 	for _, option := range options {
+		// No need to check sensitiveOptions since that should not contain ro
```
liggitt: unless you specifically document it cannot be provided in sensitiveoptions, check both

mount/mount_linux.go
```diff
@@ -347,7 +370,12 @@ func (mounter *SafeFormatAndMount) formatAndMount(source string, target string,
 		klog.Infof("Disk %q appears to be unformatted, attempting to format as type: %q with options: %v", source, fstype, args)
 		output, err := mounter.Exec.Command("mkfs."+fstype, args...).CombinedOutput()
 		if err != nil {
-			detailedErr := fmt.Sprintf("format of disk %q failed: type:(%q) target:(%q) options:(%q) errcode:(%v) output:(%v) ", source, fstype, target, options, err, string(output))
+			// Do not log sensitiveOptions only options
+			sensitiveOptionsLog := ""
+			if len(sensitiveOptions) > 0 {
+				sensitiveOptionsLog = sensitiveOptionsRemoved
```
liggitt: suggest `strings.Repeat("<masked> ", len(sensitiveOptions))` so we can see how many options were masked

liggitt: or making a `optionsForLogging` slice that starts with `options` and appends `<masked>` the right number of times, so you can include a single %q in the log message and get a single list (instead of `["foo"]"<masked>"`)

mount/mount_linux.go
```diff
@@ -43,6 +43,8 @@ const (
 	fsckErrorsCorrected = 1
 	// 'fsck' found errors but exited without correcting them
 	fsckErrorsUncorrected = 4
+	// Log message where sensitive mount options were removed
+	sensitiveOptionsRemoved = " [sensitiveOptions Removed] "
 )
```
liggitt: where we parse from the filesystem, do we also need to stop logging content from /proc/mounts in error cases:

https://github.com/kubernetes/utils/blob/861946025e3491219eaccb1bf693e23df70c2fa8/mount/mount_linux.go#L458-L461

mount/mount.go
liggitt: looks like the mount_windows.go implementation needs updating as well

mount/fake_mounter.go
saad-ali: Done.

mount/mount_linux.go
saad-ali: How strict do we want to be about that? As far as I can tell they are internal implementation details. I'd love to reduce the API surface if we can. The most important consumer of this is Kubernetes and it does not reference these methods.

liggitt: > How strict do we want to be about that?

As strict as we can? The intent of this library is to be a well-behaved, stable dependency. I agree with limiting API surface area, but the time to do that is before we publish releases.

mount/mount_linux.go
saad-ali: Will do.

mount/mount_linux.go
saad-ali: Ack. Will reintroduce and add a sensitive version as well.

mount/mount_linux.go
saad-ali: Will reintroduce and add a sensitive version as well.

mount/mount_linux.go
saad-ali: Will do.

mount/mount_linux.go
saad-ali: Good point. Unfortunate, because if it ever fails having the actual line will be valuable debugging. Will remove and comment.

mount/mount.go
saad-ali: > do we expect implementations of mount.Interface other than the fake/linux/windows/unsupported mounters provided by this package? if so, adding this method to an existing interface makes this version of k8s.io/utils incompatible with those implementations

My understand is, no. Based on the Kubernetes consumers, the callers all first construct one of the `Mounters` provided by this package (fake/linux/windows/unsupported). There is little value in using the other methods provided by the library with a custom mounter.

Updated `mount_windows.go` implementation.

mount/mount_linux.go
```diff
@@ -169,33 +181,56 @@ func detectSystemd() bool {
...
 	// Build mount command as follows:
 	//   mount [-t $fstype] [-o $options] [$source] $target
-	mountArgs := []string{}
+	mountArgs = []string{}
+	mountArgsLogStr = ""
 	if len(fstype) > 0 {
 		mountArgs = append(mountArgs, "-t", fstype)
+		mountArgsLogStr += strings.Join(mountArgs, " ")
 	}
 	if len(options) > 0 {
```
liggitt:
```suggestion
	if len(options) > 0 || len(sensitiveOptions) > 0 {
```

mount/mount_linux.go
```diff
@@ -169,33 +181,56 @@ func detectSystemd() bool {
...
-		mountArgs = append(mountArgs, "-o", strings.Join(options, ","))
+		mountArgs = append(mountArgs, "-o", strings.Join(append(options, sensitiveOptions...), ","))
```
liggitt: don't modify the passed in options slice

```suggestion
		combinedOptions := []string{}
		combinedOptions = append(combinedOptions, options...)
		combinedOptions = append(combinedOptions, sensitiveOptions...)
		mountArgs = append(mountArgs, "-o", strings.Join(combinedOptions, ","))
```

mount/mount_linux.go
```diff
@@ -169,33 +181,56 @@ func detectSystemd() bool {
...
+		// exclude sensitiveOptions from log string
+		mountArgsLogStr += " -o " + sanitizedOptionsForLogging(options, sensitiveOptions)
 	}
 	if len(source) > 0 {
 		mountArgs = append(mountArgs, source)
+		mountArgsLogStr += " " + source + " "
```
liggitt: why the trailing space?

mount/mount_linux.go
```diff
@@ -169,33 +181,56 @@ func detectSystemd() bool {
...
 	}
 	mountArgs = append(mountArgs, target)
+	mountArgsLogStr += " " + target + " "
```
liggitt: why the trailing space?

mount/mount_linux.go
liggitt: and add a unit test that would have caught this

mount/mount_linux.go
```diff
@@ -69,16 +69,26 @@ func New(mounterPath string) Interface {
...
+// MountSensitive is the same as Mount() but this method allows
+// sensitiveOptions to be passed in a separate parameter from the normal
+// mount options and ensures the sensitiveOptions are never logged. This
+// method should be used by callers that pass sensitive material (like
+// passwords) as mount options.
+func (mounter *Mounter) MountSensitive(source string, target string, fstype string, options []string, sensitiveOptions []string) error {
 	// Path to mounter binary if containerized mounter is needed. Otherwise, it is set to empty.
 	// All Linux distros are expected to be shipped with a mount utility that a support bind mounts.
 	mounterPath := ""
+	// No need to pass sensitiveOptions to MakeBindOpts because it should not contain bind options.
```
liggitt: expectations like this need to be made clear in godoc for all the public-facing `*Sensitive` methods... the current docs make it seem like I could just pass all options as sensitive options without any implications

mount/mount_linux.go
saad-ali: Fixed

mount/mount_linux.go
saad-ali: Fixed

mount/mount_linux.go
saad-ali: Unnecessary, removed.

mount/mount_linux.go
saad-ali: Unnecessary, removed.

mount/mount_linux.go
saad-ali: Good point. Added a `MakeBindOptsSensitive`.

saad-ali:
PTAL

PR #139, opened 2020-02-17
Adds nested traces to the trace library

SomtochiAma:
This PR adds the NestedTrace to the Trace and also formats the output so that it is more concise. A use case for the nested trace [kubernetes#79209]. (https://github.com/kubernetes/kubernetes/issues/79209)

Example of old output:
`
Trace[866828359]: "Create /api/v1/namespaces/default/pods" (started: 2019-03-11 08:20:14.448210695 +0000 UTC m=+579.430813234) (total time: 14.520559394s):
Trace[866828359]: [165.764543ms] [149.545551ms] Object stored in database
Trace[866828359]: [14.304157451s] [14.138392908s] Writing response
Trace[866828359]: [14.520559394s] [216.401943ms] END
`

There is no nested trace and related traces were logged separately.

Sample of the new, concise output with nested trace 

image

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -32,13 +32,13 @@ type Field struct {
 	Value interface{}
 }
 
-type stepTraceTime interface {
+type stepTrace interface {
 	time() time.Time
 	writeStep(b *bytes.Buffer, formatter string, stepDuration time.Duration, lastStepTime time.Time) time.Time
 }
 
-func (t *Trace) sortStepTrace() []stepTraceTime {
-	var arr []stepTraceTime
+func (t *Trace) sortStepTrace() []stepTrace {
+	var arr []stepTrace
 
 	for _, step := range t.steps {
 		arr = append(arr, step)
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -91,6 +91,7 @@ type Trace struct {
 	threshold   *time.Duration
 	startTime   time.Time
 	stepsTraces []stepTrace
+	parentTrace *Trace
 }
 
 func (t *Trace) time() time.Time {
@@ -99,20 +100,13 @@ func (t *Trace) time() time.Time {
 
 func (t *Trace) writeStep(b *bytes.Buffer, formatter string, parentTrace *Trace, stepThreshold time.Duration,
 	lastStepTime time.Time) time.Time {
-	stepDuration := t.startTime.Sub(lastStepTime)
-	if stepThreshold == 0 || stepDuration > stepThreshold || klog.V(4) {
-		if t.threshold != nil {
-			stepThreshold = *parentTrace.threshold - *t.threshold
-			limitThreshold := *parentTrace.threshold / 4
-			if stepThreshold < limitThreshold {
-				stepThreshold = limitThreshold
-			}
-		}
-		b.WriteString(fmt.Sprintf("%v[", formatter))
-		writeMainLog(b, t.name, t.TotalTime(), t.startTime, t.fields)
-		_ = writeTrace(b, t, formatter+" ", stepThreshold)
-		b.WriteString("]")
+	if t.threshold != nil {
+		stepThreshold = calculateStepThreshold(t)
 	}
+	b.WriteString(fmt.Sprintf("%v[", formatter))
+	writeMainLog(b, t.name, t.TotalTime(), t.startTime, t.fields)
+	_ = writeTrace(b, t, formatter+" ", stepThreshold)
+	b.WriteString("]")
 	return lastStepTime
 }
 
@@ -136,6 +130,7 @@ func (t *Trace) Step(msg string, fields ...Field) {
 // Nest adds a nested trace with the given message and fields and returns it.
 func (t *Trace) Nest(msg string, fields ...Field) *Trace {
 	newTrace := New(msg, fields...)
+	newTrace.parentTrace = t
 	t.stepsTraces = append(t.stepsTraces, newTrace)
 	return newTrace
 }
@@ -183,14 +178,43 @@ func writeTrace(b *bytes.Buffer, t *Trace, formatter string, stepThreshold time.
 // LogIfLong is used to dump steps that took longer than its share
 func (t *Trace) LogIfLong(threshold time.Duration) {
 	t.threshold = &threshold
-	if time.Since(t.startTime) >= threshold {
+	if time.Since(t.startTime) >= threshold || t.parentTrace == nil {
 		// if any step took more than it's share of the total allowed time, it deserves a higher log level
-		stepThreshold := threshold / time.Duration(len(t.stepsTraces)+1)
+		stepThreshold := calculateStepThreshold(t)
 		t.logWithStepThreshold(stepThreshold)
+	} else {
+		for _, s := range t.stepsTraces {
+			nestedTrace, ok := s.(*Trace)
+			if ok && nestedTrace.threshold != nil && time.Since(nestedTrace.startTime) >= *nestedTrace.threshold  {
+				stepThreshold := calculateStepThreshold(nestedTrace)
+				nestedTrace.logWithStepThreshold(stepThreshold)
+			}
+		}
 	}
 }
 
 // TotalTime can be used to figure out how long it took since the Trace was created
 func (t *Trace) TotalTime() time.Duration {
 	return time.Since(t.startTime)
+}
+
+func calculateStepThreshold(t *Trace) time.Duration {
+	lenTrace := len(t.stepsTraces) + 1
+	traceThreshold := *t.threshold
+	for _, s := range t.stepsTraces {
+		nestedTrace, ok := s.(*Trace)
+		if ok && nestedTrace.threshold != nil {
+			traceThreshold = traceThreshold - *nestedTrace.threshold
+			lenTrace--
+		}
+	}
+
+	limitThreshold := *t.threshold / 4
+	if traceThreshold < limitThreshold {
+		traceThreshold = limitThreshold
+		lenTrace = len(t.stepsTraces) + 1
+	}
+
+	stepThreshold := traceThreshold / time.Duration(lenTrace)
+	return stepThreshold
 }
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"fmt"
 	"math/rand"
+	"sort"
 	"time"
 
 	"k8s.io/klog/v2"
@@ -29,6 +30,36 @@ import (
 type Field struct {
 	Key   string
 	Value interface{}
+}
+
+type stepTraceTime interface {
+	time() time.Time
+}
+
+func (t *Trace) time() time.Time {
+	return t.startTime
+}
+
+func (s traceStep) time() time.Time {
+	return s.stepTime
+}
+
+func (t *Trace) sortStepTrace() []stepTraceTime {
+	var arr []stepTraceTime
+
+	for _, step := range t.steps {
+		arr = append(arr, step)
+	}
+
+	for _, trace := range t.nestedTrace {
+		arr = append(arr, trace)
+	}
+
+	sort.Slice(arr, func(i, j int) bool {
+		return arr[i].time().Before(arr[j].time())
+	})
+
+	return arr
 }
 
 func (f Field) format() string {
@@ -44,6 +75,20 @@ func writeFields(b *bytes.Buffer, l []Field) {
 	}
 }
 
+func writeMainLog(b *bytes.Buffer, msg string, totalTime time.Duration, startTime time.Time, fields []Field) {
+	b.WriteString(fmt.Sprintf("%q ", msg))
+	if len(fields) > 0 {
+		writeFields(b, fields)
+		b.WriteString(" ")
+	}
+
+	b.WriteString(fmt.Sprintf("%vms (%v)", durationToMilliseconds(totalTime), startTime.Format("15:04:00.000")))
+}
+
+func durationToMilliseconds(timeDuration time.Duration) int64 {
+	return timeDuration.Nanoseconds() / 1e6
+}
+
 type traceStep struct {
 	stepTime time.Time
 	msg      string
@@ -53,10 +98,11 @@ type traceStep struct {
 // Trace keeps track of a set of "steps" and allows us to log a specific
 // step if it took longer than its share of the total allowed time
 type Trace struct {
-	name      string
-	fields    []Field
-	startTime time.Time
-	steps     []traceStep
+	name        string
+	fields      []Field
+	startTime   time.Time
+	steps       []traceStep
+	nestedTrace []*Trace
 }
 
 // New creates a Trace with the specified name. The name identifies the operation to be traced. The
@@ -76,7 +122,14 @@ func (t *Trace) Step(msg string, fields ...Field) {
 	t.steps = append(t.steps, traceStep{stepTime: time.Now(), msg: msg, fields: fields})
 }
 
-// Log is used to dump all the steps in the Trace
+// Nest adds a nested trace with the given message and fields and returns it.
+func (t *Trace) Nest(msg string, fields ...Field) *Trace {
+	newTrace := New(msg, fields...)
+	t.nestedTrace = append(t.nestedTrace, newTrace)
+	return newTrace
+}
+
+// Log is used to dump all the steps in the Trace. It also logs the nested trace messages using indentation
 func (t *Trace) Log() {
 	// an explicit logging request should dump all the steps out at the higher level
 	t.logWithStepThreshold(0)
@@ -93,27 +146,39 @@ func (t *Trace) logWithStepThreshold(stepThreshold time.Duration) {
 		writeFields(&buffer, t.fields)
 		buffer.WriteString(" ")
 	}
-	buffer.WriteString(fmt.Sprintf("(started: %v) (total time: %v):\n", t.startTime, totalTime))
-	lastStepTime := t.startTime
-	for _, step := range t.steps {
-		stepDuration := step.stepTime.Sub(lastStepTime)
-		if stepThreshold == 0 || stepDuration > stepThreshold || klog.V(4).Enabled() {
-			buffer.WriteString(fmt.Sprintf("Trace[%d]: [%v] [%v] ", tracenum, step.stepTime.Sub(t.startTime), stepDuration))
-			buffer.WriteString(step.msg)
-			if len(step.fields) > 0 {
-				buffer.WriteString(" ")
-				writeFields(&buffer, step.fields)
-			}
-			buffer.WriteString("\n")
-		}
-		lastStepTime = step.stepTime
-	}
+	buffer.WriteString(fmt.Sprintf("(%v) (total time: %vms):", t.startTime.Format("02-Jan-2006 15:04:00.000"), totalTime.Milliseconds()))
+	lastStepTime := writeTrace(&buffer, t, fmt.Sprintf("\nTrace[%d]: ", tracenum), stepThreshold)
 	stepDuration := endTime.Sub(lastStepTime)
-	if stepThreshold == 0 || stepDuration > stepThreshold || klog.V(4).Enabled() {
-		buffer.WriteString(fmt.Sprintf("Trace[%d]: [%v] [%v] END\n", tracenum, endTime.Sub(t.startTime), stepDuration))
+	if stepThreshold == 0 || stepDuration > stepThreshold || klog.V(4).Enabled()  {
+		buffer.WriteString(fmt.Sprintf("\nTrace[%d]: [%v] [%v] END\n", tracenum, endTime.Sub(t.startTime), stepDuration))
 	}
 
 	klog.Info(buffer.String())
+}
+
+func writeTrace(b *bytes.Buffer, t *Trace, formatter string, stepThreshold time.Duration) time.Time {
+	lastStepTime := t.startTime
+	stepAndTraces := t.sortStepTrace()
+	if len(stepAndTraces) == 0 {
+		return lastStepTime
+	}
+	for _, stepOrTrace := range stepAndTraces {
+		switch s := stepOrTrace.(type) {
+		case traceStep:
+			stepDuration := s.stepTime.Sub(lastStepTime)
+			if stepThreshold == 0 || stepDuration > stepThreshold || klog.V(4).Enabled() {
+				b.WriteString(fmt.Sprintf("%v---", formatter))
+				writeMainLog(b, s.msg, stepDuration, s.stepTime, s.fields)
+			}
+			lastStepTime = s.stepTime
+		case *Trace:
+			b.WriteString(fmt.Sprintf("%v[", formatter))
+			writeMainLog(b, s.name, s.TotalTime(), s.startTime, s.fields)
+			_ = writeTrace(b, s, formatter+" ", stepThreshold)
+			b.WriteString("]")
+		}
+	}
+	return lastStepTime
 }
 
 // LogIfLong is used to dump steps that took longer than its share
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -155,7 +155,7 @@ func (t *Trace) logWithStepThreshold(stepThreshold time.Duration) {
 	buffer.WriteString(fmt.Sprintf("(%v) (total time: %vms):", t.startTime.Format("02-Jan-2006 15:04:00.000"), totalTime.Milliseconds()))
 	lastStepTime := writeTrace(&buffer, t, fmt.Sprintf("\nTrace[%d]: ", tracenum), stepThreshold)
 	stepDuration := endTime.Sub(lastStepTime)
-	if stepThreshold == 0 || stepDuration > stepThreshold || klog.V(4).Enabled()  {
+	if stepThreshold == 0 || stepDuration > stepThreshold || klog.V(4).Enabled() {
 		buffer.WriteString(fmt.Sprintf("\nTrace[%d]: [%v] [%v] END\n", tracenum, endTime.Sub(t.startTime), stepDuration))
 	}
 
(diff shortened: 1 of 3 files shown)


trace/trace.go
```diff
@@ -129,3 +137,42 @@ func (t *Trace) LogIfLong(threshold time.Duration) {
 func (t *Trace) TotalTime() time.Duration {
 	return time.Since(t.startTime)
 }
+
+func writeNestedTrace(b *bytes.Buffer, nestedT []*Trace, threshold time.Duration, formatter string) *bytes.Buffer {
```
sttts: is this not wired yet?

sttts:
The new output looks great. Can you add an example of the old output without nesting and how it looks now?

trace/trace.go
SomtochiAma: Just updated the code

SomtochiAma:
> The new output looks great. Can you add an example of the old output without nesting and how it looks now?

I will do that now

trace/trace.go
```diff
@@ -93,21 +111,19 @@ func (t *Trace) logWithStepThreshold(stepThreshold time.Duration) {
 		writeFields(&buffer, t.fields)
 		buffer.WriteString(" ")
 	}
-	buffer.WriteString(fmt.Sprintf("(started: %v) (total time: %v):\n", t.startTime, totalTime))
+	buffer.WriteString(fmt.Sprintf("(%v) (total time: %vms):\n", t.startTime.Format("02-Jan-2006 15:04:00.000"), totalTime.Milliseconds()))
```
eloyekunle:
```suggestion
	buffer.WriteString(fmt.Sprintf("(%v) (total time: %vms):\n", t.startTime.Format("02-Jan-2006 15:04:00.000"), totalTime.Nanoseconds()/1000))
```

trace/trace.go
```diff
@@ -44,6 +45,15 @@ func writeFields(b *bytes.Buffer, l []Field) {
 	}
 }
 
+func writeMainLog(b *bytes.Buffer, msg string, totalTime time.Duration, startTime time.Time, fields []Field) {
+	b.WriteString(fmt.Sprintf("%q ", msg))
+	if len(fields) > 0 {
+		writeFields(b, fields)
+		b.WriteString(" ")
+	}
+	b.WriteString(fmt.Sprintf("%vms (%v)", totalTime.Microseconds(), startTime.Format("15:04:00.000")))
```
eloyekunle:
```suggestion
	b.WriteString(fmt.Sprintf("%vms (%v)", totalTime.Nanoseconds()/1000, startTime.Format("15:04:00.000")))
```

sttts:
> Should we drop it from the travis script, or should we update the current code like this?

Would switch to 1.13.

sttts:
In the example output in the description, the `Me likey` line is formated slightly different than the other lines. Also, why don't I see the 1000ms in the timestamp of the following spans?

SomtochiAma:
> In the example output in the description, the `Me likey` line is formated slightly different than the other lines. Also, why don't I see the 1000ms in the timestamp of the following spans?

I will update the formatting to be exact. 
The timestamp is there but is 0ms because I set the time to sleep `time.Sleep` before adding the Nested traces

sttts:
I wonder whether we have a logical problem:
- the `Step` func is sometimes used before, sometimes after a certain operation. Which duration and which timestamp is printed? I.e. do they start a span or do they end it?
- the nested spans always start when created and end when `LogIfLong` is called.

sttts:
Am not sure what the best answer is to my previous comment. Two solutions come to mind:
1. somehow adapt the logic of steps and nested spans to be uniform
2. adapt the output format to make the different clear: e.g. for `New, Step("Foo"), [NestedTrace, Step(Bar), LogIfLong], Step("abc", LogIfLong`:
```
[OuterTrace @ 23:00:00.000 for 100ms
  | Foo @ 23:00:00.010 after 10ms
  [InnerTrace @ 23:00:00.020 for 50ms
    | Bar @ 23:00:00.040 after 20ms
  ]
  | abc @ 23:00:150 after 80ms
]
```
Note that the last step counts from the span, not from the previous step. It's also not clear to me that this is super natural.

SomtochiAma:
> I wonder whether we have a logical problem:
> 
> * the `Step` func is sometimes used before, sometimes after a certain operation. Which duration and which timestamp is printed? I.e. do they start a span or do they end it?
> * the nested spans always start when created and end when `LogIfLong` is called.
> 
> Having 10ms duration for the `Me likey`, but timestamp 23:00:00.10 looks odd.
> 
> Please look at the old `LogIfLong` logic. It computes which steps dominate the overall time of the trace. Which interval to they take? The one before the step or the one after?
> 
> Another hint for the logical problem: New, Step, NestedTrace, Step, LogIfLong. Where do we account for the time between Step and NestedTrace? And for the time between NestedTrace and the last Step?

The old `logIfLong` logic divided the totalTime the trace took by the number of steps to get the stepThreshold and then any step that took longer than the stepThreshold(if the time since the last step was longer than the stepThreshold) was printed.

jpbetz:
Checking in on this one. Are we still interested in pushing this forward? I have a couple PRs that could take advantage of this.

SomtochiAma:
@jpbetz I will resume on this tomorrow. A couple of changes are still left.

jpbetz:
> @jpbetz I will resume on this tomorrow. A couple of changes are still left.

Great, I'll be happy to review. I'll look over the open issues on this today as well.

trace/trace_test.go
```diff
@@ -121,7 +121,7 @@ func TestLog(t *testing.T) {
 		{
 			name: "Check fixture formatted",
 			expectedMessages: []string{
-				"URL:/api,count:3", "msg1 str:text,int:2,bool:false", "msg2 x:1",
+				"URL:/api,count:3", `"msg1" str:text,int:2,bool:false`, `"msg2" x:1`,
 			},
 			sampleTrace: fieldsTraceFixture(),
 		},
```
jpbetz, line 127: We also need to add some new tests for nested traces somewhere.

trace/trace.go
```diff
@@ -93,29 +147,47 @@ func (t *Trace) logWithStepThreshold(stepThreshold time.Duration) {
...
-		buffer.WriteString(fmt.Sprintf("Trace[%d]: [%v] [%v] END\n", tracenum, endTime.Sub(t.startTime), stepDuration))
+		buffer.WriteString(fmt.Sprintf("\nTrace[%d]: [%v] [%v] END\n", tracenum, endTime.Sub(t.startTime), stepDuration))
 	}
 
 	klog.Info(buffer.String())
 }
 
+func writeTrace(b *bytes.Buffer, t *Trace, formatter string, stepThreshold time.Duration) time.Time {
+	lastStepTime := t.startTime
+	_ = len(t.nestedTrace)
```
jpbetz, line 162: Remove?

trace/trace.go
```diff
@@ -76,6 +123,13 @@ func (t *Trace) Step(msg string, fields ...Field) {
...
 }
 
+// Nest adds a new trace to the nestedTrace array
+func (t *Trace) Nest(msg string, fields ...Field) *Trace {
+	newTrace := New(msg, fields...)
+	t.nestedTrace = append(t.nestedTrace, newTrace)
+	return newTrace
+}
+
 // Log is used to dump all the steps in the Trace
```
jpbetz, line 133: Expand this godoc a bit to clarify how nested traces are formatted in the output?

trace/trace.go
```diff
@@ -93,29 +146,41 @@ func (t *Trace) logWithStepThreshold(stepThreshold time.Duration) {
...
 }
 
+func writeTrace(b *bytes.Buffer, t *Trace, formatter string, stepThreshold time.Duration) time.Time {
+	lastStepTime := t.startTime
+	stepAndTraces := t.sortStepTrace()
+	if len(stepAndTraces) == 0 {
+		return lastStepTime
+	}
+	for _, stepOrTrace := range stepAndTraces {
+		switch s := stepOrTrace.(type) {
```
jpbetz, line 166: An alternative would be to add a `writeStep` function to the `stepTraceTime` interface (and maybe rename the interface to something more general), and then instead of having a switch here, just call `writeStep` which will call the appropriate code on both `Trace` and `traceStep`.

trace/trace_test.go
```diff
@@ -63,6 +63,50 @@ func TestStep(t *testing.T) {
 	}
 }
 
+func TestNestedTrace(t *testing.T) {
+	tests := []struct {
+		name          string
+		inputString   string
+		expectedTrace *Trace
```
jpbetz, line 70: Let's expectedMessage and verify they are rendered as expected.

trace/trace_test.go
```diff
@@ -63,6 +63,50 @@ func TestStep(t *testing.T) {
...
+					{
+						startTime: time.Now(),
+						name:      "Inner trace",
+						nestedTrace: []*Trace{
+							{name: "Inner trace"},
+						},
+					},
+				},
+			},
+		},
```
jpbetz, line 95: I case the verifies the sorting of `traceSteps` and `Trace` works as expected seems valuable.

trace/trace.go
```diff
@@ -31,6 +32,36 @@ type Field struct {
...
+	time() time.Time
+}
+
+func (t *Trace) time() time.Time {
+	return t.startTime
+}
+
+func (s traceStep) time() time.Time {
+	return s.stepTime
+}
```
jpbetz, line 45: nit: We typically arrange receiver functions for a type together with the type declaration.

trace/trace.go
```diff
@@ -44,19 +68,61 @@ func writeFields(b *bytes.Buffer, l []Field) {
...
 type Trace struct {
-	name      string
-	fields    []Field
-	startTime time.Time
-	steps     []traceStep
+	name        string
+	fields      []Field
+	startTime   time.Time
+	steps       []traceStep
+	nestedTrace []*Trace
```
jpbetz, line 112: Any reason we need two separate arrays here, or would it work to just do `steps []stepTraceTime`?

trace/trace.go
```diff
@@ -31,6 +32,29 @@ type Field struct {
 	Value interface{}
 }
 
+type stepTraceTime interface {
```
jpbetz, line 35: Name doesn't make as much sense now. Remove `Time` from name?

trace/trace.go
SomtochiAma: I think it is still logical to separate them. Although making it one array will remove the need to sort it. @sttts any thoughts?

jpbetz: It’s fine to have a slice of mixed types. I don’t think the added complexity of keeping them separate is worth it.

SomtochiAma: Okay. That's fine
The only consideration I have seen so far is in this [function](https://github.com/kubernetes/utils/blob/master/trace/trace.go#L123) where the number of steps is used to calculate the step threshold. Wdyt?

jpbetz: This is a really good question. I think we need to be careful with thresholds and logging.

Just using the number of steps (incl nested traces) of the top level trace to calculate the step threshold like you linked to doesn't seem to make much sense. Each of the nested traces can contain many more steps. Using the current logic could easily result in nested trace steps getting unreasonable thresholds, resulting in the logs getting spammed.

There is a loosely related problem. Right now it looks like if we call `LogIfLong` on a nested trace, it logs it separate from the surrounding trace. So if we ask callers to use functions like `NewNestedInContext` (https://github.com/kubernetes/kubernetes/pull/87610/files#diff-be0415fae94fa4b972093826630556d9R104) where they don't know if they ended up creating a top level trace or a nested trace, we have a problem. The caller will typically `LogIfLong` on the trace, and as a result it get it logged twice, once when `LogIfLong` is called directly on the nested trace and again when it is called on the top level trace. The more level of nesting the worse this would be.

I think we need to keep track of the duration passed to `LogIfLong` on calls to nested traces, but don't log immediately and instead use it to calculate the thresholds. If we just ignored `LogIfLong` calls to nested traces we have to try to calculate a reasonable threshold for the nested trace and risk getting the logs spammed if we get the value wrong. Better take advantage of the hint the caller has provided.

Unfortunately, this is going to complicate the threshold calculations.

Straw-man proposal:

  - Find the nested traces that have a threshold provided to them by a call to LogIfLong
  - Subtract the nested trace thresholds from the threshold of the current trace to calculate a remaining threshold for the steps of the current trace (but always using at least some ratio of the current trace threshold when calculating the stepThreshold, e.g. no less than 1/4)
  - Divide the remaining threshold evenly among the steps of the trace like we do today
  - If a nested trace does not have it's own threshold provided by a call to LogIfLong then
    treat it as a step and give it an even division of the threshold
  - Don't pass stepThreshold into nested traces. Each nested trace calculates its own stepThreshold.

Suggestions on how to better do this welcome.

SomtochiAma: I think the first hurdle is how to know if LogIfLong has been called on a trace. We could do this by adding an extra field in the struct `threshold` so LogIfLong sets this field and we can easily access it when logging the trace.

Also, I am not clear on how we want the trace to behave when LogIfLong has been called both on the parent trace and the nested trace. Do we want them to log separately or not at all?

jpbetz: > I think the first hurdle is how to know if LogIfLong has been called on a trace. We could do this by adding an extra field in the struct `threshold` so LogIfLong sets this field and we can easily access it when logging the trace.

I think this is needed. It can be pointer. If the pointer is nil, we know LogIfLong has not been called.

> Also, I am not clear on how we want the trace to behave when LogIfLong has been called both on the parent trace and the nested trace. Do we want them to log separately or not at all?

I tried to answer this in my above comment. I think we should consider this approach:

- We only log when LogIfLong is called on the top level trace.
- Calls to LogIfLong on nested traces simply keep track of the threshold.
- When actually logging the trace (when LogIfLong is called on the top level trace), log any nested traces that exceed their threshold even if the traces they are nested in do not exceed their threshold.

The one downside to this approach is that if a nested trace exceeds its threshold but the trace it is nested under never calls LogIfLong, then the nested trace won't ever get logged. In practice LogIfLong should be called in a defer so this shouldn't be a problem, but it needs to be documented.

trace/trace.go
```diff
@@ -34,25 +33,7 @@ type Field struct {
...
-	for _, trace := range t.nestedTrace {
-		arr = append(arr, trace)
-	}
-
-	sort.Slice(arr, func(i, j int) bool {
-		return arr[i].time().Before(arr[j].time())
-	})
-
-	return arr
+	writeStep(b *bytes.Buffer, formatter string, t *Trace, stepDuration time.Duration, lastStepTime time.Time) time.Time
```
jpbetz, line 36: rename `t` to `parentTrace` here too for clarity?

trace/trace.go
```diff
@@ -107,22 +88,32 @@ func (s traceStep) writeStep(b *bytes.Buffer, formatter string, stepThreshold ti
...
-	b.WriteString("]")
-	return time.Time{}
+	stepDuration := t.startTime.Sub(lastStepTime)
+	if stepThreshold == 0 || stepDuration > stepThreshold || klog.V(4) {
+		if t.threshold != nil {
+			stepThreshold = *parentTrace.threshold - *t.threshold
+			limitThreshold := *parentTrace.threshold / 4
+			if stepThreshold < limitThreshold {
+				stepThreshold = limitThreshold
+			}
```
jpbetz, line 109: I'm not following the logic behind this math (and it look different than what I had expected based on discussion). Explain somewhere?

trace/trace.go
```diff
@@ -178,25 +169,23 @@ func (t *Trace) logWithStepThreshold(stepThreshold time.Duration) {
...
 }
 
 // LogIfLong is used to dump steps that took longer than its share
 func (t *Trace) LogIfLong(threshold time.Duration) {
+	t.threshold = &threshold
 	if time.Since(t.startTime) >= threshold {
 		// if any step took more than it's share of the total allowed time, it deserves a higher log level
-		stepThreshold := threshold / time.Duration(len(t.steps)+1)
+		stepThreshold := threshold / time.Duration(len(t.stepsTraces)+1)
 		t.logWithStepThreshold(stepThreshold)
```
jpbetz, line 189: I wasn't expecting nested traces to get logged. https://github.com/kubernetes/utils/pull/139#discussion_r405783089

trace/trace_test.go
```diff
@@ -246,11 +245,11 @@ func TestNestedTraceLog(t *testing.T) {
...
-					{
+				stepsTraces: []stepTrace{
+					&Trace{
 						startTime:   time.Now(),
 						name:        "msg1",
-						nestedTrace: []*Trace{{name: "nested1", startTime: time.Now()}},
+						stepsTraces: []stepTrace{&Trace{name: "nested1", startTime: time.Now()}},
 					},
 				},
 			},
```
jpbetz, line 255: We need nested trace cases in `TestLogIfLong` (or in a new test function that test `LogIfLong` for nested traces).

trace/trace.go
SomtochiAma: >Subtract the nested trace thresholds from the threshold of the current trace to calculate a >remaining threshold for the steps of the current trace
line 105: subtracts the threshold of the trace from that of the parent trace

>less than 1/4 of the current trace threshold is used to calculate the stepThreshold for the >steps that are direct children of the trace
line 106 - 108: if the resulting threshold(from the calculation) is less than 1/4 of theparent trace. It uses the 1/4 of the parent trace instead

Did I miss it somewhere?

trace/trace_test.go
SomtochiAma: Okay, I will add those

trace/trace.go
SomtochiAma: So we should use only steps for calculating the stepThreshold? Should I ping you on slack?

trace/trace.go
jpbetz: Just to capture what we discussed offline:

Say we have something like:

```
Trace1: threshold=1000ms (provided by call to LogIfLong)
  step1.1
  Trace2: threshold=600ms (provided by call to LogIfLong)
  step1.2
  Trace3: (no threshold provided)
```

My thought was that the effective thresholds of step1.1, step1.2 and Trace3 might be calculated based on the threshold of Trace1 and Trace2. Since Trace2 accounts for 600ms of the 1000ms of Trace1, I would expect the remaining time to then be divided up evenly, e.g.:

```
Trace1: threshold=1000ms
  step1.1: stepThreshold=100ms
  Trace2: threshold=600ms
  step1.2: stepThreshold=100ms
  Trace3: threshold=100ms
```

Where there is 100ms still left for Trace1 directly (which is why, I assume we do the `+1` in `len(steps)+1` already).

trace/trace.go
```diff
@@ -93,39 +152,69 @@ func (t *Trace) logWithStepThreshold(stepThreshold time.Duration) {
...
+	traceThreshold := *t.threshold
+	for _, s := range t.stepsTraces {
+		nestedTrace, ok := s.(*Trace)
+		if ok && nestedTrace.threshold != nil {
+			traceThreshold = traceThreshold - *nestedTrace.threshold
+			lenTrace--
+		}
+	}
+
+	limitThreshold := *t.threshold / 4
```
jpbetz, line 212: We need a comment here summarizing why we have a limitThreshold.

trace/trace.go
```diff
@@ -93,39 +152,69 @@ func (t *Trace) logWithStepThreshold(stepThreshold time.Duration) {
...
+		lastStepTime = stepTime
+	}
+	return lastStepTime
+}
+
 // LogIfLong is used to dump steps that took longer than its share
 func (t *Trace) LogIfLong(threshold time.Duration) {
-	if time.Since(t.startTime) >= threshold {
+	t.threshold = &threshold
+	if time.Since(t.startTime) >= threshold || t.parentTrace == nil {
```
jpbetz, line 181: Should be `&&` instead of `||` ?

trace/trace.go
```diff
@@ -93,39 +152,69 @@ func (t *Trace) logWithStepThreshold(stepThreshold time.Duration) {
...
 		// if any step took more than it's share of the total allowed time, it deserves a higher log level
-		stepThreshold := threshold / time.Duration(len(t.steps)+1)
+		stepThreshold := calculateStepThreshold(t)
 		t.logWithStepThreshold(stepThreshold)
+	} else {
+		for _, s := range t.stepsTraces {
+			nestedTrace, ok := s.(*Trace)
+			if ok && nestedTrace.threshold != nil && time.Since(nestedTrace.startTime) >= *nestedTrace.threshold {
+				stepThreshold := calculateStepThreshold(nestedTrace)
+				nestedTrace.logWithStepThreshold(stepThreshold)
```
jpbetz, line 190: Does this just goes one level deep in the nested traces? I was expecting any nested trace that needs to be logged would get logged directly if none of its parents get logged, no matter how deeply nested. Good thing to have a test case for.

SomtochiAma: The function is recursive so it should work. I will write some tests for it.

jpbetz: Okay, I was expecting the for loop surrounding this call to logWithStepThreshold to itself be in logWithStepThreshold for a recursive approach, but maybe I don't follow how the recursion works.

SomtochiAma: Oh true. I missed that. Thank you

jpbetz:
@sttts @thockin, this one is coming along. Would either of you be willing to give it a review pass?

jpbetz:
@sttts @thockin Friendly nudge for review. I'm pretty happy with this one at this point.

trace/trace.go
```diff
@@ -44,19 +44,82 @@ func writeFields(b *bytes.Buffer, l []Field) {
...
+func durationToMilliseconds(timeDuration time.Duration) int64 {
+	return timeDuration.Nanoseconds() / 1e6
+}
+
+type traceItem interface {
+	// time returns when the trace was recorded as completed.
+	time() time.Time
+	// writeItem outputs the traceItem to the buffer. If stepThreshold is non-nil, only output the
+	// traceItem if its the duration exceeds the stepThreshold.
+	writeItem(b *bytes.Buffer, formatter string, startTime time.Time, stepThreshold *time.Duration)
```
sttts, line 66: doc what formatter is

trace/trace.go
```diff
@@ -69,63 +132,121 @@ func New(name string, fields ...Field) *Trace {
...
+	t.traceItems = append(t.traceItems, traceStep{stepTime: time.Now(), msg: msg, fields: fields})
 }
 
-// Log is used to dump all the steps in the Trace
-func (t *Trace) Log() {
-	// an explicit logging request should dump all the steps out at the higher level
-	t.logWithStepThreshold(0)
+// Nest adds a nested trace with the given message and fields and returns it.
+func (t *Trace) Nest(msg string, fields ...Field) *Trace {
+	newTrace := New(msg, fields...)
```
sttts, line 144: for convenience we might want to handle the `t==nil` case too, i.e. no new parent. Then you can write `GetTraceFromContext(ctx).Nest` without case distinction between nil or non-nil.

sttts: only requires to skip the append below if I see it correctly.

trace/trace.go
```diff
@@ -69,63 +132,121 @@ func New(name string, fields ...Field) *Trace {
...
-			buffer.WriteString("\n")
+	t.Log()
+}
+
+// logTopLevelTraces finds all traces in a hierarchy of nested traces that should be logged but do not have any
+// parents that will be logged, due to threshold limits, and logs them as top level traces.
+func (t *Trace) logTrace() {
+	if t.durationIsWithinThreshold() {
+		var buffer bytes.Buffer
+		tracenum := rand.Int31()
```
sttts, line 179: nit: camlCase

trace/trace.go
```diff
@@ -69,63 +132,121 @@ func New(name string, fields ...Field) *Trace {
...
+		t.writeTraceSteps(&buffer, fmt.Sprintf("\nTrace[%d]: ", tracenum), stepThreshold)
+		buffer.WriteString(fmt.Sprintf("\nTrace[%d]: [%v] [%v] END\n", tracenum, t.endTime.Sub(t.startTime), totalTime))
+
+		klog.Info(buffer.String())
+		return
 	}
-	stepDuration := endTime.Sub(lastStepTime)
-	if stepThreshold == 0 || stepDuration > stepThreshold || klog.V(4).Enabled() {
-		buffer.WriteString(fmt.Sprintf("Trace[%d]: [%v] [%v] END\n", tracenum, endTime.Sub(t.startTime), stepDuration))
+	// If the trace should not be logged, still check if nested traces should be logged
```
sttts, line 197: nit: empty line above

trace/trace.go
```diff
@@ -69,63 +132,121 @@ func New(name string, fields ...Field) *Trace {
...
 }
 
 // TotalTime can be used to figure out how long it took since the Trace was created
 func (t *Trace) TotalTime() time.Duration {
 	return time.Since(t.startTime)
 }
+
+// calculateStepThreshold returns a threshold for the individual steps of a trace, or nil if there is no threshold and
+// all steps should be written.
+// TODO: If Log is called instead of LogIfLong, we should print all steps, right? This should be documented clearly.
```
sttts, line 227: isn't it documented?

trace/trace.go
jpbetz: That would be handy. Adding.

trace/trace.go
jpbetz: Yes. This shouldn't have been committed. I'll remove it.

sttts:
Am just waiting for some updated example output in the PR description which matches the code. Then let's ship it 🚀

jpbetz:
@sttts would you tag if this looks okay to you? Thanks!

sttts:
Ship it :ship:

jpbetz:
Thanks @SomtochiAma!!

trace/trace.go
```diff
@@ -69,63 +134,145 @@ func New(name string, fields ...Field) *Trace {
...
+	lastStepTime := t.startTime
+	for _, stepOrTrace := range t.traceItems {
+		stepOrTrace.writeItem(b, formatter, lastStepTime, stepThreshold)
+		lastStepTime = stepOrTrace.time()
+	}
 }
 
-// LogIfLong is used to dump steps that took longer than its share
-func (t *Trace) LogIfLong(threshold time.Duration) {
-	if time.Since(t.startTime) >= threshold {
```
liggitt, line 121: this PR changed the behavior of LogIfLong to unconditionally log all steps of all traces if verbosity is at V(4) or higher

cc @jpbetz

sttts: > The prior behavior of logging all steps of a trace that exceeded the threshold is more useful, in my opinion.

+1

dims: @sttts can you please file the change needed in `k8s.io/utils`?

sttts: Looking into it

sttts: https://github.com/kubernetes/utils/pull/177

liggitt: https://github.com/kubernetes/kubernetes/pull/93253

Issue #142, opened 2020-02-29
Introduce a cross-platform typecheck for utils repo


saad-ali:
https://github.com/kubernetes/utils/pull/138 introduced a build error for `mount_windows.go` that was not caught until attempt to integrate with k/k in https://github.com/kubernetes/kubernetes/pull/88684

We should have CI for this utils repo that catches this before merge in the future.

PR #143, opened 2020-02-29
Fix mount_windows build error

saad-ali:
https://github.com/kubernetes/utils/pull/138 introduced a build error for `mount_windows.go`
This PR fixes that.
This was caught while trying to integrate with k/k in https://github.com/kubernetes/kubernetes/pull/88684
Opened  https://github.com/kubernetes/utils/issues/142 to track work to prevent such future regressions.

### DIFF
--- a/mount/mount_windows.go
+++ b/mount/mount_windows.go
@@ -80,7 +80,7 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
 	bindSource := source
 
 	// tell it's going to mount azure disk or azure file according to options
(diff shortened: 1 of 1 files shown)


saad-ali:
Prior to this fix:
```
$ GOOS=windows GOARCH=amd64 go install
# k8s.io/utils/mount
./mount_windows.go:83:19: assignment mismatch: 4 variables but MakeBindOpts returns 3 values
./mount_windows.go:83:34: too many arguments in call to MakeBindOpts
	have ([]string, []string)
	want ([]string)
```
After this fix:
```
$ GOOS=windows GOARCH=amd64 go install
```

/assign @liggitt

Issue #141, opened 2020-02-28
[mount] Addition of "checkAndRepairXfsFilesystem" inadvertently prevents XFS self-recovery via mounting


nktpro:
[PR #126 ](https://github.com/kubernetes/utils/pull/126) added an extra step to run `xfs_repair` before mounting a XFS file system. However instead of helping to automatically correct FS issues due to prior unclean shutdowns, it actually prevented auto recovery from happening, which led to complete unavailability of the corresponding volume and subsequently required manual human intervention.

The sequence of events is as follows:

1. A node loss / unclean shutdown occurs.
2. A stateful pod is restarted on another healthy node; its volume is re-attached to the new node.
3. xfs_repair is run against the volume. The relevant logs would look like these (in the context of rook-ceph but should apply to any other user of the mounter)

```
Filesystem corruption was detected for /dev/rbd1, running xfs_repair to repair
ID: 29 Req-ID: 0001-0009-rook-ceph-0000000000000001-9adb43bf-4e25-11ea-aa19-2ecc193be507 failed to mount device path (/dev/rbd1) to staging path (/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-423b5a86-7c03-43c4-a7e9-4921934016de/globalmount/0001-0009-rook-ceph-0000000000000001-9adb43bf-4e25-11ea-aa19-2ecc193be507) for volume (0001-0009-rook-ceph-0000000000000001-9adb43bf-4e25-11ea-aa19-2ecc193be507) error 'xfs_repair' found errors on device /dev/rbd1 but could not correct them: Phase 1 - find and verify superblock...

Phase 2 - using internal log
        - zero log...
ERROR: The filesystem has valuable metadata changes in a log which needs to
be replayed.  Mount the filesystem to replay the log, and unmount it before
re-running xfs_repair.  If you are unable to mount the filesystem, then use
the -L option to destroy the log and attempt a repair.
Note that destroying the log may cause corruption -- please attempt a mount
of the filesystem before doing this.
```

4. The volume is then prevented from being mounted and manual intervention is required
5. All that's needed then, is to manually mount the volume, which replay XFS logs automatically, then unmount it and restart the corresponding pod.

Note that step #5 was what has always happened prior to this change. The volume is simply mounted without any attempt to perform FS check / xfs_repair. It can simply correct itself as part of just being mounted, as per XFS design.

[...]

PR #144, opened 2020-03-01
Ignore xfs dirty logs

27149chen:
This PR try to fix the urgent issue #141  by ignoring the xfs dirty logs issue (do not replay dirty logs by default), and letting subsequent operations handle it.
But this is not a perfect solution, more discussion is in pr #132 .

### DIFF
--- a/mount/mount_linux.go
+++ b/mount/mount_linux.go
@@ -43,6 +43,10 @@ const (
 	fsckErrorsCorrected = 1
 	// 'fsck' found errors but exited without correcting them
 	fsckErrorsUncorrected = 4
+	// 'xfs_repair' found errors but exited without correcting them
+	xfsRepairErrorsUncorrected = 1
+	// 'xfs_repair' was unable to proceed due to a dirty log
+	xfsRepairErrorsDirtyLogs = 2
 )
 
 // Mounter provides the default implementation of mount.Interface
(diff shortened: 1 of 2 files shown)


27149chen:
@saad-ali , @gnufied  please have a look.

mount/mount_linux.go
```diff
@@ -330,12 +334,25 @@ func (mounter *SafeFormatAndMount) checkAndRepairXfsFilesystem(source string) er
...
+				return NewMountError(HasFilesystemErrors, "failed to run 'xfs_repair' on device %s: %v\n", source, err)
+
+			}
+			switch e.ExitStatus() {
+			case xfsRepairErrorsUncorrected:
+				return NewMountError(HasFilesystemErrors, "'xfs_repair' found errors on device %s but could not correct them: %s\n", source, string(out))
+			case xfsRepairErrorsDirtyLogs:
+				// mark the a dirty log issue as a warning, do you replay it by default.
+				klog.Warningf("A dirty log is detected on this xfs disk.")
+				return nil
```
whymatter: Hi I came across this pr because I do have exactly this problem using ceph rbd with xfs and rook.io 
Wouldn't this cause the same potential "compounding of problems" as in #132 ?

27149chen: yes, I will. But 132 is still under discussion.

PR #146, opened 2020-03-13
Move k8s.io/apiserver/pkg/util/term/term.go to k8s.io/utils/term/

dims:
- terminal related stuff should NOT be in apiserver
- earlier this used docker/docker, now we have a stripped down moby/term
that we can use
- usages:
  https://cs.k8s.io/?q=TerminalSize%5C(&i=nope&files=cmd%2F.*&repos=kubernetes/kubernetes


dims:
cc @mattjmcnaughton

thockin:
IMO this is pretty borderline.  It's a fairly trivial block of code (shorter than some examples found in doc-comments).  We're encouraging people to take a dependency on k8s.io/utils to get that, which seems to violate the "a little copying is better than a little dependency" mindset.

Perhaps more importantly, we're imposing this Moby dep on everyone who already uses k8s.io/utils, and that feels wrong.  Especially the size of the transitive deps.

Put another way, if I said you have to make a new repo for this, would you say "OK", or would you say "not worth it"?

We don't have strong guidelines on this, but we should.  I lean towards this being a net negative.

liggitt:
I agree. Taking on all the deps that showed up in our go.sum file for a single helper method is not worth it.

liggitt:
I'd be comfortable recommending anyone that wanted to use this having this method as a private method in their package instead and depend directly on the moby/term dependency

dims:
@thockin @liggitt currently we are carrying this `TerminalSize` in

liggitt:
if we're using it to output help in component-base CLI utilities, I could see moving it there

PR #148, opened 2020-03-20
Fix windows MountSensitive error

mboersma:
After kubernetes/kubernetes#88684 was included in 1.18.0-rc.1, Windows volume mounts began failing in our AKS Engine e2e tests. It appears that the code path for Windows `MountSensitive` should have been looking at the `sensitiveOptions` instead of `options`. We see `options` empty in every call.

We've tested this interactively and found it to resolve Windows mounting problems.

cc: @kkmsft @marosset

### DIFF
--- a/mount/mount_windows.go
+++ b/mount/mount_windows.go
@@ -79,14 +79,15 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
 		sanitizedOptionsForLogging, source, target, fstype)
 	bindSource := source
 
-	// tell it's going to mount azure disk or azure file according to options
 	if bind, _, _, _ := MakeBindOptsSensitive(options, sensitiveOptions); bind {
-		// mount azure disk
 		bindSource = NormalizeWindowsPath(source)
 	} else {
-		if len(options) < 2 {
+		allOptions := []string{}
+		allOptions = append(allOptions, options...)
+		allOptions = append(allOptions, sensitiveOptions...)
+		if len(allOptions) < 2 {
 			klog.Warningf("mount options(%q) command number(%d) less than 2, source:%q, target:%q, skip mounting",
-				sanitizedOptionsForLogging, len(options), source, target)
+				sanitizedOptionsForLogging, len(allOptions), source, target)
 			return nil
 		}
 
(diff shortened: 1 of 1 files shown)


mboersma:
This is a quick fix--let's let @kkmsft weigh in because he may have a smarter one.

andyzhangx:
@gnufied coud you approve this? Thanks.

mount/mount_windows.go
```diff
@@ -84,7 +84,7 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
 		// mount azure disk
 		bindSource = NormalizeWindowsPath(source)
 	} else {
-		if len(options) < 2 {
+		if len(sensitiveOptions) < 2 {
```
liggitt: /hold

options can be in options or sensitiveOptions... I would expect length checking to consider the total length of options+sensitiveOptions.

mount/mount_windows.go
```diff
@@ -99,13 +99,13 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
 		getSMBMountMutex.LockKey(source)
 		defer getSMBMountMutex.UnlockKey(source)
 
-		if output, err := newSMBMapping(options[0], options[1], source); err != nil {
+		if output, err := newSMBMapping(sensitiveOptions[0], sensitiveOptions[1], source); err != nil {
```
liggitt: Nothing in the documentation indicates the order or distribution between options/sensitiveOptions is important... assuming a particular option will be in one or the other is fairly fragile. This was raised in https://github.com/kubernetes/utils/pull/138#discussion_r378635098

liggitt: given the order and distribution of options *is* important here, I would expect something more like this:

```go
allOptions := []string{}
allOptions = append(allOptions, options...)
allOptions = append(allOptions, sensitiveOptions...)
if len(allOptions) < 2 {
...
}
...
if output, err := newSMBMapping(allOptions[0], allOptions[1], source); err != nil {
...
}
```

and to document that `sensitiveOptions` are always placed after `options`

liggitt:
was there a presubmit job that could have been run against https://github.com/kubernetes/kubernetes/pull/88684 to catch this before merge?

mount/mount_windows.go
liggitt: (and don't log the content of allOptions anywhere, since it can contain sensitive options)

msau42: Would it be better to have a map and search for specific option keys rather than assuming the options are passed in a specific order?

mount/mount_windows.go
```diff
@@ -84,7 +84,7 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
 		// mount azure disk
```
msau42: tiny nit, remove implementation specific comments from this generic library.

mount/mount_windows.go
liggitt: iiuc, these options don't have identifying keys

mboersma:
Thanks for the comments on this hasty PR. I'll make changes and re-test locally.

We think reverting just the azure_file.go changes in kubernetes/kubernetes#88684 would be a safer approach, and buy more time to fix this cleanly and to add unit tests.

mount/mount_windows.go
msau42: they appear to be passed in the form "user=abc", "password=abc", so we could at least try to parse and search for the correct option?

liggitt:
go ahead and squash to a single commit

PR #149, opened 2020-03-24
fix for no Go source files on macOS

dims:
Make things a bit more easier on macOS
```
~/go/src/github.com/google/cadvisor
$ go list all
build k8s.io/utils/inotify: cannot load k8s.io/utils/inotify: no Go source files
```

### DIFF
--- a/inotify/inotify_others.go
+++ b/inotify/inotify_others.go
@@ -1,0 +1,54 @@
+// +build !linux
+
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


inotify/inotify_others.go
```diff
@@ -0,0 +1,52 @@
...
+package inotify // import "k8s.io/utils/inotify"
+
+import (
+	"fmt"
+	"runtime"
+)
+
+// NewWatcher creates and returns a new inotify instance using inotify_init(2)
+func NewWatcher() (*Watcher, error) {
+	return nil, fmt.Errorf("watch not supported on %s", runtime.GOOS)
```
neolit123: possibly the error can shared across all functions.

```
var notSupportedError = fmt.Errorf("watch not supported on %s", runtime.GOOS)
...

func NewWatcher() (*Watcher, error) {
	return nil, notSupportedError 
}
...
```

dims: DONE!

neolit123:
> Errors from golint:
inotify/inotify_others.go:26:5: error var notSupportedError should have name of the form errFoo

my bad, looks like it wants `errNotSupported` instead.

PR #150, opened 2020-03-24
Do not perform xfs_repair on xfs filesystem

gnufied:
This partly reverts https://github.com/kubernetes/utils/pull/126 but still we perform filesystem repair just before mounting. 

fixes https://github.com/kubernetes/utils/issues/141

### DIFF
--- a/mount/mount_linux.go
+++ b/mount/mount_linux.go
@@ -314,33 +314,6 @@ func (mounter *SafeFormatAndMount) checkAndRepairFilesystem(source string) error
 	return nil
 }
 
-// checkAndRepairXfsFilesystem checks and repairs xfs filesystem using command xfs_repair.
-func (mounter *SafeFormatAndMount) checkAndRepairXfsFilesystem(source string) error {
-	klog.V(4).Infof("Checking for issues with xfs_repair on disk: %s", source)
-
-	args := []string{source}
-	checkArgs := []string{"-n", source}
-
-	// check-only using "xfs_repair -n", if the exit status is not 0, perform a "xfs_repair"
-	_, err := mounter.Exec.Command("xfs_repair", checkArgs...).CombinedOutput()
-	if err != nil {
-		if err == utilexec.ErrExecutableNotFound {
-			klog.Warningf("'xfs_repair' not found on system; continuing mount without running 'xfs_repair'.")
-			return nil
-		} else {
-			klog.Warningf("Filesystem corruption was detected for %s, running xfs_repair to repair", source)
-			out, err := mounter.Exec.Command("xfs_repair", args...).CombinedOutput()
-			if err != nil {
-				return NewMountError(HasFilesystemErrors, "'xfs_repair' found errors on device %s but could not correct them: %s\n", source, out)
-			} else {
-				klog.Infof("Device %s has errors which were corrected by xfs_repair.", source)
-				return nil
-			}
-		}
-	}
-	return nil
-}
-
 // formatAndMount uses unix utils to format and mount the given disk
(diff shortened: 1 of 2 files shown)


gnufied:
cc @jsafrane @msau42

mount/mount_linux.go
```diff
@@ -410,14 +383,7 @@ func (mounter *SafeFormatAndMount) formatAndMountSensitive(source string, target
...
 			// Run check tools on the disk to fix repairable issues, only do this for formatted volumes requested as rw.
-			var err error
-			switch existingFormat {
-			case "xfs":
-				err = mounter.checkAndRepairXfsFilesystem(source)
-			default:
-				err = mounter.checkAndRepairFilesystem(source)
-			}
-
+			err := mounter.checkAndRepairFilesystem(source)
```
gnufied: code before change:

```
	if !readOnly {
		// Run fsck on the disk to fix repairable issues, only do this for volumes requested as rw.
		klog.V(4).Infof("Checking for issues with fsck on disk: %s", source)
		args := []string{"-a", source}
		out, err := mounter.Exec.Command("fsck", args...).CombinedOutput()
		if err != nil {
			ee, isExitError := err.(utilexec.ExitError)
			switch {
			case err == utilexec.ErrExecutableNotFound:
				klog.Warningf("'fsck' not found on system; continuing mount without running 'fsck'.")
			case isExitError && ee.ExitStatus() == fsckErrorsCorrected:
				klog.Infof("Device %s has errors which were corrected by fsck.", source)
			case isExitError && ee.ExitStatus() == fsckErrorsUncorrected:
				return NewMountError(HasFilesystemErrors, "'fsck' found errors on device %s but could not correct them: %s", source, string(out))
			case isExitError && ee.ExitStatus() > fsckErrorsUncorrected:
				klog.Infof("`fsck` error %s", string(out))
			}
		}
	}
```
and hence should be same as before.

saad-ali: The code that is part of `checkAndRepairFilesystem(...)` used to be called before `// Check if the disk is already formatted`. As part of https://github.com/kubernetes/utils/pull/126 it was moved down here. Should we revert this change as well or is it necessary for some reason?

gnufied: We could revert that change too but I figured it is harmless and helps to keep diff of this PR small. We will have to fix all the breaking tests because order in which `fsck` gets called will change. 

Also it is perfectly fine to *not* run `fsck` on a unformatted disk.

saad-ali: Ack. Sounds good to me. @msau42 @xing-yang @jsafrane any objections?

xing-yang: Sounds good to me.

dims:
/approve
/hold

@xing-yang @saad-ali (others!) feel free to remove hold when this is ready

mount/mount_linux.go
msau42: lgtm. The bug report had the error past the fs format check, so it's not the fs format check that's failing.

saad-ali: > The bug report had the error past the fs format check, so it's not the fs format check that's failing.

@gnufied can you confirm?

gnufied: which bug report? https://github.com/kubernetes/utils/issues/141 or https://github.com/kubernetes/utils/issues/125 ? I assume #141 - yes the problem isn't format checking but problem was that repairing the disk actually prevented volume from being mounted and recovery (although in some cases XFS volume can't recover itself and requires `xfs_repair`).

So, the only valid reason I did not do a full revert is because I liked not having to do a `fsck` if disk is unformatted.

saad-ali: Ack. SGTM

Issue #153, opened 2020-04-10
investigate windows unit test failures


dims:
/kind feature
/help
/good-first-issue

See a log file here https://github.com/dims/utils/runs/576625118?check_suite_focus=true

```
2020-04-10T11:04:24.4738710Z === RUN   TestExecutorWithArgs
2020-04-10T11:04:24.4799121Z ##[error]    TestExecutorWithArgs: exec_test.go:83: expected success, got exec: "/bin/sh": file does not exist
2020-04-10T11:04:24.4803979Z ##[error]    TestExecutorWithArgs: exec_test.go:86: unexpected output: ""
2020-04-10T11:04:24.4804404Z --- FAIL: TestExecutorWithArgs (0.03s)
2020-04-10T11:04:24.4809272Z === RUN   TestTimeout
2020-04-10T11:04:24.4882836Z ##[error]    TestTimeout: exec_test.go:141: expected context deadline exceeded but got exit status 1
2020-04-10T11:04:24.4884120Z --- FAIL: TestTimeout (0.02s)
2020-04-10T11:04:24.4884602Z === RUN   TestSetEnv
2020-04-10T11:04:24.4903444Z ##[error]    TestSetEnv: exec_test.go:150: expected success, got exec: "/bin/sh": file does not exist
2020-04-10T11:04:24.4905427Z ##[error]    TestSetEnv: exec_test.go:153: unexpected output: ""
2020-04-10T11:04:24.4907085Z ##[error]    TestSetEnv: exec_test.go:160: expected success, got exec: "/bin/sh": file does not exist
2020-04-10T11:04:24.4918387Z ##[error]    TestSetEnv: exec_test.go:163: unexpected output: ""
2020-04-10T11:04:24.4936454Z --- FAIL: TestSetEnv (0.00s)
2020-04-10T11:04:24.4936662Z === RUN   TestStdIOPipes
2020-04-10T11:04:24.4937773Z ##[error]    TestStdIOPipes: exec_test.go:190: expected Start() not to error, got: exec: "/bin/sh": file does not exist
2020-04-10T11:04:24.4939120Z ##[error]    TestStdIOPipes: exec_test.go:194: expected StdOut to be 'OUT
2020-04-10T11:04:24.5021977Z         ', got: ''
2020-04-10T11:04:24.5023434Z ##[error]    TestStdIOPipes: exec_test.go:198: expected StdErr to be 'ERR
2020-04-10T11:04:24.5040816Z         ', got: ''
[...]

PR #154, opened 2020-04-10
Switch to GitHub actions for better cross platform testing

dims:
Sets us up in a good position to add support for windows based testing later ( need https://github.com/kubernetes/utils/issues/153 to be fixed first) as we are facing issues as mentioned in https://github.com/kubernetes/utils/issues/142

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,31 @@
+name: Test
+on: [push, pull_request]
+jobs:
+  test:
+    strategy:
+      matrix:
+        go-versions: [1.12.x, 1.13.x, 1.14.x]
+        platform: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.platform }}
+    steps:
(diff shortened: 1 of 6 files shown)


dims:
cc @neolit123

.github/workflows/test.yml
```diff
@@ -0,0 +1,31 @@
+name: Test
+on: [push, pull_request]
+jobs:
+  test:
+    strategy:
+      matrix:
+        go-versions: [1.12.x, 1.13.x, 1.14.x]
+        platform: [ubuntu-latest, macos-latest]
```
neolit123, line 8: is windows-latest failing here too?
i'd enable it to know whether we need to log issue.

dims: @neolit123 i filed one already following your lead https://github.com/kubernetes/utils/issues/153

neolit123: ack, thanks

PR #160, opened 2020-04-15
Moving flag from k8s.io/kubernetes/pkg/util/flag

cici37:


flag/flags.go
```diff
@@ -0,0 +1,161 @@
...
+
+import (
+	"fmt"
+	"net"
+	"strconv"
+
+	"github.com/spf13/pflag"
+	"k8s.io/klog"
+
+	utilnet "k8s.io/apimachinery/pkg/util/net"
```
liggitt, line 27: thanks for the PR, but things in this repo cannot depend on other Kubernetes repos (see https://github.com/kubernetes/utils#criteria-for-adding-code-here)

cici37: Thanks for the tip. Will modify for sure. Sorry I thought a draft PR would not bother others. Should I close this and reopen when everything is ready or it is ok to keep it here?

liggitt: it's not clear to me that the dependencies that come along with this are ones we want in k8s.io/utils... was there an issue that prompted this move?

cici37: This is part of this [issue](https://github.com/kubernetes/cloud-provider/issues/29). I will close this pr and move the dependency to somewhere else. Thanks

PR #161, opened 2020-04-21
Fix OverlayFS mountpoint check for files

kvaps:
This PR disables `IsLikelyNotMountPoint()` check for files, because it might give a false positive result in case if `/var/lib/kubelet` is located on overlayfs filesystem, eg:

```bash
mkdir /tmp/overlayfs-test

mkdir /tmp/overlayfs-test/{upper,work,mount}

sudo mount -t overlay -o rw,relatime,lowerdir=/,upperdir=/tmp/overlayfs-test/upper/,workdir=/tmp/overlayfs-test/work/ none /tmp/overlayfs-test/mount/

mkdir -p /tmp/overlayfs-test/mount/some/dir/

touch /tmp/overlayfs-test/mount/some/dir/some_file
```

Then try simulate `IsLikelyNotMountPoint()` check:
```console
sh-5.0$ df -h /tmp/overlayfs-test/mount/some/dir/some_file
Filesystem      Size  Used Avail Use% Mounted on
tmpfs           3.7G   16K  3.7G   1% /tmp

sh-5.0$ df -h /tmp/overlayfs-test/mount/some/dir/
Filesystem      Size  Used Avail Use% Mounted on
none            3.7G   16K  3.7G   1% /tmp/overlayfs-test/mount
```

Since devices for file and for parent directory are different it will retrurn `false` (means is mountpoint), but it is actually doesn't:

```console
sh-5.0$ readlink -f /tmp/overlayfs-test/mount/some/dir/some_file 
/tmp/overlayfs-test/mount/some/dir/some_file

sh-5.0$ mountpoint /tmp/overlayfs-test/mount/some/dir/some_file 
/tmp/overlayfs-test/mount/some/dir/some_file is not a mountpoint
```

I can reproduce it on `4.15.18-24-pve` and `5.5.2-arch1-1` kernels

---

This PR completely fixes https://github.com/kubernetes/kubernetes/issues/89181#issuecomment-617326253

We also should consider reverting https://github.com/kubernetes/utils/pull/147, because it does not solves the original bug.

I've tested both cases with reverting and without it. Both are working fine to me.

### DIFF
--- a/mount/mount_linux.go
+++ b/mount/mount_linux.go
@@ -261,6 +261,11 @@ func (mounter *Mounter) IsLikelyNotMountPoint(file string) (bool, error) {
 	if err != nil {
 		return true, err
 	}
+	// This check is not applicable to files due to the fact that some filesystems like
+	// OverlayFS may show different devices for the file and its parent directory
+	if !stat.IsDir() {
+		return true, nil
+	}
 	rootStat, err := os.Stat(filepath.Dir(strings.TrimSuffix(file, "/")))
 	if err != nil {
 		return true, err


mount/mount_linux.go
```diff
@@ -261,6 +261,11 @@ func (mounter *Mounter) IsLikelyNotMountPoint(file string) (bool, error) {
 	if err != nil {
 		return true, err
 	}
+	// This check is not applicable to files due to the fact that some filesystems like
+	// OverlayFS always show different devices for the file and its parent directory
+	if !stat.IsDir() {
```
msau42, line 266: I'm not sure I understand how this fixes overlay fs. You can have bind-mounted files, and that should still be considered a mount point.

kvaps: Problem is that `IsLikelyNotMountPoint()` check might give a false positive result if `/var/lib/kubelet` is located on overlayfs filesystem, eg:

```bash
mkdir /tmp/overlayfs-test

mkdir /tmp/overlayfs-test/{upper,work,mount}

sudo mount -t overlay -o rw,relatime,lowerdir=/,upperdir=/tmp/overlayfs-test/upper/,workdir=/tmp/overlayfs-test/work/ none /tmp/overlayfs-test/mount/

mkdir -p /tmp/overlayfs-test/mount/some/dir/
touch /tmp/overlayfs-test/mount/some/dir/some_file
```


```console
sh-5.0$ df -h /tmp/overlayfs-test/mount/some/dir/some_file
Filesystem      Size  Used Avail Use% Mounted on
tmpfs           3.7G   16K  3.7G   1% /tmp

sh-5.0$ df -h /tmp/overlayfs-test/mount/some/dir/
Filesystem      Size  Used Avail Use% Mounted on
none            3.7G   16K  3.7G   1% /tmp/overlayfs-test/mount
```

I can reproduce it on `4.15.18-24-pve` and `5.5.2-arch1-1` kernels

jsafrane: On Fedora / CentOS / Debian, both `/tmp/overlayfs-test/mount/some/dir/some_file` and `/tmp/overlayfs-test/mount/some/dir/` have the same device and mount detection works well.

It looks like a bug / overlayfs weirdness that a created *file* and *directory* in /tmp/overlayfs-test/mount  have different devices.

Is related to Arch? Not being kernel not overlayfs expert, can it be because of these extra options? https://github.com/archlinux/svntogit-packages/blob/packages/linux/trunk/config#L9766-L9769

kvaps: Hi sorry for long answer, try this:
```bash
mkdir /tmp/1
mount -t tmpfs tmpfs /tmp/1
cd /tmp/1
mkdir 1 2 3 4
fallocate -l 1G some.img
mkfs.ext4 some.img
mount -o loop some.img 1
mount -t overlay -o lowerdir=/tmp/1/1,upperdir=/tmp/1/2,workdir=/tmp/1/3 blabla 4
touch 4/aaa
df -h 4/ 4/aaa
```

example output:
```
Filesystem      Size  Used Avail Use% Mounted on
blabla           16G  1.1M   16G   1% /tmp/1/4
tmpfs            16G  1.1M   16G   1% /tmp/1
```

I can reproduce it on `5.4.0-37-generic` (Ubuntu 20.04) and `4.19.0-9-amd64` (Debian 10)
I can't reproduce on `5.7.8-arch1-1` and `5.7.9-arch1-1` (Arch Linux)

and to cleanup
```bash
cd -
umount -R /tmp/1
rm -rf /tmp/1
```

jingxu97:
@jsafrane just want to check this pr still need to be reviewed?

hzliangbin:
Any progress? @kvaps @jingxu97

PR #162, opened 2020-04-28
mount: force unique scope names

vespian:
Some of the older distros (Debian 8.0, Centos 7.8, Ubuntu 14.04 LTS) use systemd,
which uses PIDs for scope names:

https://lists.freedesktop.org/archives/systemd-devel/2015-October/034591.html

This was fixed in the systemd >=228, commit systemd/systemd@9c8d1e1, but
for the time being, we need to support these older distros.

The idea is to force scope names to use UUID. The chance for the
collision is small enough to be negligible (128 bits) so that checking for pre-existsing scope names is not necessary.

Related GitHub issue: https://github.com/kubernetes/kubernetes/issues/90327

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.12
 
 require (
 	github.com/davecgh/go-spew v1.1.1
+	github.com/google/uuid v1.1.1
 	github.com/spf13/afero v1.2.2
 	github.com/stretchr/testify v1.3.0
 	k8s.io/klog/v2 v2.0.0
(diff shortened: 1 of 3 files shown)


vespian:
> It may take a couple minutes for the CLA signature to be fully registered; after that, please reply here with a new comment and we'll verify. Thanks.

CLA signed via CCLA.

vespian:
@apelisse @saad-ali Friendly ping

lavalamp:
I think you need someone from:
/sig node

/unassign

vespian:
@apelisse @saad-ali Ping?

apelisse:
I don't know anything about this.
/unassign

PR #164, opened 2020-05-12
use lazy umount to avoid pod terminating problems caused by volume mounting dir accessed by other pods

borgerli:
fix issue #163


borgerli:
I signed it

borgerli:
I signed it

borgerli:
I signed it

aviau:
I am currently investigating possibly-related issues where I have an unmount command issued by kubelet that has been hanging for days.

I have a volume that mounts unreliably and I was wondering if it could be because the mount is due to be executed after the hanging unmount so it never ends up running?

PR #165, opened 2020-05-16
Import BoundedFrequencyRunner from k8s.io/kubernetes

danwinship:
kube-proxy uses `BoundedFrequencyRunner` to batch up changes to iptables for performance reasons. This logic could be used in other places. It is already used in openshift-sdn to batch up NetworkPolicy-related OVS flow updates in the same way (openshift-sdn vendors k/k because it incorporates kube-proxy into itself) and recently we wanted to use it in ovn-kubernetes for something similar, but ovn-kubernetes doesn't vendor k/k and really shouldn't.

Also, I don't know the current staging master plan, but if kube-proxy is ever to move out of k/k then its dependencies would also have to move, right?

This preserves the package name `async` from `k8s.io/kubernetes/pkg/util/async`, but doesn't include [`async.Runner`](https://github.com/kubernetes/kubernetes/blob/master/pkg/util/async/runner.go) from there as well (which appears to currently only be used by `k8s.io/kubernetes/pkg/master/controller.go`) Maybe `async` isn't the best name for it here.

This also has the side effect of pulling in `k8s.io/client-go/util/flowcontrol` in as a dependency, to get `flowcontrol.RateLimiter`. client-go also pulls in some packages from utils, so probably this is bad, dependency-chain-wise? (Aside: if so maybe `make verify` should verify this?) One possibility would be to move `k8s.io/client-go/util/flowcontrol` here as well. It kind of goes well with `BoundedFrequencyRunner` and they could form a package together?

/assign @thockin 
/sig network

### DIFF
--- a/flowcontrol/bounded_frequency_runner.go
+++ b/flowcontrol/bounded_frequency_runner.go
@@ -21,8 +21,6 @@ import (
 	"sync"
 	"time"
 
-	"k8s.io/client-go/util/flowcontrol"
-
 	"k8s.io/klog/v2"
 )
 
@@ -181,7 +179,7 @@ func construct(name string, fn func(), minInterval, maxInterval time.Duration, b
 	} else {
 		// allow burst updates in short succession
 		qps := float32(time.Second) / float32(minInterval)
-		bfr.limiter = flowcontrol.NewTokenBucketRateLimiterWithClock(qps, burstRuns, timer)
+		bfr.limiter = NewTokenBucketRateLimiterWithClock(qps, burstRuns, timer)
 	}
 	return bfr
 }
@@ -213,7 +211,7 @@ func (bfr *BoundedFrequencyRunner) Loop(stop <-chan struct{}) {
 // may be dropped - it is just guaranteed that we will try calling the
 // underlying function as soon as possible starting from now.
 func (bfr *BoundedFrequencyRunner) Run() {
-	// If it takes a lot of time to run the underlying function, noone is really
+	// If it takes a lot of time to run the underlying function, no one is really
 	// processing elements from <run> channel. So to avoid blocking here on the
 	// putting element to it, we simply skip it if there is already an element
 	// in it.
--- a/flowcontrol/bounded_frequency_runner.go
+++ b/flowcontrol/bounded_frequency_runner.go
@@ -1,0 +1,313 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package flowcontrol
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"k8s.io/client-go/util/flowcontrol"
+
+	"k8s.io/klog/v2"
+)
+
+// BoundedFrequencyRunner manages runs of a user-provided function.
+// See NewBoundedFrequencyRunner for examples.
+type BoundedFrequencyRunner struct {
+	name        string        // the name of this instance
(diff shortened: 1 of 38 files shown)


async/bounded_frequency_runner.go
```diff
@@ -0,0 +1,313 @@
...
+*/
+
+package async
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"k8s.io/client-go/util/flowcontrol"
```
liggitt, line 24: Yeah, we can't take this dependency. With it, k8s.io/utils wouldn't be able to be used by k8s.io/kubernetes

liggitt: And yes, we need a verify job checking https://github.com/kubernetes/utils#criteria-for-adding-code-here

> No dependencies on any other Kubernetes repository

And that `go mod tidy` has been run

danwinship: > > No dependencies on any other Kubernetes repository

lol I'm so used to tools and bots that automatically enforce all the rules I didn't think to check the README

But also, that should be "any other Kubernetes repository except `k8s.io/klog/v2`"

danwinship: Not that it helps with the `client-go` dependency directly, but would this make more sense in `apimachinery` than `utils`? The apimachinery docs claim that that module is for "Scheme, typing, encoding, decoding, and conversion packages for Kubernetes and Kubernetes-like API objects" but... that's not really what currently there.

liggitt: > that should be "any other Kubernetes repository except k8s.io/klog/v2"

I'd actually like to see us drop that dep as well in favor of a logging interface eventually

thockin: +1 to logging interfaces

Could we move client-go/util/flowcontrol up here or duplicate the subset of it that we need?

danwinship:
updated: now moves `k8s.io/client-go/util/flowcontrol/throttle.go` and `k8s.io/kubernetes/pkg/util/async/bounded_frequency_runner.go` into `k8s.io/utils/flowcontrol/`. I used the directions in `HOWTOMOVE.md` to preserve git history (more or less).

This still leaves `Backoff` in client-go/util/flowcontrol and `Runner` in kubernetes/pkg/util/async

/priority backlog

flowcontrol/bounded_frequency_runner.go
```diff
@@ -0,0 +1,311 @@
...
+	lastRun time.Time   // time of last run
+	timer   timer       // timer for deferred runs
+	limiter rateLimiter // rate limiter for on-demand runs
+
+	retry     chan struct{} // schedule a retry
+	retryMu   sync.Mutex    // guards retryTime
+	retryTime time.Time     // when to retry
+}
+
+// designed so that RateLimiter satisfies
```
thockin, line 47: nit if you make other changes: s/RateLimiter/flowcontrol.Ratelimiter

flowcontrol/bounded_frequency_runner.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+}
+
+// Run the function as soon as possible.  If this is called while Loop is not
+// running, the call may be deferred indefinitely.
+// If there is already a queued request to call the underlying function, it
+// may be dropped - it is just guaranteed that we will try calling the
+// underlying function as soon as possible starting from now.
+func (bfr *BoundedFrequencyRunner) Run() {
+	// If it takes a lot of time to run the underlying function, noone is really
```
thockin, line 214: nit if you make other changes: s/noone/no one/ (or nobody)

danwinship:
@thockin: updated for your previous nitpicks, but also, this is claiming "`cncf-cla: no`", presumably because I followed the directions to import the git history and so have commits from now-un-CLA'ed people? So this would either need to be force-merged or I'd need to squash the commits

danwinship:
> @thockin added `cncf-cla: yes` and removed `cncf-cla:no` labels 3 days ago

but this doesn't appear to have cleared the `cla/linuxfoundation — danwinship authorized, but 1 other problem` condition.

thockin:
I can't tell if f5655637a283a915d38dead20d5f4696c847a0de actually holds any deltas - that seems to be the one tripping it up

thockin:
When code freeze lifts, we can manually merge, I guess...

rikatz:
@danwinship @thockin so, our conclusion for this one is: should we re-open this, move to k8s.io/async or leave this alone? (maybe move to components-helpers?)

PR #166, opened 2020-05-18
Remove driver letter assignment during volume format

jingxu97:
This PR removes the driver letter assignment during volume format and
mount because driver letter might run out and cause issues during mount.
Intead, it uses volume id to mount the target dir.

### DIFF
--- a/mount/mount_windows.go
+++ b/mount/mount_windows.go
@@ -26,7 +26,6 @@ import (
 	"strings"
 
 	"k8s.io/klog/v2"
-	utilexec "k8s.io/utils/exec"
 	"k8s.io/utils/keymutex"
 	utilpath "k8s.io/utils/path"
 )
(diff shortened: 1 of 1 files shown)


jingxu97:
cc @andyzhangx

jingxu97:
@andyzhangx , could you add /lgtm? The label is not on yet. Thanks!

mount/mount_windows.go
```diff
@@ -230,17 +229,17 @@ func (mounter *SafeFormatAndMount) formatAndMountSensitive(source string, target
 
 	// format disk if it is unformatted(raw)
 	cmd := fmt.Sprintf("Get-Disk -Number %s | Where partitionstyle -eq 'raw' | Initialize-Disk -PartitionStyle MBR -PassThru"+
-		" | New-Partition -AssignDriveLetter -UseMaximumSize | Format-Volume -FileSystem %s -Confirm:$false", source, fstype)
```
jingxu97, line 233: driver letter is not assigned during disk initialization. However, this will cause issues during evalSymlink  https://github.com/kubernetes/utils/issues/172

PR #177, opened 2020-07-20
trace: on klog.V(4) only log all nested traces inside trace above threshold

sttts:
Addresses https://github.com/kubernetes/utils/pull/139/files#r456819382


dims:
thanks!

ugh. needs `gofmt`...

Issue #172, opened 2020-06-15
Remove drive letter during formatAndMount causes subPath evalSymlink issue


jingxu97:
In PR https://github.com/kubernetes/utils/pull/166, we remove driver letter during formatAndMount due to the issue of driver letter runout. Now, volume mount will create symlink from volumeId to a path. e.g., 
mklink testlink  \\?\Volume{2f230ace-0000-0000-0000-010000000000}\

However, in subPath setup, https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/util/subpath/subpath_windows.go#L53
filepath.EvalSymlinks(hostPath) 
will fail with error: too many links. 

We need to find another way to evaluate the symlink when driver letter is not assigned.

andyzhangx:
actually I the author of that code `finalSubPath, err := filepath.EvalSymlinks(hostPath)`, can we skip this `too many links` directly in the code if `hostPath` starts with `\`, this should be bug in `filepath.EvalSymlinks`

jingxu97:
@andyzhangx you mean skip this EvalSymlinks  check or if the error is "too many links", we ignore the error?

jingxu97:
@andyzhangx but that means we don't check symlink for subpath in this case. Then it will open up the security issue that was fixed by this symlink check, I think.

jingxu97:
> Is that windows specific code, or will that open a hole in Linux symlink checking?

This is mount_windows.go code. So the changes here will not affect linux symlink checking.

networkimprov:
Re Go filepath.EvalSymlinks, see https://github.com/golang/go/issues/40180

It is probably not a good solution to any problem, except to show an administrator where a certain path leads *at this moment*.

jingxu97:
/remove-lifecycle stale

We made a workaround to solve this, but still want to investigate more to find a better way..

PR #179, opened 2020-07-23
Remove driver letter assignment during volume format

jingxu97:
This PR removes the driver letter assignment during volume format and
mount because driver letter might run out and cause issues during mount.
Intead, it uses volume id to mount the target dir.

The original PR was reverted due to issue https://github.com/kubernetes/utils/issues/172
After PR https://github.com/kubernetes/kubernetes/pull/92328 and https://github.com/kubernetes/kubernetes/pull/93148 are in, we will try to push it again.

### DIFF
--- a/mount/mount_windows.go
+++ b/mount/mount_windows.go
@@ -26,7 +26,6 @@ import (
 	"strings"
 
 	"k8s.io/klog/v2"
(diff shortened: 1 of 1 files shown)


jingxu97:
cc @ddebroy

PR #180, opened 2020-07-23
Add nfs mount support for windows in mount utils

jingxu97:
This PR adds mount nfs support in mount_windows.go

### DIFF
--- a/mount/mount_windows.go
+++ b/mount/mount_windows.go
@@ -55,6 +55,45 @@ func (mounter *Mounter) Mount(source string, target string, fstype string, optio
 	return mounter.MountSensitive(source, target, fstype, options, nil /* sensitiveOptions */)
 }
 
+//mountNfs uses powershell command "net use" to connect NFS share
+func mountNfs(source, target string, options []string) error {
+	mountCmd := "net use " + source
+	output, err := exec.Command(mountCmd, options...).CombinedOutput()
+	if err != nil {
+		return err
+	}
+
+	klog.V(4).Infof("NFS mount source(%q) on target(%q) successfully, output: %q", source, target, string(output))
+	return nil
+}
+
+//mountSmb tries to create SMB mapping to a SMB share
+func mountSmb(source, target string, options []string) error {
+	if len(options) < 2 {
+		klog.Warningf("mount options(%q) command number(%d) less than 2, source:%q, target:%q, skip mounting",
+			sanitizedOptionsForLogging, len(options), source, target)
+		return nil
+	}
+	// lock smb mount for the same source
+	getSMBMountMutex.LockKey(source)
+	defer getSMBMountMutex.UnlockKey(source)
+
+	if output, err := newSMBMapping(options[0], options[1], source); err != nil {
+		if isSMBMappingExist(source) {
+			klog.V(2).Infof("SMB Mapping(%s) already exists, now begin to remove and remount", source)
+			if output, err := removeSMBMapping(source); err != nil {
+				return fmt.Errorf("Remove-SmbGlobalMapping failed: %v, output: %q", err, output)
+			}
+			if output, err := newSMBMapping(options[0], options[1], source); err != nil {
+				return fmt.Errorf("New-SmbGlobalMapping remount failed: %v, output: %q", err, output)
+			}
+		} else {
+			return fmt.Errorf("New-SmbGlobalMapping failed: %v, output: %q", err, output)
+		}
+	}
+	return nil
+}
+
 // MountSensitive is the same as Mount() but this method allows
 // sensitiveOptions to be passed in a separate parameter from the normal
 // mount options and ensures the sensitiveOptions are never logged. This
@@ -84,43 +123,22 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
 		allOptions := []string{}
 		allOptions = append(allOptions, options...)
 		allOptions = append(allOptions, sensitiveOptions...)
-		if len(allOptions) < 2 {
-			klog.Warningf("mount options(%q) command number(%d) less than 2, source:%q, target:%q, skip mounting",
-				sanitizedOptionsForLogging, len(allOptions), source, target)
-			return nil
-		}
-
-		// currently only cifs mount is supported
-		if strings.ToLower(fstype) != "cifs" {
-			return fmt.Errorf("only cifs mount is supported now, fstype: %q, mounting source (%q), target (%q), with options (%q)", fstype, source, target, sanitizedOptionsForLogging)
-		}
-
-		// lock smb mount for the same source
-		getSMBMountMutex.LockKey(source)
-		defer getSMBMountMutex.UnlockKey(source)
-
-		if output, err := newSMBMapping(allOptions[0], allOptions[1], source); err != nil {
-			if isSMBMappingExist(source) {
-				klog.V(2).Infof("SMB Mapping(%s) already exists, now begin to remove and remount", source)
-				if output, err := removeSMBMapping(source); err != nil {
-					return fmt.Errorf("Remove-SmbGlobalMapping failed: %v, output: %q", err, output)
-				}
-				if output, err := newSMBMapping(allOptions[0], allOptions[1], source); err != nil {
-					return fmt.Errorf("New-SmbGlobalMapping remount failed: %v, output: %q", err, output)
-				}
-			} else {
-				return fmt.Errorf("New-SmbGlobalMapping failed: %v, output: %q", err, output)
-			}
+		fstype = strings.ToLower(fstype)
+		switch fstype {
+		case "nfs":
+			mountNfs(source, target, allOptions)
+		case "cifs":
+			mountSmb(source, target, allOptions)
+		default:
+			return fmt.Errorf("mount failed: only fstypes nfs or cifs are supported")
 		}
 	}
-
 	output, err := exec.Command("cmd", "/c", "mklink", "/D", target, bindSource).CombinedOutput()
 	if err != nil {
 		klog.Errorf("mklink failed: %v, source(%q) target(%q) output: %q", err, bindSource, target, string(output))
 		return err
 	}
 	klog.V(2).Infof("mklink source(%q) on target(%q) successfully, output: %q", bindSource, target, string(output))
-
 	return nil
 }
 
@@ -164,7 +182,7 @@ func removeSMBMapping(remotepath string) (string, error) {
 
 // Unmount unmounts the target.
 func (mounter *Mounter) Unmount(target string) error {
-	klog.V(4).Infof("azureMount: Unmount target (%q)", target)
+	klog.V(4).Infof("Unmount target (%q)", target)
 	target = NormalizeWindowsPath(target)
 	if output, err := exec.Command("cmd", "/c", "rmdir", target).CombinedOutput(); err != nil {
 		klog.Errorf("rmdir failed: %v, output: %q", err, string(output))


jingxu97:
cc @andyzhangx

mount/mount_windows.go
```diff
@@ -55,6 +55,46 @@ func (mounter *Mounter) Mount(source string, target string, fstype string, optio
 	return mounter.MountSensitive(source, target, fstype, options, nil /* sensitiveOptions */)
 }
 
+//mountNfs uses powershell command "net use" to connect NFS share
+func mountNfs(source, target string, options []string) error {
+	mountCmd := "powershell /c net use " + source
```
andyzhangx, line 60: do you know `powershell /c net use ` works for container volume mapping?

jingxu97: you mean whether it can start container with this? I tried manually and it works for me.

jingxu97: I am also wondering whether it can work for smb too.

andyzhangx: `net use` won't work for smb, that's the reason why we use `SMBGlobalMapping`

jingxu97:
@andyzhangx do you have other comments about PR?

jingxu97:
ping @jsafrane @msau42 PTAL

mount/mount_windows.go
```diff
@@ -166,6 +185,18 @@ func removeSMBMapping(remotepath string) (string, error) {
 func (mounter *Mounter) Unmount(target string) error {
 	klog.V(4).Infof("azureMount: Unmount target (%q)", target)
 	target = NormalizeWindowsPath(target)
+	source, err := os.Readlink(target)
+	if err != nil {
+		klog.Infof("failed to get source from target %q, %v", target, err)
+		return err
+	}
+	output, err := exec.Command("powershell", "/c", "net use", source, "/delete").CombinedOutput()
```
KnicKnic, line 193: you should not open powershell just to use net

mount/mount_windows.go
KnicKnic, line 60: you should not use powershell to launch net

mount/mount_windows.go
jingxu97: It seems working for what I tested. So the better way to execute just execute it directly? exec.Command("net use", source, "/delete")

mount/mount_windows.go
```diff
@@ -55,6 +55,46 @@ func (mounter *Mounter) Mount(source string, target string, fstype string, optio
 	return mounter.MountSensitive(source, target, fstype, options, nil /* sensitiveOptions */)
 }
 
+//mountNfs uses powershell command "net use" to connect NFS share
+func mountNfs(source, target string, options []string) error {
```
msau42, line 59: make sure options are santized before logging

mount/mount_windows.go
msau42, line 193: is this function used for unmount volume types other than nfs?

mount/mount_windows.go
```diff
@@ -166,6 +185,18 @@ func removeSMBMapping(remotepath string) (string, error) {
 func (mounter *Mounter) Unmount(target string) error {
 	klog.V(4).Infof("azureMount: Unmount target (%q)", target)
```
msau42, line 186: This can be done separately, but can we clean up these "azure" strings from the logs, since this is supposed to be a provider-independent library?

mount/mount_windows.go
```diff
@@ -55,6 +55,46 @@ func (mounter *Mounter) Mount(source string, target string, fstype string, optio
 	return mounter.MountSensitive(source, target, fstype, options, nil /* sensitiveOptions */)
 }
 
+//mountNfs uses powershell command "net use" to connect NFS share
+func mountNfs(source, target string, options []string) error {
+	mountCmd := "powershell /c net use " + source
+	output, err := exec.Command(mountCmd, options...).CombinedOutput()
+	if err != nil {
+		klog.Errorf("NFS mount failed: %v, source(%q) output: %q", err, source, target, string(output))
```
msau42, line 63: Could there be any sensitive output like credentials in the error message?

mount/mount_windows.go
```diff
@@ -55,6 +55,46 @@ func (mounter *Mounter) Mount(source string, target string, fstype string, optio
...
+	klog.V(4).Infof("NFS mount source(%q) on target(%q) successfully, output: %q", source, target, string(output))
+	return nil
+}
+
+//mountSmb tries to create SMB mapping to a SMB share
+func mountSmb(source, target string, options []string) error {
+	if len(options) < 2 {
+		klog.Warningf("mount options(%q) command number(%d) less than 2, source:%q, target:%q, skip mounting",
+			sanitizedOptionsForLogging, len(options), source, target)
+		return nil
```
msau42, line 76: This isn't your change, but is skipping mounting the right thing to do?

mount/mount_windows.go
```diff
@@ -166,6 +185,18 @@ func removeSMBMapping(remotepath string) (string, error) {
...
 	klog.V(4).Infof("azureMount: Unmount target (%q)", target)
 	target = NormalizeWindowsPath(target)
+	source, err := os.Readlink(target)
+	if err != nil {
+		klog.Infof("failed to get source from target %q, %v", target, err)
+		return err
+	}
+	output, err := exec.Command("powershell", "/c", "net use", source, "/delete").CombinedOutput()
+	if err != nil {
+		klog.Infof("failed to delete connection from source %q %v", source, err)
```
msau42, line 195: Can we return this message as the error and let the caller log?

mount/mount_windows.go
```diff
@@ -166,6 +185,18 @@ func removeSMBMapping(remotepath string) (string, error) {
...
+		return err
+	}
+	klog.V(2).Infof("delete connection from source %q, output: ", source, output)
```
msau42, line 198: what does the output show?

better wording would be "connection deleted from source" to indicate that it's already happened.

mount/mount_windows.go
jingxu97: I feel it should fail the operation. @andyzhangx is there some reason for skipping here?

mount/mount_windows.go
jingxu97: I don't think I logged options here.

mount/mount_windows.go
jingxu97: I tried out this command, and password information is not in error message.

mount/mount_windows.go
jingxu97: currently only nfs and smb are supported here. For smb, it is not needed, but since we don't have fstype information during unmount, we cannot make different choices here. I tried net use source /delete, it does not return error if source does not exist. 

So unlike Linux, the mount/unmount functions used for different file systems in Windows are different.  We need to consider add fstype information in Unmount APIs, I think,

mount/mount_windows.go
andyzhangx: I think we can return error here if `len(options) < 2 `

mount/mount_windows.go
jingxu97: removed it.

mount/mount_windows.go
jingxu97: removed

mount/mount_windows.go
jingxu97: actually, I checked this part again, and decided to remove this /delete code. Both smb and nfs have the same issue here. During mount, there are two steps
1. Set up network connection to remote share

- smb: this uses New-SmbGlobalMapping to set up the connection to the smb share

- nfs: use "net use" to set up the connection.

2. create a symlink from the remote connection to local path

During unmount, we will remove the symlink, but we could not delete the remote connection. If there are multiple pods on the same node share the same remote connection, deleting the remote connection will affect other pods. Or even outside of kubernetes, user might directly connect remote share to local host.

So currently, for smb, we do not remove the mapping during unmount, similarly, the option for nfs is to not delete the connection.

jingxu97:
@msau42 @KnicKnic I updated the PR and added some comments. PTAL

mount/mount_windows.go
jingxu97: removed

mount/mount_windows.go
msau42: When does it get cleaned up then?

msau42: What happens if you mount the same share multiple times on the same node?

mount/mount_windows.go
msau42: Can we return this as the error message and leave logging to the caller?

mount/mount_windows.go
jingxu97: It will not get cleaned up (delete connection)  because we cannot know when it is safe to do so. In linux, the mount is 1:1 mapping from remote to individual local path. So we could safely do unmount for each "mount". But in windows, remote connection is first connected to local host as a global connection, and each mount will create symlink from this global connection to individual path. 
If mount same share multiple times on the same node, it should work as expected. Do you see any issues with it?

msau42: What happens if you create a connection and one already exists?

I'm concerned about resource leakage and maybe over time after many mounts, we will run out of resources on the node and will be unable to mount anymore. Or we will reach the max number of connections to an nfs/smb server due to all the stale client connections that still remain.  It seems like maybe we need to extend the nfs + smb plugins to support MountDevice so that it can clean up these resources.

jingxu97: I could not find whether there is a limit on it. Will check with sig-windows. @andyzhangx 
I think we could extend plugin to support MountDevice, only concern here is out of band connection, and we might accidentally disconnect user's remote share.

mount/mount_windows.go
jingxu97: removed

mount/mount_windows.go
jingxu97: will create another PR for it.

mount/mount_windows.go
andyzhangx: the main diff on smb mount between linux and windows is that on LInux, every smb mount is independent, while on Windows, one remote smb server address(e.g. `smb-server/fileshare`) can only have one SMB-GlobalMapping. So `if you create a connection and one already exists` on Windows, that is a difficulty:
 - in some cases, we should skip create SMB-GlobalMapping since it already exists
 - in other cases, original password is incorrect or expired, user wants to delete SMB-GlobalMapping, and then create a new SMB-GlobalMapping

On Windows, there is no way to check whether original SMB-GlobalMapping is valid or not, that was the answer about 1 years ago from windows team, will update this thread(also about the max SMB-GlobalMapping limit) if there is any new update recently.

msau42: What is the behavior that we do now in the plugin? Is it possible to have the plugin do reference counting so that we can clean up the global mapping when all mounts are gone? We have the MountDevice interface which is supposed to automatically do the reference counting, but I'm not sure if it handles subshares correctly, or the same share referenced in multiple PVs.

jingxu97: right now for smb global mapping, during setup, it will remove existing global mapping first due to the reason @andyzhangx mentioned above. There might cause issues for existing pod which uses the global mapping, but we haven't got a better way to handle it yet. https://github.com/kubernetes/utils/blob/f32512634ab7ee0599334344e026b1b47d39e4f4/mount/mount_windows.go#L103
During teardown, we do not do anything about the global mapping (no cleanup) except removing the symlink of local path from the mapping.

For nfs here, I didn't delete existing nfs connection during set up. We could execute "net use /nfsmountpoint" multiple times. We could do similar thing as smb, but both has issues. We could do some further discussion about it, and document the implication.

msau42: For the smb case, does that mean we can't support multiple pods using the same share?

For the nfs case, what happens if you run "net use" multiple times on the same mount? Is there a chance we will exhaust resources on the node if we don't clean them up?

jingxu97: for smb, when new smb mapping is created, I am not sure whether data access can be recovered or not. @andyzhangx do you have tests  for it with azure? I can also try csi smb driver later.

If run "net use" multiple times on the same mount, it will succeed. I don't see any additional mount is created, it will be the same one.

andyzhangx: @msau42 @jingxu97 that's a great discussion, I filed an issue here: https://github.com/kubernetes/kubernetes/issues/93805
will spend some time fixing the smb mapping issue.

andyzhangx: and about the SMBGlobalMapping limit, here is the answer from windows expert, so currently it's ok.
```
They're isn't any hard limit. But each global mapping will consume resources in the rdr. (Connections, authenticated sessions, file objects etc).  I'm assuming you are using a driveless global mapping. 
```

andyzhangx: there is a PR to fix the new smb mapping issue: https://github.com/kubernetes/utils/pull/184

andyzhangx:
does `net use` really works for NFS server? there is a discussion here: https://github.com/kubernetes/kubernetes/issues/56188#issuecomment-397223626

And also I think we should enable NFS client on Windows first according to https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2008-R2-and-2008/cc754350(v=ws.10), while that only supports NFSv3

andyzhangx:
here is the way to connect to NFS server on Windows:

### Instructions for using NFS on Windows Server
```
C:\Users\azureuser>powershell
Windows PowerShell
Copyright (C) Microsoft Corporation. All rights reserved.

PS C:\Users\azureuser> Install-WindowsFeature NFS-Client
PS C:\Users\azureuser> exit

C:\Users\azureuser>mount \\{NFS-share} G:
G: is now successfully connected to \\{NFS-share}
The command completed successfully.
```

mount/mount_windows.go
jingxu97: discussed offline, it can work after creating symlink of local dir to remote network connection.

mount/mount_windows.go
```diff
@@ -55,6 +55,45 @@ func (mounter *Mounter) Mount(source string, target string, fstype string, optio
 	return mounter.MountSensitive(source, target, fstype, options, nil /* sensitiveOptions */)
 }
 
+//mountNfs uses powershell command "net use" to connect NFS share
+func mountNfs(source, target string, options []string) error {
+	mountCmd := "net use " + source
```
andyzhangx, line 60: `net use /persistent:yes`

https://www.windowscommandline.com/net-use/
```
/persistent: {yes | no}: Controls the use of persistent network connections. The default is the setting used last. Deviceless connections are not persistent. Yes saves all connections as they are made, and restores them at next logon. No does not save the connection being made or subsequent connections. Existing connections are restored at the next logon. Use /delete to remove persistent connections.
```

andyzhangx:
> does `net use` really works for connecting to NFS server on Windows? there is a discussion here: [kubernetes/kubernetes#56188 (comment)](https://github.com/kubernetes/kubernetes/issues/56188#issuecomment-397223626)
> 
> And also I think we should enable NFS client on Windows first according to https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2008-R2-and-2008/cc754350(v=ws.10), while that only supports NFSv3

I was wrong, use `net use` works
```
Microsoft Windows [Version 10.0.17763.1339]
(c) 2018 Microsoft Corporation. All rights reserved.
azureuser@4820k8s000 C:\Users\azureuser>powershell
Windows PowerShell
Copyright (C) Microsoft Corporation. All rights reserved.
PS C:\Users\azureuser> powershell
Windows PowerShell
Copyright (C) Microsoft Corporation. All rights reserved.
PS C:\Users\azureuser> Install-WindowsFeature NFS-Client
Success Restart Needed Exit Code      Feature Result
------- -------------- ---------      --------------
True    No             Success        {Client for NFS}
PS C:\Users\azureuser> exit
azureuser@4820k8s000 C:\Users\azureuser>net use /PERSISTENT:YES \\account.blob.core.windows.net\andyblobnfs\pvc-80e0b096-a126-4a05-a100-b957c2335e2e
The command completed successfully.

azureuser@4820k8s000 C:\Users\azureuser>mklink /D blobnfs \\account.blob.core.windows.net\andyblobnfs\pvc-80e0b096-a126-4a05-a100-b957c2335e2e
symbolic link created for blobnfs <<===>> \\account.blob.core.windows.net\andyblobnfs\pvc-80e0b096-a126-4a05-a100-b957c2335e2e

docker run -d -v C:\Users\azureuser\blobnfs:C:\mnt --name busybox e2eteam/busybox:1.29
```

andyzhangx:
can we continue with this PR?

jingxu97:
yes, I will start to work on it again soon. Is there any concern with the current code?

andyzhangx:
> yes, I will start to work on it again soon. Is there any concern with the current code?

@jingxu97 that's great, I am just following up with the progress since I also have the requirement on this feature, thanks.

PR #182, opened 2020-07-29
Fix tracing reporting

wojtek-t:
Ref https://github.com/kubernetes/kubernetes/issues/93531

/assign @dims @jpbetz

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -197,7 +197,7 @@ func (t *Trace) logTrace() {
 		}
 
 		// if any step took more than it's share of the total allowed time, it deserves a higher log level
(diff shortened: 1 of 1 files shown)


trace/trace.go
```diff
@@ -197,7 +197,7 @@ func (t *Trace) logTrace() {
 		}
 
 		// if any step took more than it's share of the total allowed time, it deserves a higher log level
-		buffer.WriteString(fmt.Sprintf("(%v) (total time: %vms):", t.startTime.Format("02-Jan-2006 15:04:00.000"), totalTime.Milliseconds()))
+		buffer.WriteString(fmt.Sprintf("(%v) (total time: %vms):", t.startTime.Format("02-Jan-2006 15:04:05.000"), totalTime.Milliseconds()))
```
liggitt, line 200: would be good to get a test verifying this

PR #183, opened 2020-07-31
Add PtrDerefOr for more types

zianke:
The `pointer` package currently has `Int32PtrDerefOr`. This PR has added `XPtrDerefOr` for the other types (int64, bool, string, float32, float64). These functions can be useful in dereferencing pointer fields in Kubernetes objects with default values.

`pointer_test.go` is updated to cover all functions in `pointer.go`.

### DIFF
--- a/pointer/pointer_test.go
+++ b/pointer/pointer_test.go
@@ -18,7 +18,6 @@ package pointer
 
 import (
 	"fmt"
-	"reflect"
 	"testing"
 )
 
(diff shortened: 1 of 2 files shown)


zianke:
/assign @thockin
Thanks.

pointer/pointer_test.go
```diff
@@ -63,9 +62,230 @@ func TestAllPtrFieldsNil(t *testing.T) {
...
-			if !reflect.DeepEqual(tc.expected, actualErr) {
-				t.Errorf("%s: expected %t, got %t", name, tc.expected, !tc.expected)
+			if actual := AllPtrFieldsNil(tc.obj); actual != tc.expected {
+				t.Errorf("%s: expected %t, got %t", name, tc.expected, actual)
+			}
+		})
+	}
+}
+
+func TestInt32Ptr(t *testing.T) {
```
apelisse, line 72: I feel like these tests are harder to read because of the table-driven testing, while there are only two tests in each. Would it be better to just unwind the loop?

zianke: Thanks for reviewing. Yes that makes sense. I've updated the tests.

apelisse: That's a lot of temporary variables, but that's fine :-)

/lgtm
/approve

PR #184, opened 2020-08-09
fix: smb remount issue

andyzhangx:
This PR could fix https://github.com/kubernetes/kubernetes/issues/93805

https://github.com/kubernetes/kubernetes/pull/73661 not fixed the SMB mount issue totally, when there is a SMB mount with same remote path, original SMB mapping would be removed and SMB mapping re-established, that would cause temporally data inaccessible with the original container. 

#### fix
should use `test-path \\server\share` to check whether original smb mapping is valid, here is an example showing that after SMB server password changed, `Test-Path` could validate whether connection is ok:
```
Microsoft Windows [Version 10.0.17763.1339]
(c) 2018 Microsoft Corporation. All rights reserved.

azureuser@3460k8s000 C:\Users\azureuser>powershell
Windows PowerShell
Copyright (C) Microsoft Corporation. All rights reserved.

PS C:\Users\azureuser> Get-SmbGlobalMapping

Status Local Path Remote Path
------ ---------- -----------
OK                \\accountname.file.core.windows.net\andy-win1186-dynamic-pvc-b23d5a60-50dd-4c0c-bdc1-f4ff40349609


PS C:\Users\azureuser> Test-Path \\accountname.file.core.windows.net\andy-win1186-dynamic-pvc-b23d5a60-50dd-4c0c-bdc1-f4ff40349609
True

# After SMB server password changed

PS C:\Users\azureuser> Test-Path \\accountname.file.core.windows.net\andy-win1186-dynamic-pvc-b23d5a60-50dd-4c0c-bdc1-f4ff40349609
False

# Get-SmbGlobalMapping status is always OK
PS C:\Users\azureuser> Get-SmbGlobalMapping

Status Local Path Remote Path
------ ---------- -----------
OK                \\fca1c6d57af91439aa77889.file.core.windows.net\andy-win1186-dynamic-pvc-b23d5a60-50dd-4c0c-bdc1-f4ff40349609
```

/assign @msau42 @jingxu97

### DIFF
--- a/mount/mount_windows.go
+++ b/mount/mount_windows.go
@@ -98,18 +98,31 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
 		getSMBMountMutex.LockKey(source)
 		defer getSMBMountMutex.UnlockKey(source)
 
-		if output, err := newSMBMapping(allOptions[0], allOptions[1], source); err != nil {
+		var output string
+		var err error
+		username := allOptions[0]
+		password := allOptions[1]
+		if output, err = newSMBMapping(username, password, source); err != nil {
+			klog.Warningf("SMB Mapping(%s) returned with error(%v), output(%s)", source, err, string(output))
 			if isSMBMappingExist(source) {
-				klog.V(2).Infof("SMB Mapping(%s) already exists, now begin to remove and remount", source)
-				if output, err := removeSMBMapping(source); err != nil {
-					return fmt.Errorf("Remove-SmbGlobalMapping failed: %v, output: %q", err, output)
+				valid, errPath := isValidPath(source)
+				if errPath != nil {
+					return errPath
 				}
-				if output, err := newSMBMapping(allOptions[0], allOptions[1], source); err != nil {
-					return fmt.Errorf("New-SmbGlobalMapping remount failed: %v, output: %q", err, output)
+				if valid {
+					err = nil
+					klog.V(2).Infof("SMB Mapping(%s) already exists and is still valid, skip error", source)
+				} else {
+					klog.V(2).Infof("SMB Mapping(%s) already exists while it's not valid, now begin to remove and remount", source)
+					if output, err = removeSMBMapping(source); err != nil {
+						return fmt.Errorf("Remove-SmbGlobalMapping failed: %v, output: %q", err, output)
+					}
+					output, err = newSMBMapping(username, password, source)
 				}
-			} else {
-				return fmt.Errorf("New-SmbGlobalMapping failed: %v, output: %q", err, output)
 			}
+		}
+		if err != nil {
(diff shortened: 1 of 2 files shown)


andyzhangx:
how can I fixed the mac test failure? @jingxu97 @msau42 
I cannot find detailed failure logs:
https://github.com/kubernetes/utils/pull/184/checks?check_run_id=965284976

msau42:
@andyzhangx I think longer term we should move towards being able to delete the global mapping once all pods on the node are not using it anymore. Slowly leaking global resources over time will take up resources and potentially cause node instability if there's a frequent churn of unique mounts. I think it will also help improve the scenario you mention where the password changes, so we don't unmount the device while a pod is still using it.

andyzhangx:
> @andyzhangx I think longer term we should move towards being able to delete the global mapping once all pods on the node are not using it anymore. Slowly leaking global resources over time will take up resources and potentially cause node instability if there's a frequent churn of unique mounts. I think it will also help improve the scenario you mention where the password changes, so we don't unmount the device while a pod is still using it.

@msau42 agree, in long term we should fix it, while still needs to figure out how to get global mapping according to the pod mount path, that's a blocking issue AFAIK.
can we make this PR merge first? I would also like to make same fix in CSI-Proxy after this merged? thanks.
cc @jingxu97

msau42:
Yes that sounds like a good plan

/assign @jingxu97 
for review

mount/mount_windows.go
```diff
@@ -98,19 +98,32 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
 		getSMBMountMutex.LockKey(source)
 		defer getSMBMountMutex.UnlockKey(source)
 
-		if output, err := newSMBMapping(allOptions[0], allOptions[1], source); err != nil {
+		var output string
+		var err error
+		username := allOptions[0]
+		password := allOptions[1]
```
jingxu97, line 104: could we fail if options has less than 2 args?

mount/mount_windows.go
```diff
@@ -98,19 +98,32 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
...
+		if output, err = newSMBMapping(username, password, source); err != nil {
+			klog.Warningf("SMB Mapping(%s) returned with error(%v), output(%s)", source, err, string(output))
```
liggitt, line 106: will this output contain the username/password? if so, it must not be logged

andyzhangx: no, it won't, it uses `sensitiveOptionsRemoved` var in the error msg:
https://github.com/kubernetes/utils/blob/e0bba7c5b4d1bddbe8503553fe6d0389f06e2172/mount/mount_windows.go#L142-L144

mount/mount_windows.go
andyzhangx: I think yes, let me work out another PR to fix it.

PR #186, opened 2020-08-15
fix: return error with fewer mount options on Windows

andyzhangx:
as discussed in https://github.com/kubernetes/utils/pull/180, return error with fewer mount options on Windows
/assign @msau42 @jingxu97

### DIFF
--- a/mount/mount_windows.go
+++ b/mount/mount_windows.go
@@ -84,9 +84,8 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
 		allOptions = append(allOptions, options...)
 		allOptions = append(allOptions, sensitiveOptions...)
 		if len(allOptions) < 2 {
-			klog.Warningf("mount options(%q) command number(%d) less than 2, source:%q, target:%q, skip mounting",
+			return fmt.Errorf("mount options(%q) should have at least 2 options, current number:%d, source:%q, target:%q",
 				sanitizedOptionsForLogging, len(allOptions), source, target)
-			return nil
 		}
 
 		// currently only cifs mount is supported


mount/mount_windows.go
```diff
@@ -84,9 +84,8 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
 		allOptions = append(allOptions, options...)
 		allOptions = append(allOptions, sensitiveOptions...)
 		if len(allOptions) < 2 {
-			klog.Warningf("mount options(%q) command number(%d) less than 2, source:%q, target:%q, skip mounting",
+			return fmt.Errorf("mount options(%q) should have at least 2 options, current number:%d, source:%q, target:%q",
 				sanitizedOptionsForLogging, len(allOptions), source, target)
```
jingxu97, line 88: It is up to the called to make sure using senstiveOptions for password, but it there a way we can verify that?

andyzhangx: actually `sensitiveOptions` is not used on Windows:
```
func (mounter *Mounter) Mount(source string, target string, fstype string, options []string) error {
	return mounter.MountSensitive(source, target, fstype, options, nil /* sensitiveOptions */)
}
```

msau42: Is that a concern?

andyzhangx: @msau42 I don't think so, windows has different implementation with linux, and as long as we don't print password in logging, that would be ok

jingxu97: actually could you verify that whether it will be printed out or not?
The caller (driver) should call MountSensitive instead of Mount if it has password information, I think.

andyzhangx: I checked the code again, use `MountSensitive`  func won't print password.

jingxu97: thanks. After kubernetes/kubernetes#94126 is merged, we can merge this.

andyzhangx: hi @jingxu97 that's not related, `len(allOptions)` would be changed by https://github.com/kubernetes/kubernetes/pull/94126

jingxu97: I see. thanks!

PR #187, opened 2020-09-08
fix inotify memory leak

yanghaichao12:
**What type of PR is this?**
/kind bug

fix inotify memory leak
Fixes  the issue
https://github.com/kubernetes/kubernetes/issues/94583

**Release note**:


go.mod
```diff
@@ -6,5 +6,6 @@ require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/spf13/afero v1.2.2
 	github.com/stretchr/testify v1.3.0
+	golang.org/x/lint v0.0.0-20200302205851-738671d3881b // indirect
```
dims, line 9: please remove this change

go.sum
```diff
@@ -10,7 +10,20 @@ github.com/spf13/afero v1.2.2/go.mod h1:9ZxEEn6pIJ8Rxe320qSDBk6AsU0r9pR7Q4OcevTd
 github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
 github.com/stretchr/testify v1.3.0 h1:TivCn/peBQ7UY8ooIcPgZFpTNSz0Q2U6UrFlUfqbe0Q=
 github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UVUgZn+9EI=
+golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2/go.mod h1:djNgcEr1/C05ACkg1iLfiJU5Ep61QUkGW8qpdssI0+w=
```
dims, line 13: please remove these changes.

hack/verify-golint.sh
```diff
@@ -17,7 +17,6 @@
 set -o errexit
 set -o nounset
 set -o pipefail
-
```
dims, line 20: also an unnecessary change.

inotify/inotify_linux.go
```diff
@@ -134,6 +135,65 @@ func (w *Watcher) RemoveWatch(path string) error {
...
+				// Move to the next event in the buffer
+				offset += syscall.SizeofInotifyEvent + nameLen
+			}
+
+		}
+
+	}()
 
 	for {
 		n, err := syscall.Read(w.fd, buf[:])
```
liggitt, line 199: readEvents is already called in a goroutine... I'm having a hard time understanding the scenario that is fixed by moving the inner portion of this loop into a second goroutine

liggitt:
a unit test exercising the scenario this is fixing (which fails on master without this change) would be helpful

hack/verify-golint.sh
yanghaichao12: okay,i'll do it later

liggitt:
I saw a few other areas that might need fixing:

1. Fix `Watcher#AddWatch` to handle concurrent calls by checking `watchEntry, found := w.watches[path]` again after taking the lock
2. Fix `Watcher#RemoveWatch` to handle concurrent calls by protecting w.watches

From the original issue:

> i make some trace and found some watched objects have never been released

Do you have the paths that were being watched? Do we know where those were registered?

> and print the length of every event，some of which is beyond the length of buf, even when i increase the array by 10 times，the problem is still there and the limits are broken through again and again. so i realized the process of reading cannot be blocked.

do we have details about the problematic events?

yanghaichao12:
@liggitt hi, at first, i plan to have a new consumer and producer to replace 'readEvents', but i don't want to modify the caller for it and i think unlike 'process', goroutine are independent to each other, so....
below is my data,
image
the units of vertical axis is bytes，the highest is original code；the scecond is my first modification， i just add a queue for events
https://github.com/kubernetes/utils/blob/5e75c0163111533bca81c92d71ee05d00ff07574/inotify/inotify_linux.go#L194
as you see ， it's not so good and i do another changes. the result is better  and i use pprof to watch heap, and this function 'readEvents'  is not top1 now.

inotify/inotify_linux.go
```diff
@@ -165,37 +225,11 @@ func (w *Watcher) readEvents() {
...
-					event.Name += "/" + strings.TrimRight(string(bytes[0:nameLen]), "\000")
-				}
-				// Send the event on the events channel
-				w.Event <- event
-			}
-			// Move to the next event in the buffer
-			offset += syscall.SizeofInotifyEvent + nameLen
-		}
+		cond.L.Lock()
+		bufs = append(bufs, bufn{buf, n})
```
liggitt, line 229: This change appears to be repeatedly reading into the same buffer (`buf`), then adding a reference to the same memory to a queue (`bufs`) which is asynchronously processed. Doesn't that mean the `syscall.Read(w.fd, buf[:])` call is overwriting data that is queued to be processed?

If I'm reading this correctly, this change is "improving" memory use in a way that breaks the data being processed.

liggitt: is it possible that the root cause might be a slow event reader that never drains the file event channel?

yanghaichao12:
@liggitt hi,it's the profile before modification
image

image

inotify/inotify_linux.go
yanghaichao12: yes, there are two channel are used: one's size is 16,another is 1.
in fact， i've tried to do two other changes：
the first is，i had make it  a goroutine
https://github.com/kubernetes/kubernetes/blob/3fc1bc76a95d8a6a6cd33394526f00c00a98ce07/vendor/github.com/google/cadvisor/container/raw/watcher.go#L230
another  is：i change the bufsize from SizeofInotifyEvent * 4096 to SizeofInotifyEvent * 40960, but that‘s also no use. 
the channels are always blocked finnally

liggitt: > i change the bufsize from SizeofInotifyEvent * 4096 to SizeofInotifyEvent * 40960, but that‘s also no use.

I don't think that is the problematic buffer... it seems like w.Event blocking is the issue

yanghaichao12:
@liggitt hi, It seems a little difficult for me to do performance testing with unit test cases......
if fact, i try my best to design a appropriate unit test but can't.  could you give more suggetion? i would  appriciate it . i'v done some  performance testing and profile analysis, some results are shown above

inotify/inotify_linux.go
yanghaichao12: right, i think so too for now

yanghaichao12:
@liggit  hi， below is the first version 

```go
var buf [syscall.SizeofInotifyEvent * 40960]byte

	for {
		n, err := syscall.Read(w.fd, buf[:])
		// See if there is a message on the "done" channel
		var done bool
		select {
		case done = <-w.done:
		default:
		}

		// If EOF or a "done" message is received
		if n == 0 || done {
			// The syscall.Close can be slow.  Close
			// w.Event first.
			close(w.Event)
			err := syscall.Close(w.fd)
			if err != nil {
				w.Error <- os.NewSyscallError("close", err)
			}
			close(w.Error)
			return
		}
		if n < 0 {
			w.Error <- os.NewSyscallError("read", err)
			continue
		}
		if n < syscall.SizeofInotifyEvent {
			w.Error <- errors.New("inotify: short read in readEvents()")
			continue
		}

		var offset uint32
		//yanghaichao beg
		var event_array []*Event
		var locker = new(sync.Mutex)
		var cond = sync.NewCond(locker)
		go func() {
			for {
				cond.L.Lock()
				if event_array == nil || len(event_array) ==0{
					cond.Wait()
				}
				w.Event <- event_array[0]
				event_array = event_array[1:]
				cond.L.Unlock()
			}

		}()
		//yanghaichao end



		// We don't know how many events we just read into the buffer
		// While the offset points to at least one whole event...
		for offset <= uint32(n-syscall.SizeofInotifyEvent) {
			// Point "raw" to the event in the buffer
			raw := (*syscall.InotifyEvent)(unsafe.Pointer(&buf[offset]))
			event := new(Event)
			event.Mask = uint32(raw.Mask)
			event.Cookie = uint32(raw.Cookie)
			nameLen := uint32(raw.Len)
			// If the event happened to the watched directory or the watched file, the kernel
			// doesn't append the filename to the event, but we would like to always fill the
			// the "Name" field with a valid filename. We retrieve the path of the watch from
			// the "paths" map.
			w.mu.Lock()
			name, ok := w.paths[int(raw.Wd)]
			w.mu.Unlock()
			if ok {
				event.Name = name
				if nameLen > 0 {
					// Point "bytes" at the first byte of the filename
					bytes := (*[syscall.PathMax]byte)(unsafe.Pointer(&buf[offset+syscall.SizeofInotifyEvent]))
					// The filename is padded with NUL bytes. TrimRight() gets rid of those.
					event.Name += "/" + strings.TrimRight(string(bytes[0:nameLen]), "\000")
				}
				// Send the event on the events channel
                 //yanghaichao move//w.Event <- event

				//yanghaichao beg
				cond.L.Lock()
				event_array = append(event_array,event)
				cond.L.Unlock()
				cond.Signal()
				//yanghaichao end
			}
			// Move to the next event in the buffer
			offset += syscall.SizeofInotifyEvent + nameLen
		}
	}
```
the red part is memory usage:
image

 and buf [syscall.SizeofInotifyEvent * **40960**] is overflow very ten seconds， so i have now version. but as you say, i dont think it's 
correct now..... :(
maybe i should do some other changes and more testing....

yanghaichao12:
i did more observation, and found there are a lot of dead container that haven't been recycled:

```shell
Every 2.0s: docker ps -a|grep Exited|wc -l                                                                                         Fri Sep 11 16:26:07 2020

6225

```
the logs show the problem:
```shell
I0911 10:40:00.767586    5891 cni.go:380] Deleting test_rc1-2xzzk/2e9c91c76a8b3f496a74dd866bd0ff467515cf33ab7d0c39f05301bf05b76a24 from network knitter/knitter netns ""
I0911 10:40:15.804967    5891 cni.go:388] Deleted test_rc1-2xzzk/2e9c91c76a8b3f496a74dd866bd0ff467515cf33ab7d0c39f05301bf05b76a24 from network knitter/knitter
```
it's too slow for deleting network. it took 15s every time.

then  i checked the parameter 
```shell
ubuntu@dev-lzx:/proc/sys/fs/inotify$ cat /proc/sys/fs/inotify/max_queued_events
16384
```

```bash
   /proc/sys/fs/inotify/max_queued_events
              The value in this file is used when an application calls inotify_init(2) to set an upper limit on the number of events that can be queued
              to the corresponding inotify instance.  Events in excess of this limit are dropped, but an IN_Q_OVERFLOW event is always generated.
```

inotify/inotify_linux.go
```diff
@@ -134,6 +135,22 @@ func (w *Watcher) RemoveWatch(path string) error {
 // received events into Event objects and sends them via the Event channel
 func (w *Watcher) readEvents() {
 	var buf [syscall.SizeofInotifyEvent * 4096]byte
+	var eventArray []*Event
```
Cweiping, line 138: `eventArray ` will be shared by goroutine, which seems more prone to problems.

yanghaichao12: hi，it must be accesse with lock，you could see that below’

Cweiping: It's just my confusion, not the code.
_Instead of communicating through shared memory, share memory through communication._

thockin:
Hi, What's the status on this?  Is it still valid?  Should it be reviewed now?  Sorry I lost track...

wzshiming:
Hi @yanghaichao12 
I found that this PR fix may be wrong, the main problem is that the watcher is not closed. https://github.com/kubernetes/kubernetes/pull/100326

attlee-wang:
> Hi @yanghaichao12
> I found that this PR fix may be wrong, the main problem is that the watcher is not closed. [kubernetes/kubernetes#100326](https://github.com/kubernetes/kubernetes/pull/100326)

I fix [kubernetes/kubernetes#100326](https://github.com/kubernetes/kubernetes/pull/100326) and  [# 3014 ]https://github.com/google/cadvisor/pull/3014] to my k8s, but [#3218 ](https://github.com/google/cadvisor/issues/3218) problem still exists.  @yanghaichao12 @liggitt

thockin:
@attlee-wang you can just check this PR out and open a new one from your own fork

pacoxu:
> I fix [kubernetes/kubernetes#100326](https://github.com/kubernetes/kubernetes/pull/100326) and [# 3014 ](https://github.com/google/cadvisor/pull/3014) to my k8s, but [#3218](https://github.com/google/cadvisor/issues/3218) problem still exists.

You can checkout the release branch you are using, for instance, `git checkout release-v1.24` and cherry-pick this fix to your branch and build a customized kubelet binary. Test it in your cluster.
```
build/run.sh make kubelet
```

PR #188, opened 2020-09-10
fix: smb valid path error

andyzhangx:
**What type of PR is this?**
/kind bug

fix: smb valid path error, continue fix with https://github.com/kubernetes/utils/pull/184

`isValidPath` func error should be ignored since if password changed, it would return following access error:
```
PathValid(\\xxx\share) failed with returned output: Test-Path : Access is denied
```

By that case, should remove original SMB remote path and mount again.

Fixes #

/assign @jingxu97 

**Release note**:
```
fix: smb valid path error
```

### DIFF
--- a/mount/mount_windows.go
+++ b/mount/mount_windows.go
@@ -29,6 +29,10 @@ import (
 	"k8s.io/utils/keymutex"
 )
 
+const (
+	accessDenied string = "access is denied"
+)
+
 // Mounter provides the default implementation of mount.Interface
 // for the windows platform.  This implementation assumes that the
 // kubelet is running in the host's root mount namespace.
@@ -97,31 +101,28 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
 		getSMBMountMutex.LockKey(source)
 		defer getSMBMountMutex.UnlockKey(source)
 
-		var output string
-		var err error
 		username := allOptions[0]
 		password := allOptions[1]
-		if output, err = newSMBMapping(username, password, source); err != nil {
+		if output, err := newSMBMapping(username, password, source); err != nil {
 			klog.Warningf("SMB Mapping(%s) returned with error(%v), output(%s)", source, err, string(output))
 			if isSMBMappingExist(source) {
-				valid, errPath := isValidPath(source)
-				if errPath != nil {
-					return errPath
-				}
-				if valid {
-					err = nil
-					klog.V(2).Infof("SMB Mapping(%s) already exists and is still valid, skip error", source)
-				} else {
-					klog.V(2).Infof("SMB Mapping(%s) already exists while it's not valid, now begin to remove and remount", source)
-					if output, err = removeSMBMapping(source); err != nil {
-						return fmt.Errorf("Remove-SmbGlobalMapping failed: %v, output: %q", err, output)
+				valid, err := isValidPath(source)
+				if !valid {
+					if err == nil || isAccessDeniedError(err) {
+						klog.V(2).Infof("SMB Mapping(%s) already exists while it's not valid, return error: %v, now begin to remove and remount", source, err)
+						if output, err = removeSMBMapping(source); err != nil {
+							return fmt.Errorf("Remove-SmbGlobalMapping failed: %v, output: %q", err, output)
+						}
+						if output, err := newSMBMapping(username, password, source); err != nil {
+							return fmt.Errorf("New-SmbGlobalMapping(%s) failed: %v, output: %q", source, err, output)
+						}
 					}
-					output, err = newSMBMapping(username, password, source)
+				} else {
+					klog.V(2).Infof("SMB Mapping(%s) already exists and is still valid, skip error(%v)", source, err)
 				}
+			} else {
+				return fmt.Errorf("New-SmbGlobalMapping(%s) failed: %v, output: %q", source, err, output)
 			}
-		}
-		if err != nil {
-			return fmt.Errorf("New-SmbGlobalMapping(%s) failed: %v, output: %q", source, err, output)
 		}
 	}
 
@@ -176,6 +177,10 @@ func isValidPath(remotepath string) (bool, error) {
 	}
 
 	return strings.HasPrefix(strings.ToLower(string(output)), "true"), nil
+}
+
+func isAccessDeniedError(err error) bool {
+	return err != nil && strings.Contains(strings.ToLower(err.Error()), accessDenied)
 }
 
 // remove SMB mapping
--- a/mount/mount_windows_test.go
+++ b/mount/mount_windows_test.go
@@ -361,3 +361,29 @@ func TestIsValidPath(t *testing.T) {
 		}
 	}
 }
+
+func TestIsAccessDeniedError(t *testing.T) {
+	tests := []struct {
+		err            error
+		expectedResult bool
+	}{
+		{
+			nil,
+			false,
+		},
+		{
+			fmt.Errorf("other error message"),
+			false,
+		},
+		{
+			fmt.Errorf(`PathValid(\\xxx\share) failed with returned output: Test-Path : Access is denied`),
+			true,
+		},
+	}
+
+	for _, test := range tests {
+		result := isAccessDeniedError(test.err)
+		assert.Equal(t, result, test.expectedResult, "Expect result not equal with isAccessDeniedError(%v) return: %q, expected: %q",
+			test.err, result, test.expectedResult)
+	}
+}


mount/mount_windows.go
```diff
@@ -106,7 +106,7 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
 			if isSMBMappingExist(source) {
 				valid, errPath := isValidPath(source)
 				if errPath != nil {
-					return errPath
+					klog.Warningf("isValidPath(%s) returned with error(%v), ignore error", source, errPath)
```
jingxu97, line 109: how about

if valid && errPath == nil {
     //log valid path
} else {
     //log 
     //remove existing one, create a new one..
}

andyzhangx: done, PTAL.

mount/mount_windows.go
```diff
@@ -105,14 +105,11 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
 			klog.Warningf("SMB Mapping(%s) returned with error(%v), output(%s)", source, err, string(output))
 			if isSMBMappingExist(source) {
 				valid, errPath := isValidPath(source)
-				if errPath != nil {
-					return errPath
-				}
-				if valid {
+				if valid && errPath == nil {
 					err = nil
```
jingxu97, line 109: can we avoid setting err = nil here?

mount/mount_windows.go
```diff
@@ -105,14 +105,11 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
...
 					klog.V(2).Infof("SMB Mapping(%s) already exists and is still valid, skip error", source)
 				} else {
-					klog.V(2).Infof("SMB Mapping(%s) already exists while it's not valid, now begin to remove and remount", source)
+					klog.V(2).Infof("SMB Mapping(%s) already exists while it's not valid, return error: %v, now begin to remove and remount", source, errPath)
 					if output, err = removeSMBMapping(source); err != nil {
 						return fmt.Errorf("Remove-SmbGlobalMapping failed: %v, output: %q", err, output)
 					}
```
jingxu97, line 115: return the error too if the newSMBMapping below failed. So line 120, can log success case.

mount/mount_windows.go
```diff
@@ -105,14 +105,11 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
...
-				if errPath != nil {
-					return errPath
-				}
-				if valid {
+				if valid && errPath == nil {
 					err = nil
 					klog.V(2).Infof("SMB Mapping(%s) already exists and is still valid, skip error", source)
 				} else {
-					klog.V(2).Infof("SMB Mapping(%s) already exists while it's not valid, now begin to remove and remount", source)
+					klog.V(2).Infof("SMB Mapping(%s) already exists while it's not valid, return error: %v, now begin to remove and remount", source, errPath)
```
msau42, line 112: Could `errPath` be some temporary error that resolves itself on the next call?

Does removing the global mapping impact running containers?

mount/mount_windows.go
andyzhangx: fixed

mount/mount_windows.go
andyzhangx: fixed

mount/mount_windows.go
andyzhangx: it could be temp error, so pod accessing that path at that time would still get error, so remove the global mapping has the same impact since at that time, that path is inaccessible either

msau42: Could it be some temporary error that only impacts this particular call, like a timeout or network issue?

andyzhangx: if it's a temporary error, remove and create new connection is also ok, since at that time, connection was broken.

msau42: Ok I'm mainly concerned that if there's some temporary error and we cannot reconnect for some reason, then we'll be breaking all existing mounts.

Typically a mount for one Pod should not impact mounts for running Pods.

jingxu97: In the case of password change scenario, there are two options
1 remove old connection, make new connection so allowing new pods work, but it seems affect existing ones. Or
2. allow new pods fail while keeping existing ones work.

Similar, for other type of errors, we might take different options based on the error.

andyzhangx: if password changed, existing connection would be invalid, remove old connection would be the only option.

andyzhangx: Windows smb mount is different than Linux, on Windows, the smb mount is shared, so at that time, if connection is broken, all running pods using that smb mount are broken too, I think remove and create new connection is ok.

jingxu97: Thanks, Andy. I think the only concern left is how reliable of validating path is. For example, in case of some temporal network error, the checking failed. In the case, we want to keep the existing mount connection. Is that possible to identify such temporal network error?

msau42:
The mount repo has been moved to https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/mount-utils. Can we move this change to there?

andyzhangx:
> The mount repo has been moved to https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/mount-utils. Can we move this change to there?

sure, let's merge this PR first.

mount/mount_windows.go
andyzhangx: thanks, I did a change first, remove old connection and create new connection only happens when there is `Access is denied` error

mount/mount_windows.go
```diff
@@ -97,32 +101,25 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
...
-				if errPath != nil {
-					return errPath
-				}
-				if valid {
-					err = nil
-					klog.V(2).Infof("SMB Mapping(%s) already exists and is still valid, skip error", source)
-				} else {
-					klog.V(2).Infof("SMB Mapping(%s) already exists while it's not valid, now begin to remove and remount", source)
+				valid, err := isValidPath(source)
+				if !valid && err != nil && strings.Contains(strings.ToLower(err.Error()), accessDenied) {
```
jingxu97, line 110: nit: make a function isAccessDenied() to check error is this type?

mount/mount_windows.go
jingxu97: is there any other type of error we are sure that old connection will not work, and better to delete it and create a new one?

mount/mount_windows.go
andyzhangx: I think I could move this logic into `isValidPath` func

mount/mount_windows.go
andyzhangx: > is there any other type of error we are sure that old connection will not work, and better to delete it and create a new one?

@jingxu97 I don't know, let's fix this error type first.

andyzhangx:
@jingxu97 @msau42 could you take a look? thanks.

mount/mount_windows.go
```diff
@@ -97,32 +101,27 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
...
-				if errPath != nil {
-					return errPath
-				}
-				if valid {
-					err = nil
-					klog.V(2).Infof("SMB Mapping(%s) already exists and is still valid, skip error", source)
-				} else {
-					klog.V(2).Infof("SMB Mapping(%s) already exists while it's not valid, now begin to remove and remount", source)
+				valid, err := isValidPath(source)
+				if err == nil && !valid {
```
jingxu97, line 110: I understand your logic, just thinking how to return the error in a more clear way. How about 
option1:
if isValid() return valid, move on,
if isValid() return error, call shouldRemovePath(error) to check whether the error indicates removing path is the only choice
optino2:
isValid return three value, (valid, shouldRemvoe, error) to clearly indicate the invalid path should be removed. 

What do you think? @andyzhangx @msau42

andyzhangx: thanks, PTAL.

mount/mount_windows.go
```diff
@@ -97,32 +101,31 @@ func (mounter *Mounter) MountSensitive(source string, target string, fstype stri
...
-		if output, err = newSMBMapping(username, password, source); err != nil {
+		if output, err := newSMBMapping(username, password, source); err != nil {
 			klog.Warningf("SMB Mapping(%s) returned with error(%v), output(%s)", source, err, string(output))
 			if isSMBMappingExist(source) {
-				valid, errPath := isValidPath(source)
-				if errPath != nil {
-					return errPath
+				valid, err := isValidPath(source)
+				shouldRemovePath := (err == nil && !valid)
+				if err != nil && strings.Contains(strings.ToLower(err.Error()), accessDenied) {
```
jingxu97, line 111: sorry I kind of misunderstand the logic a little bit before. So if no error, and path is not valid, it will remove mapping..

valid, err := isValidPath(source)
if !valid {
   if err == nil || isAccessDenied(err) {
       // perform remove mapping and create a new one
   } else {
      // log and return error
  }
}

func isAccessDenied (err error) bool {
    return strings.Contains(strings.ToLower(err.Error()), accessDenied)
}

andyzhangx: yes, if `!valid`, should still remove mapping

andyzhangx: there are two conditions that should remove mapping: 
 - `err == nil && !valid`
 -  `isAccessDenied(err)`

jingxu97: isAccessDenied(err) also means !valid, right?

jingxu97:
```
valid, err := isValidPath(source)
if !valid {
   if err == nil || isAccessDenied(err) {
       // perform remove mapping and create a new one
   } else { // err might be temporal, return err and will retry later
      // log and return error
  }
} // if valid, do nothing
```

andyzhangx: thanks, I have made the changes, PTAL.

PR #192, opened 2020-09-24
[WIP] move iptables and conntrack utils out of k/k

aojea:
POC for moving iptables and conntrack out of kubernetes repo

Those are very useful libraries, specially iptables, and there are no similar ones in the golang ecosystem


danwinship:
The README claims you're supposed to follow the directions in https://github.com/kubernetes/utils/blob/master/HOWTOMOVE.md to move things with all of their git history, though that clearly hasn't been done in a lot of cases...

dims:
@aojea @danwinship i think it would be better to create a new repo for this instead of `k8s.io/utils`

@thockin ^^^

dims:
+1 to preserve history in the new repo.

thockin:
2 issues:

1) the third_party is unpleasant.  Is that to break cycles?  I don't see any.  Could we coax the apimachinery repos to move those 5 or 6 libs to utils?  Or could we remove the deps on them?

2) @dims We could do a different repo.  I suggested that in the original discussion too.  This repo already has "ebtables" which is very similar.  If we made a new repo, I'd love to EOL this ebtables, too, but that is hard.  What is motivating the suggestion from your POV?

aojea:
> 1. the third_party is unpleasant.  Is that to break cycles?  I don't see any.  Could we coax the apimachinery repos to move those 5 or 6 libs to utils?  Or could we remove the deps on them?

this was not supposed to be a proposal, just to have an idea of the code that will needed to move if we don't want to import k8s.io/apimachinery. 
I don't know if t it should be possible to move those libraries easily, lot of projects depend on it.
I think that having a staging repo will allow to keep the code inside k/k and solve the dependency problem, no?

dims:
@thockin the `mount` package which moved from `staging/` to `utils` went back to `staging/src/k8s.io/mount-utils` 

All the reasons that went into it is documented here:
https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20200508-move-utils-mount-2staging.md#motivation

Basically they were not able to test things and we ended up breaking k8s when changes were made in utils/ and then vendored in

Do we have a mitigation strategy to avoid the same situation happening with this set of code?

aojea:
@thockin what about moving this to the new staging repo component-helpers https://github.com/kubernetes/kubernetes/pull/92507/files

`k8s.io/component-helpers/networking/{iptables,ebtables,hostport}`

danwinship:
That sounds like exactly what we're looking for

rikatz:
https://github.com/kubernetes/kubernetes/pull/100931 <- Started to move some functions internal to the API (as sig scheduling), I was wondering if the iptables/ipvs/conntrack should be kept on components-helper repo, or if we should create a new network-utils one.

Issue #196, opened 2021-01-03
Issues with Clock


ash2k:
https://github.com/kubernetes/utils/blob/67b214c5f920060488eda08cd2b5c30597f2f1d7/clock/clock.go#L31-L37

1. `After(d time.Duration) <-chan time.Time` does not allow to stop the underlying timer to free resources if it's no longer needed. This can lead to memory leaks when a lot of long-lived timers pile up. It's a [well-known pitfall](https://medium.com/@oboturov/golang-time-after-is-not-garbage-collected-4cbc94740082). The fix is to remove the method and use the `Timer()` method instead.

2. `Tick(d time.Duration) <-chan time.Time` has the same issue as `After()`. This method should be replaced with [`NewTicker()`](https://github.com/kubernetes/apimachinery/blob/15c5dba13c59f42b25aba4e3a6b17061af770216/pkg/util/clock/clock.go#L39) from the sister package (which I added here https://github.com/kubernetes/kubernetes/pull/61976).

3. `Sleep(d time.Duration)` should take `context.Context` to allow for early interruption of sleep. This method is used e.g. in client-go to [sleep between request retries](https://github.com/kubernetes/client-go/blob/e9d996dae239d57f7d96af0c3cc19e3b6e2eaf83/rest/urlbackoff.go#L105-L107) and often there is a context to signal that the attempt should be aborted. Currently there is no way to interrupt this sleep.

ash2k:
I'd like to replace https://github.com/kubernetes/kubernetes/pull/93709 with clock usage.

liggitt:
Removing methods from the interface is not API compatible, and is not permitted for the utils library due to its wide use across versions and across projects.

Adding methods to surface the NewTimer and NewTicker methods provided by the `time` package and marking the `After` and `Tick` methods deprecated (but not removing them) could be reasonable.

> Sleep(d time.Duration) should take context.Context to allow for early interruption of sleep. This method is used e.g. in client-go to sleep between request retries and often there is a context to signal that the attempt should be aborted. Currently there is no way to interrupt this sleep.

These methods mirror the ones provided by the `time` package. There is no `Sleep(context.Context, time.Duration)` method, so I would not expect our clock fake to provide that.

ash2k:
[...]

PR #197, opened 2021-01-10
Add NewTicker() to Clock

ash2k:
**What type of PR is this?**
/kind feature

See linked issues. TL;DR adds a missing method to allow safer usage and to bridge the gap between two copies of the clock package so that the one in apimachinery can be deleted.

Fixes https://github.com/kubernetes/utils/issues/196.

Updates https://github.com/kubernetes/kubernetes/issues/94738.

**Release note**:
```
NONE
```

### DIFF
--- a/clock/clock.go
+++ b/clock/clock.go
@@ -30,13 +30,36 @@ type PassiveClock interface {
 // needs to do arbitrary things based on time.
 type Clock interface {
 	PassiveClock
+	// After returns the channel of a new Timer.
+	// This method does not allow to free/GC the backing timer before it fires. Use
+	// NewTimer instead.
 	After(d time.Duration) <-chan time.Time
+	// NewTimer returns a new Timer.
 	NewTimer(d time.Duration) Timer
+	// Sleep sleeps for the provided duration d.
+	// Consider making the sleep interruptible by using 'select' on a context channel and a timer channel.
 	Sleep(d time.Duration)
+	// Tick returns the channel of a new Ticker.
+	// This method does not allow to free/GC the backing ticker. Use
+	// NewTicker from WithTicker instead.
 	Tick(d time.Duration) <-chan time.Time
 }
 
-var _ = Clock(RealClock{})
+// WithTicker allows for injecting fake or real clocks into code that
+// needs to do arbitrary things based on time.
+type WithTicker interface {
+	Clock
+	// NewTicker returns a new Ticker.
+	NewTicker(time.Duration) Ticker
+}
+
+// Ticker defines the Ticker interface.
+type Ticker interface {
+	C() <-chan time.Time
+	Stop()
+}
+
+var _ = WithTicker(RealClock{})
 
 // RealClock really calls time.Now()
 type RealClock struct{}
@@ -52,6 +75,8 @@ func (RealClock) Since(ts time.Time) time.Duration {
 }
 
 // After is the same as time.After(d).
+// This method does not allow to free/GC the backing timer before it fires. Use
+// NewTimer instead.
 func (RealClock) After(d time.Duration) <-chan time.Time {
 	return time.After(d)
 }
@@ -64,11 +89,21 @@ func (RealClock) NewTimer(d time.Duration) Timer {
 }
 
 // Tick is the same as time.Tick(d)
+// This method does not allow to free/GC the backing ticker. Use
+// NewTicker instead.
 func (RealClock) Tick(d time.Duration) <-chan time.Time {
 	return time.Tick(d)
 }
 
+// NewTicker returns a new Ticker.
+func (RealClock) NewTicker(d time.Duration) Ticker {
+	return &realTicker{
+		ticker: time.NewTicker(d),
+	}
+}
+
 // Sleep is the same as time.Sleep(d)
+// Consider making the sleep interruptible by using 'select' on a context channel and a timer channel.
 func (RealClock) Sleep(d time.Duration) {
 	time.Sleep(d)
 }
@@ -101,4 +136,16 @@ func (r *realTimer) Stop() bool {
 // Reset calls Reset() on the underlying timer.
 func (r *realTimer) Reset(d time.Duration) bool {
 	return r.timer.Reset(d)
+}
+
+type realTicker struct {
+	ticker *time.Ticker
+}
+
+func (r *realTicker) C() <-chan time.Time {
+	return r.ticker.C
+}
+
+func (r *realTicker) Stop() {
+	r.ticker.Stop()
 }
--- a/clock/testing/fake_clock.go
+++ b/clock/testing/fake_clock.go
@@ -25,8 +25,8 @@ import (
 
 var (
 	_ = clock.PassiveClock(&FakePassiveClock{})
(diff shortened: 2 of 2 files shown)


liggitt:
Thanks for the PR. make sure all the new methods in the interface have associate godoc.

Also, since golang has not marked the time methods as deprecated, I don't think we should mark ours as deprecated. Doing so will cause golint to fail.

liggitt:
I also suspect adding a method to the interface will get flagged as an API incompatibility in the apidiff presubmit (which is currently failing early because of a github actions change)

apelisse:
You can work around the API change by making a "ClockV2" Interface that embeds the other one and adds the method. And a "NewClockV2" that returns that interface (but actually returns the exact same object).

ash2k:
@apelisse Thanks for the tip. I've pushed the changes.

clock/clock.go
```diff
@@ -30,13 +30,36 @@ type PassiveClock interface {
...
+	// Tick returns the channel of a new Ticker.
+	// This method does not allow to free/GC the backing ticker. Use
+	// NewTicker instead.
 	Tick(d time.Duration) <-chan time.Time
 }
 
-var _ = Clock(RealClock{})
+// V2 allows for injecting fake or real clocks into code that
+// needs to do arbitrary things based on time.
+type V2 interface {
```
apelisse, line 50: V2 is not a great name :-)

apelisse: Since the interfaces are pretty much frozen once built, I think you can call it "ClockWithTicker" or something like this.

ash2k: I've tried both `ClockV2` and `ClockWithTicker` but linted doesn't like such names - see https://github.com/kubernetes/utils/pull/197/checks?check_run_id=1685665492#step:4:62 That's why I tried `V2`, which you don't like :)

apelisse: Yeah, "WithTicker" is fine (to me at least ;-))

ash2k: @apelisse PTAL

ash2k:
@apelisse @liggitt @lavalamp Would be awesome to get this merged (I think I've addressed all comments) so that I can start working on https://github.com/kubernetes/kubernetes/issues/94738.

clock/clock.go
```diff
@@ -30,13 +30,36 @@ type PassiveClock interface {
...
+	// NewTimer instead.
 	After(d time.Duration) <-chan time.Time
+	// NewTimer returns a new Timer.
 	NewTimer(d time.Duration) Timer
+	// Sleep sleeps for the provided duration d.
+	// Consider making the sleep interruptible by using 'select' on a context channel and a timer channel.
 	Sleep(d time.Duration)
+	// Tick returns the channel of a new Ticker.
+	// This method does not allow to free/GC the backing ticker. Use
+	// NewTicker instead.
```
lavalamp, line 44: Explain: how does one call NewTicker? It's not in this interface?

clock/clock.go
```diff
@@ -30,13 +30,36 @@ type PassiveClock interface {
...
 	Tick(d time.Duration) <-chan time.Time
 }
 
-var _ = Clock(RealClock{})
+// WithTicker allows for injecting fake or real clocks into code that
+// needs to do arbitrary things based on time.
+type WithTicker interface {
```
lavalamp, line 50: I think it'd be much more convenient if there were either a) a ClockV2 interface, or b) we made a v2 version of the clock package. (and v1 can be a wrapper that just calls the v2 package.) The latter is probably most correct but it might take some mucking about with go mod files.

ash2k: I've tried `ClockV2` - https://github.com/kubernetes/utils/pull/197#discussion_r555531160

I'm not sure how to make a "v2 version of the clock package". I can make a v2 of the whole repo, but I'm not sure what you mean here, sorry.

clock/clock.go
ash2k: I've clarified the comment :)

clock/clock.go
ash2k: I think it might be worth considering a `v2` of the package rather than adding a new clock interface. A new major version shouldn't break anyone who is using go modules.

lavalamp: It's unfortunate if we have to go to v2 of everything in utils. @liggitt might know if we can do some go mod magic to make clock its own module without breaking anyone, and then bump just that to v2. I think it might be possible?

@thockin might have an opinion on whether it's worth rev'ing everything under utils if it's hard to just rev the clock package.

liggitt: > @liggitt might know if we can do some go mod magic to make clock its own module without breaking anyone, and then bump just that to v2. I think it might be possible?

There's no magic. It's possible but really tedious to manage multiple published modules in a single repo (releases have to have per-module tags, etc). I think you'd get about the same effect with a `k8s.io/utils/clock/v2` package

lavalamp: Yeah that actually sounds perfect, if a little bit of work to set up (e.g. we'd want the regular clock package to just call into the v2 package for as much as possible).

clock/testing/fake_clock.go
```diff
@@ -231,6 +249,12 @@ func (*IntervalClock) Tick(d time.Duration) <-chan time.Time {
 	panic("IntervalClock doesn't implement Tick")
 }
 
+// NewTicker is currently unimplemented, will panic.
+// TODO: make interval clock use FakeClock so this can be implemented.
+func (*IntervalClock) NewTicker(d time.Duration) clock.Ticker {
```
apelisse, line 254: Why do we need this function again? If we didn't, IntervalClock wouldn't implement the "WithTicker" interface, which means that it would fail at compile time. By adding it, it now implements WithTicker, but will fail at runtime?

MikeSpreitzer: True.  BTW, IntervalClock also has a lot of unimplemented methods for `Clock`.

MikeSpreitzer: We could remove them all, leaving IntervalClock implementing `PassiveClock` --- which would be a compile-time tight characterization.

MikeSpreitzer: There are no uses of IntervalClock in k8s.io/utils.

In the main repo there are only two files that use it: `client-go/tools/record/event_test.go` and `client-go/tools/record/events_cache_test.go`.

lavalamp: Let's make it possible to remove the api machinery clock. We should also not implement interfaces that don't actually work -- this library is for public consumption and that will be very annoying to a consumer.

MikeSpreitzer: Ugh, there is non-trivial entanglement with code that needs a non-passive clock.  In `client-go/tools/record/events_cache.go`, the Filter method of EventSourceObjectSpamFilter passes its clock to `flowcontrol.NewTokenBucketRateLimiterWithClock` --- which needs a clock that can `Sleep`.  Why does this work?

MikeSpreitzer: @lavalamp : I think #212 does what you said.

ash2k:
Closing in favor of https://github.com/kubernetes/utils/pull/212.

PR #200, opened 2021-02-01
enhance: filter for file systems in the call to blkid

pacoxu:
**What type of PR is this?**
/kind cleanup

https://github.com/kubernetes/kubernetes/pull/59050#issuecomment-457906487
> Why not use '--usages filesystem' to filter for file systems in the call to blkid?

/sig storage

It uses fewer read calls as below. Per https://github.com/kubernetes/kubernetes/pull/59050#issuecomment-457976014
```
# strace blkid -p -s TYPE -s PTTYPE -o export /dev/sda1   |& grep -i '^read(' | wc -l
29
$ strace blkid -p -u filesystems -s TYPE -s PTTYPE -o export /dev/sda1   |& grep -i '^read(' | wc -l
25
```

**Release note**:
```
None
```

### DIFF
--- a/mount/mount_linux.go
+++ b/mount/mount_linux.go
@@ -401,7 +401,7 @@ func (mounter *SafeFormatAndMount) formatAndMountSensitive(source string, target
 
 // GetDiskFormat uses 'blkid' to see if the given disk is unformatted
 func (mounter *SafeFormatAndMount) GetDiskFormat(disk string) (string, error) {
-	args := []string{"-p", "-s", "TYPE", "-s", "PTTYPE", "-o", "export", disk}
+	args := []string{"-p", "-u", "filesystems", "-s", "TYPE", "-s", "PTTYPE", "-o", "export", disk}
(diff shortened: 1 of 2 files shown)


msau42:
The mount library has been moved back to k/k. This mount library is not used anymore.

cc @brahmaroutu we should remove/deprecate this pkg?

pacoxu:
> The mount library has been moved back to k/k. This mount library is not used anymore.
> 
> cc @brahmaroutu we should remove/deprecate this pkg?

Does it mean that `./vendor/k8s.io/utils/mount` would be removed later?
`./staging/src/k8s.io/mount-utils` in k/k would be the right place to hold mount codes?

brahmaroutu:
@pacoxu cadvisor still depend on this package and once that is removed we will remove this repo. 
https://github.com/kubernetes/kubernetes/pull/96130 is waiting for update to cadvisor version
Please use the repo https://github.com/kubernetes/mount-utils

I will work on removing this package soon.

pacoxu:
@brahmaroutu thanks for your information.

/close

PR #202, opened 2021-04-15
Move slice function to utils

rikatz:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**: This is needed to start cleaning up / moving kube-proxy from k/k to its own repo

Part of https://github.com/kubernetes/kubernetes/issues/92369

**Release note**:
```
NONE
```

### DIFF
--- a/slice/slice.go
+++ b/slice/slice.go
@@ -1,0 +1,75 @@
+/*
+Copyright 2015 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package slice provides utility methods for common operations on slices.
+package slice
+
+import (
+	"sort"
+)
+
+// CopyStrings copies the contents of the specified string slice
+// into a new slice.
+func CopyStrings(s []string) []string {
+	if s == nil {
+		return nil
+	}
+	c := make([]string, len(s))
+	copy(c, s)
+	return c
+}
+
+// SortStrings sorts the specified string slice in place. It returns the same
+// slice that was provided in order to facilitate method chaining.
+func SortStrings(s []string) []string {
+	sort.Strings(s)
+	return s
+}
+
+// ContainsString checks if a given slice of strings contains the provided string.
+// If a modifier func is provided, it is called with the slice item before the comparation.
+func ContainsString(slice []string, s string, modifier func(s string) string) bool {
+	for _, item := range slice {
+		if item == s {
+			return true
+		}
+		if modifier != nil && modifier(item) == s {
+			return true
+		}
+	}
+	return false
+}
+
+// RemoveString returns a newly created []string that contains all items from slice that
+// are not equal to s and modifier(s) in case modifier func is provided.
+func RemoveString(slice []string, s string, modifier func(s string) string) []string {
+	newSlice := make([]string, 0)
+	for _, item := range slice {
+		if item == s {
+			continue
+		}
+		if modifier != nil && modifier(item) == s {
+			continue
+		}
+		newSlice = append(newSlice, item)
+	}
+	if len(newSlice) == 0 {
+		// Sanitize for unit tests so we don't need to distinguish empty array
+		// and nil.
+		newSlice = nil
+	}
+	return newSlice
+}
--- a/slice/slice.go
+++ b/slice/slice.go
@@ -17,28 +17,6 @@ limitations under the License.
 // Package slice provides utility methods for common operations on slices.
 package slice
 
-import (
-	"sort"
-)
-
-// CopyStrings copies the contents of the specified string slice
-// into a new slice.
-func CopyStrings(s []string) []string {
-	if s == nil {
-		return nil
-	}
-	c := make([]string, len(s))
-	copy(c, s)
-	return c
-}
-
-// SortStrings sorts the specified string slice in place. It returns the same
-// slice that was provided in order to facilitate method chaining.
-func SortStrings(s []string) []string {
-	sort.Strings(s)
-	return s
-}
-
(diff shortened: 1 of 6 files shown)


rikatz:
Note: That bunch of commits is related to the procedure of maintaining the files history :)

rikatz:
Note2: Probably CLA is failing due to the process being implemented after the files were created

apelisse:
I'm not personally a big fan on preserving a history that is completely irrelevant and confusing :-)

rikatz:
> I'm not personally a big fan on preserving a history that is completely irrelevant and confusing :-)

Ra! Agreed! We're discussing this now in sig-network meeting, @thockin will jump on this :D

apelisse:
Since the source repo is not going anywhere, I think it's reasonable to include instructions to find that history though (former location of the file, commit where it was removed or introduced, etc ...).

rikatz:
@apelisse so I've added a README that points to the original file location.

Maybe once this file gets removed from k/k we can update this README to point to the commit that did this, so the history can still be tracked :)

apelisse:
Thanks, I appreciate that.

/approve

I'm fine with it in principle, I'll happily let someone else look at the code ;-)

rikatz:
@apelisse ok, so I'm reflecting here. I guess the slice util is good, mostly used by volume controller and kubelet.

From networking part (which I'm trying to move/remove from k-proxy), it uses two functions:

`slice.CopyStrings` once, which could easily be migrated to a local function (+ 6 lines if declared only in the function that uses it). BTW this function is only used by pkg/proxy, so this can be removed from utils (no value adding it here, IMO)

The other one is `slice.SortString`, that can be replaced by `sort.Strings(s)` everytime it's called (6 times, to be more precise) in k/k repo.

So I'm wondering if we should keep this open, or, in case of kube-proxy just change the desired code for the extraction above (maybe can also remove the slice.CopyString from original pkg/util).

Thoughts?

slice/slice.go
```diff
@@ -0,0 +1,75 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package slice provides utility methods for common operations on slices.
+package slice
```
apelisse, line 18: From @thockin https://github.com/kubernetes/kubernetes/pull/101194#issuecomment-821455532:
> I would be OK with that idea, too, but I would probably ask for the API to change. e.g.
> `slices.String(variable).Copy()`

I'm quite in favor of creating a:
```
package strings
type Slice []string

func NewSlice(strs ...string) Slice { ... }
```

I think that'd help a lot more with the chaining mentioned in `SortStrings`. I'd also remove the modifier from `ContainString` and `RemoveString` and replace with maybe `Filter`/`Map` though these would require copying (if you're creative you can use a channel). What do you think?

rikatz: Hum, I'm gonna try then. I'll close the other PR and move all the effort here, adding Tim's comments.

About being creative and using channels, I can do for sure, will just try to start with something simpler now and evolve in this PR (I'm bad at channels :P )

rikatz:
@apelisse @thockin I've added the new slice.go inside strings, and moved the 'easy' functions.

CI will fail because I didn't moved the ContainString and RemoveString yet.

Anyway, if you can take a look and check if this is the expected, I'll really appreciate :)

slice/slice.go
thockin: Is the discussion about `slice.String(s).Whatever()` vs `strings.Slice(s).Whatever()` ?

If we had generics, I think we'd obviously lean towards `slices.Slice[string]` as a pattern wouldn't we?  On the other hand, `strings` exists as a package already...

apelisse: > Is the discussion about `slice.String(s).Whatever()` vs `strings.Slice(s).Whatever()` ?

Yeah, not sure it matters so much? Maybe I've seen a discussion about it somewhere.

`slices.Slice` is a little "smurfy", but yeah, we would probably do `Slice[string]` (or whatever the syntax is). But that's a good point, even if we wanted to generate slices for arbitrary types, we would probably all have them in a slice package, and then just generate `strings.go`, `ints.go` etc.

rikatz: Well, this subject being 'decided' I can move back to slice package and have some strings.go again there.

Now, about the "modifiers" function I'm having some hard time figuring out why is this necessary. I've taken a look into Kubernetes code, and every part that uses slices.ContainsString or slices.RemoveString does not passes the modifier function, but instead 'nil'.

So I'm thinking if in this case is worth to migrate also with the modifier (or Map/Filter), as this might add some non used unnecessary complexity.

But I might not be seeing on the long term, so please correct me if I'm wrong :D

Thanks!

rikatz:
In a (far) future, we might migrate everything here to: https://github.com/golang/go/issues/45955

Using generics AND a common package as proposed by Tim, that implements everything in core of Go.

slice/slice.go
uablrek: The "modifier" function is probably a hack to allow for instance "remove all strings beginning with zzz". The entire function should be replaced with a `StringFilter` function, see https://github.com/kubernetes/kubernetes/pull/103331

uablrek:
Ref https://github.com/kubernetes/kubernetes/pull/103331

slice/slice.go
uablrek: IMO we MUST NOT introduce a homebrewed type for `[]string`. I think most go progammers would like to use the builtin type as-is. I would not use a such type myself, it obfuscates the code and even experienced go programers would have to check the code/documentation. The alternative is to type-cast when these functions are used which is not very good either. The `Copy` function is BTW not a copy, it is a clone.

And don't re-do well documented and understandable functions like "sort". If you don't want in-place, copy the slice with `[:]`.

A new type replacing the well known `[]string` and these functions will _not_ make the go code more maintainable.

Please keep the built-in types that everybody knows and limit the functions to those that really add some value, not "sort" and "copy". Filter and map are good candidates, but something like
```go
  filteredSlice := utilstrings.FilterSlice(validator, mySlice)
```

uablrek:
There seems to be a ongoing work to add a "slices" package in go, can we _please_ drop the idea of an own type and keep as close to that wannabe standard as possible?

https://github.com/golang/go/issues/45955

uablrek:
Replaced by https://github.com/kubernetes/utils/pull/208

/close

PR #204, opened 2021-04-26
update the `interger` package for more comprehensive support

BinacsLee:
**What type of PR is this?**

/kind feature

Please refer to https://github.com/kubernetes/kubernetes/pull/101315#issue-620131991 and https://github.com/kubernetes/kubernetes/pull/101315#discussion_r619857147

We need more support in `interger` package.

Fixes #

Thanks for review.

**Release note**:
```
NONE
```

### DIFF
--- a/integer/integer.go
+++ b/integer/integer.go
@@ -32,6 +32,32 @@ func IntMin(a, b int) int {
 	return a
 }
 
+// IntBounded bound the value in range [lower, upper]
+func IntBounded(value, lower, upper int) int {
+	return IntMin(IntMax(value, lower), upper)
+}
+
+// Uint8Max returns the maximum of the params
+func Uint8Max(a, b uint8) uint8 {
+	if b > a {
+		return b
+	}
+	return a
+}
+
+// Uint8Min returns the minimum of the params
+func Uint8Min(a, b uint8) uint8 {
+	if b < a {
+		return b
+	}
+	return a
+}
+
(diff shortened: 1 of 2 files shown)


BinacsLee:
/cc @lavalamp @thockin 
Thanks a lot!

lavalamp:
I'm lukewarm on this change, from the library's perspective it's weird to only support some types, but also I wouldn't like copy-paste of this for every type.

From the perspective of the intended use, I'm actually not convinced changing the `if` statements makes things clearer, and in fact the go ethos has been that a tiny bit of duplication is superior to expanding the dependency tree, I don't think that's changed?

I think I'll be more excited about stuff like this when we get go2 with some type parameterization.

BinacsLee:
> I'm lukewarm on this change, from the library's perspective it's weird to only support some types, but also I wouldn't like copy-paste of this for every type.
> 
> From the perspective of the intended use, I'm actually not convinced changing the `if` statements makes things clearer, and in fact the go ethos has been that a tiny bit of duplication is superior to expanding the dependency tree, I don't think that's changed?
> 
> I think I'll be more excited about stuff like this when we get go2 with some type parameterization.

I understand what you mean and agree with it. Copying functions for every type is really bad (aha, especially for those who have written C++ programs and so on).
Anyway, let us look forward to the go2 with type parameterization ^-^

/close

PR #205, opened 2021-05-20
Add explicit error to ConsistentRead

jsafrane:
**What type of PR is this?**
/kind bug

Return a specific error when `ConsistentRead` fails because it cannot get a consistent content on given file to give caller an opportunity to make better decision what to do about it.

Helps to work around https://github.com/kubernetes/kubernetes/issues/101791
Prerequisite of https://github.com/kubernetes/kubernetes/pull/102059

**Release note**:
```
ConsistentRead now returns distinct error code when it cannot get a consistent read of a file.
```

### DIFF
--- a/io/read.go
+++ b/io/read.go
@@ -30,6 +30,9 @@ var ErrLimitReached = errors.New("the read limit is reached")
 // ConsistentRead repeatedly reads a file until it gets the same content twice.
 // This is useful when reading files in /proc that are larger than page size
 // and kernel may modify them between individual read() syscalls.
+// It returns InconsistentReadError when it cannot get a consistent read in
+// given nr. of attempts. Caller should retry, kernel is probably under heavy
+// mount/unmount load.
 func ConsistentRead(filename string, attempts int) ([]byte, error) {
 	return consistentReadSync(filename, attempts, nil)
 }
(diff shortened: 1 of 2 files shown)


jsafrane:
cc @gnufied @jingxu97, PTAL

gnufied:
returning of specific error in itself is a good idea and hence lgtm.

/lgtm

jsafrane:
Fixed gofmt

jsafrane:
/assign @dims 
for approval.  See https://github.com/kubernetes/kubernetes/pull/102059 for the full story.

io/read.go
```diff
@@ -56,7 +56,25 @@ func consistentReadSync(filename string, attempts int, sync func(int)) ([]byte,
 		// Files are different, continue reading
 		oldContent = newContent
 	}
-	return nil, fmt.Errorf("could not get consistent content of %s after %d attempts", filename, attempts)
+	return nil, InconsistentReadError{filename, attempts}
```
jingxu97, line 59: maybe add this new error type into function comment too?

jsafrane: Added. And added a comment to `InconsistentReadError` type.

dims:
/approve
/lgtm

thanks @jsafrane

PR #207, opened 2021-06-26
Fork Go's ParseIP and ParseCIDR

thockin:
/kind bug

This forks Go's `ParseIP()` and `ParseCIDR()` functions and reverts https://go-review.googlesource.com/c/go/+/325829/ , which changed the parsing to error on leading zeros.  We don't know if anyone has stored objects with those leading zeros, so this represents an unknown break for us.
 
xref https://github.com/kubernetes/kubernetes/issues/100895

**Release note**:
```
Add k8s.io/utils/net.ParseIP() and k8s.io/utils/net.ParseCIDR() functions.  ALL KUBERNETES CODE should use these instead of the standard equivalents.  They are identical in every way except the parsing of leading zeros.
```

cc @liggitt


third_party/forked/golang/net/ip_test.go
```diff
@@ -2,26 +2,38 @@
...
+
+var IPv4Mask = stdnet.IPv4Mask
+
 var parseIPTests = []struct {
 	in  string
 	out IP
 }{
 	{"127.0.1.2", IPv4(127, 0, 1, 2)},
 	{"127.0.0.1", IPv4(127, 0, 0, 1)},
+	{"127.001.002.003", IPv4(127, 1, 2, 3)}, // see https://issue.k8s.io/100895
```
liggitt, line 34: one that demonstrates it parses as decimal as well?

```
{"127.010.020.030", IPv4(127, 10, 20, 30)}, // see https://issue.k8s.io/100895
```

net/parse.go
```diff
@@ -0,0 +1,33 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package net
+
+import (
+	forkednet "k8s.io/utils/third_party/forked/golang/net"
```
liggitt, line 20: given these are locked at a point in time, should we name the package `.../forked/golang_116/net` or something?

net/parse.go
```diff
@@ -0,0 +1,33 @@
...
+// leading '0' characters on numbers.  Go used to allow this and then changed
+// the behavior in 1.17.  We're choosing to keep it for compat with potential
+// stored values.
+var ParseIP = forkednet.ParseIP
+
+// ParseCIDR is identical to Go's standard net.ParseCIDR, except that it allows
+// leading '0' characters on numbers.  Go used to allow this and then changed
+// the behavior in 1.17.  We're choosing to keep it for compat with potential
+// stored values.
+var ParseCIDR = forkednet.ParseCIDR
```
liggitt, line 33: do you think we should name these in a way that indicates why they exist / what the difference is?

liggitt (commented):
Do you envision trying to transition API validation to require new objects to pass the stricter stdlib methods over time?

> ALL KUBERNETES CODE should use these instead of the standard equivalents.  They are identical in every way except the parsing of leading zeros.

That's not going to stick (in this repo or in kubernetes/kubernetes) unless we have some sort of static analysis ensuring that happens. Any ideas for that (starting in this repo would be a good first step).

liggitt:
lgtm, just a couple nits/questions and the painful question of how we ensure consistent use

net/parse.go
liggitt: I also wonder if this is a good place to use `k8s.io/utils/internal/...` so things that import this repo don't use our third_party package directly

thockin: I loooove the idea of internal.  Will do that!  We should do that with the forked reflect too

thockin: As for naming - it's not quite 1.16 - it's more likely that it will be 1.17 (as seldom as this code is touched).  But if it WERE changed, we could update it, so it's not 16 or 17 any more.

net/parse.go
thockin: I waffled on this and fell on "not to".  But since you brought it up too, I will.  `ParseIPSloppy()` seemed apropos?

third_party/forked/golang/net/ip_test.go
thockin: good call.  Also will test 007 vs 008 to prove octal's edge

thockin (commented):
New push is up.  The issue of a static checker needs to be resolved.

net/parse.go
liggitt: yeah, if we're not exporting it I care way less about the exact package name

net/parse.go
liggitt: sgtm

liggitt:
side-benefit of a different name is that a static checker could be easier to write

this is about as dumb a check as I can imagine but would catch a lot in k/k
```
egrep -r '\.(ParseIP|ParseCIDR)\(' pkg staging cmd plugin
```

PR #208, opened 2021-07-01
Add string slice functions

uablrek:
**What type of PR is this?**

/kind cleanup

Part of https://github.com/kubernetes/kubernetes/issues/92369. The first goal is to replace `k8s.io/kubernetes/pkg/util/slice/`.

The lack of a simple []string filter function has spawned quite weird specialized help functions in K8s. Use of these functions (and later the go "slices" package) will improve code maintainability.

Usage example:
```go
func Example() {
	a := []string{
		"10.0.0.0", "FOO", "1000::1", "fd80::4%eth0", "BAR", "192.168.1.300",
		"172.12.0.1", "fc00::5000",
	}
	v := func(s string) bool { return net.ParseIP(s) == nil }
	if notIP := Filter(nil, a, v); len(notIP) > 0 {
		fmt.Println("Invalid", notIP)
	}
	fmt.Println(Filter(nil, a, utilnet.IsIPv6String))
	// Output:
	// Invalid [FOO fd80::4%eth0 BAR 192.168.1.300]
	// [1000::1 fc00::5000]
}
```

Part of https://github.com/kubernetes/kubernetes/issues/92369

There is an open issue to add a "slices" package to go: https://github.com/golang/go/issues/45955

The intention is to keep this implementation as close to the go slices proposal as possible. Ideal is if the proposal becomes a go standard package all applications need to do is to replcace "stringslices" -> "slices"

Related to https://github.com/kubernetes/utils/pull/202.

View generated doc:
```
cd /your/cloned/k8s.io/utils/
godoc -goroot .
firefox -new-window http://localhost:6060/pkg/k8s.io/utils/strings/
```

**Release note**:
```
NONE
```

### DIFF
--- a/strings/slices/slices.go
+++ b/strings/slices/slices.go
@@ -1,0 +1,82 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package slices defines various functions useful with slices of string type.
+// The goal is to be as close as possible to
+// https://github.com/golang/go/issues/45955. Ideal would be if we can just
+// replace "stringslices" if the "slices" package becomes standard.
+package slices
+
+// Equal reports whether two slices are equal: the same length and all
+// elements equal. If the lengths are different, Equal returns false.
+// Otherwise, the elements are compared in index order, and the
+// comparison stops at the first unequal pair.
+func Equal(s1, s2 []string) bool {
+	if len(s1) != len(s2) {
+		return false
+	}
+	for i, n := range s1 {
+		if n != s2[i] {
+			return false
+		}
+	}
+	return true
+}
+
+// Filter appends to d each element e of s for which keep(e) returns true.
+// It returns the modified d. d may be s[:0], in which case the kept
+// elements will be stored in the same slice.
+// if the slices overlap in some other way, the results are unspecified.
+// To create a new slice with the filtered results, pass nil for d.
+func Filter(d, s []string, keep func(string) bool) []string {
+	for _, n := range s {
+		if keep(n) {
+			d = append(d, n)
+		}
+	}
+	return d
+}
+
+// Contains reports whether v is present in s.
+func Contains(s []string, v string) bool {
+	return Index(s, v) >= 0
+}
+
+// Index returns the index of the first occurrence of v in s, or -1 if
+// not present.
+func Index(s []string, v string) int {
+	// "Contains" may be replaced with "Index(s, v) >= 0":
+	// https://github.com/golang/go/issues/45955#issuecomment-873377947
+	for i, n := range s {
+		if n == v {
+			return i
+		}
+	}
+	return -1
+}
+
+// Functions below are not in https://github.com/golang/go/issues/45955
+
+// Clone returns a new clone of s.
+func Clone(s []string) []string {
+	// https://github.com/go101/go101/wiki/There-is-not-a-perfect-way-to-clone-slices-in-Go
+	if s == nil {
+		return nil
+	}
+	c := make([]string, len(s))
+	copy(c, s)
+	return c
+}
--- a/strings/slices/slices_test.go
+++ b/strings/slices/slices_test.go
@@ -1,0 +1,183 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package slices
+
+import (
+	"fmt"
+	"net"
+	"reflect"
+	"regexp"
+	"strings"
+	"testing"
+
+	utilnet "k8s.io/utils/net"
+)
+
+func Example() {
+	a := []string{
+		"10.0.0.0", "FOO", "1000::1", "fd80::4%eth0", "BAR", "192.168.1.300",
+		"172.12.0.1", "fc00::5000",
+	}
+	v := func(s string) bool { return net.ParseIP(s) == nil }
+	if notIP := Filter(nil, a, v); len(notIP) > 0 {
+		fmt.Println("Invalid", notIP)
+	}
+	fmt.Println(Filter(nil, a, utilnet.IsIPv6String))
+	// Output:
+	// Invalid [FOO fd80::4%eth0 BAR 192.168.1.300]
+	// [1000::1 fc00::5000]
+}
+
+func Example_regexp() {
+	a := []string{
+		"10.0.0.0", "FOO", "1000::1", "fd80::4%eth0", "BAR", "192.168.1.300",
+		"172.12.0.1", "fc00::5000",
+	}
+	re := regexp.MustCompile("^[A-Z]+$")
+	fmt.Println(Filter(nil, a, re.MatchString))
+	// Output: [FOO BAR]
+}
+
+func ExampleFilter_empty() {
+	s := []string{"FOO", "", "", "BAR", "fd80::8888", ""}
+	fmt.Println(Filter(nil, s, func(s string) bool { return s != "" }))
+	// Output: [FOO BAR fd80::8888]
+}
+
+func TestFilter(t *testing.T) {
+	testCases := []struct {
+		validator func(string) bool
+		input     []string
+		res       []string
+	}{
+		// Filter all
+		{
+			regexp.MustCompile("zzz.*").MatchString,
+			[]string{"FOO", "1000::", "BAR", "fd80::8888"},
+			nil,
+		},
+		// Filter none
+		{
+			regexp.MustCompile(".*").MatchString,
+			[]string{"FOO", "1000::", "BAR", "fd80::8888"},
+			[]string{"FOO", "1000::", "BAR", "fd80::8888"},
+		},
+		// Filter some
+		{
+			func(s string) bool { return strings.Contains(s, ".") },
+			[]string{"10.0.0.0", "1000::", "8.8.8.8", "fd80::8888"},
+			[]string{"10.0.0.0", "8.8.8.8"},
+		},
+	}
+	for i, tc := range testCases {
+		res := Filter(nil, tc.input, tc.validator)
+		if !reflect.DeepEqual(tc.res, res) {
+			t.Errorf("TC %d: %v expected %v", i, res, tc.res)
+		}
+	}
+}
+
+func TestFilterInplace(t *testing.T) {
+	testCases := []struct {
+		validator  func(string) bool
+		input      []string
+		res        []string
+		inputAfter []string
+	}{
+		// Filter all
+		{
+			func(s string) bool { return s != "" },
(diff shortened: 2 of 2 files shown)


uablrek:
Made a comment in the go issue;
https://github.com/golang/go/issues/45955#issuecomment-873063129

uablrek:
When there seems to be a ongoing work to add a "slices" package in go, can we _please_ drop the idea of an own type;
```
type Slice []string
```
See https://github.com/kubernetes/utils/pull/202#discussion_r615058626

uablrek:
@thockin @rikatz @apelisse My comment on slack was a bit premature. After some feedback on the [go issue](https://github.com/golang/go/issues/45955) I see that they are right and I am wrong. I will drop the "Validator" and rewrite the filter with a function.

I will keep as close to the go proposal that I can but limit to []string.

And, for the record, I still think the own `Slice` type must be dropped.

uablrek:
I keep the "validator" package for now. I still think it's handy. Reviewers may have a different opinion :smiley:

strings/validator/validator.go
```diff
@@ -0,0 +1,117 @@
...
+package validator
+
+import (
+	"net"
+	"regexp"
+	"strings"
+)
+
+// Validator interface must be implemented by all validators
+type Validator interface {
```
thockin, line 29: Rather than a type, why not define this in terms of simple functions?

e.g. https://play.golang.org/p/XJG8QmDUEk5

strings/slices/slices.go
```diff
@@ -0,0 +1,52 @@
...
+// https://github.com/golang/go/issues/45955. Ideal would be if just
+// the import can be updated if the "slices" package becomes standard.
+package slices
+
+// func Filter[T any](d, s []T, keep func(T) bool) []T
+
+// Filter returns a new slice with the strings from the passed slice that are valid.
+
+// Filter appends to d each element e of s for which keep(e) returns true.
+// It returns the modified d. d may be s[:0], in which case the kept
```
aojea, line 28: hmm, I think that this is returning a pointer to the original `d` and mutating it,

strings/slices/slices.go
```diff
@@ -0,0 +1,52 @@
...
+	for _, n := range s {
+		if keep(n) {
+			d = append(d, n)
+		}
+	}
+	return d
+}
+
+// Contains reports whether v is present in s.
+func Contains(s []string, v string) bool {
```
aojea, line 45: if both are empty it should return true?

strings/slices/slices.go
```diff
@@ -0,0 +1,52 @@
...
+// Package slices defines various functions useful with slices of string type.
+// The goal is to be as close as possible to
+// https://github.com/golang/go/issues/45955. Ideal would be if just
+// the import can be updated if the "slices" package becomes standard.
+package slices
+
+// func Filter[T any](d, s []T, keep func(T) bool) []T
+
+// Filter returns a new slice with the strings from the passed slice that are valid.
+
```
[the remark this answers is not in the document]
uablrek: Sorry, another leftover. I was in a bit of a hurry to adapt to the possible go standard package

strings/slices/slices.go
```diff
@@ -0,0 +1,52 @@
...
+limitations under the License.
+*/
+
+// Package slices defines various functions useful with slices of string type.
+// The goal is to be as close as possible to
+// https://github.com/golang/go/issues/45955. Ideal would be if just
+// the import can be updated if the "slices" package becomes standard.
+package slices
+
+// func Filter[T any](d, s []T, keep func(T) bool) []T
```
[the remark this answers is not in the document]
uablrek: Yes, will be removed

strings/slices/slices.go
uablrek: The comment is taken from https://github.com/golang/go/issues/45955. Looked weird to me but it works as the "inplace" unit-test shows. But `d` must be returned since `s` is mutated but keeps it's original length.

strings/slices/slices.go
uablrek: Good question. Should it? I am unsure. Maybe a good question in https://github.com/golang/go/issues/45955

uablrek: https://github.com/golang/go/issues/45955#issuecomment-873348807

strings/validator/validator.go
uablrek: Looks better. I will change. Thanks for the comment. This is why I called on slack. I do not know all fine points of `go` yet and can take a wrong turn.

strings/slices/slices.go
```diff
@@ -0,0 +1,52 @@
...
+
+// Filter returns a new slice with the strings from the passed slice that are valid.
+
+// Filter appends to d each element e of s for which keep(e) returns true.
+// It returns the modified d. d may be s[:0], in which case the kept
+// elements will be stored in the same slice.
+// if the slices overlap in some other way, the results are unspecified.
+// To create a new slice with the filtered results, pass nil for d.
+func Filter(d, s []string, keep func(string) bool) []string {
+	if d == nil {
```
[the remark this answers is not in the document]
uablrek: Check removed

strings/slices/slices.go
aojea: my question is if `d` should be returned, hence modifying also `d`, or if a new slice should be returned

strings/slices/slices.go
aojea: :+1:

uablrek: Answer: false

uablrek:
#### Remove the validators?

I have removed `validator.Regexp` since regexp.MatchString can be used as-is. But now only validator.IP and validator.CIDR remains. While I think they belong in `k8s.io/utils` they should probably be in `k8s.io/utils/net`.

I still think the concept of providing complex validator functions as libs but perhaps not in `k8s.io/strings`. For instance a `k8s.io/file` validator may check if files exist.

Ideas on an appropriate structure?

strings/validator/validator.go
thockin: I really like this pattern, with first-class functions as the crux.  Rob Pike wrote this up in https://commandcenter.blogspot.com/2014/01/self-referential-functions-and-design.html

strings/slices/slices.go
thockin: If that's the semantic Go is choosing, we should stick to it.  It makes sense to me - if you want a copy, pass a copy.  Don't pessimize by default.

strings/slices/slices.go
```diff
@@ -0,0 +1,110 @@
...
+	s1clone := Clone(s1)
+	sort.Strings(s1clone)
+	s2clone := Clone(s2)
+	sort.Strings(s2clone)
+	return Equal(s1clone, s2clone)
+}
+
+// Remove returns a newly created []string that contains all
+// items from slice that are not equal to s.
+func Remove(slice []string, s string) []string {
```
thockin, line 104: Sticking with the apparently preferred Go semantics - perhaps this should do an in-place remove?  (https://zetcode.com/golang/filter-slice/)

strings/slices/slices_test.go
```diff
@@ -0,0 +1,281 @@
...
+
+	"k8s.io/utils/strings/validator"
+)
+
+func Example() {
+	a := []string{
+		"10.0.0.0", "FOO", "1000::1", "fd80::4%eth0", "BAR", "192.168.1.300",
+		"172.12.0.1", "fc00::5000",
+	}
+	v := validator.IP(0)
```
thockin, line 33: This package is so "leaf" that I don't think I would rely on validator in the tests.  I would just open-code some trivial logic here.

strings/validator/validator.go
```diff
@@ -0,0 +1,81 @@
...
+// Package validator defines various validator functions. The functions
+// are mainly intended to be used in strings/slices functions.
+package validator
+
+import (
+	"net"
+)
+
+// IP returns a validator for net.IP. IPVersion=0 matches both families
+func IP(IPVersion uint) func(string) bool {
```
thockin, line 26: we have k/utils/net `IsIPv6String()` and `IsIPv4String` (ditto CIDRString).  We could add `IsIPString` (which is just `net.ParseIP() != nil`) and then we don't need this validator package at all.

thockin (commented):
We have k/k/pkg/util/slice - will this obviate that whole package?

I think the validator package is too thin to justify.

strings/slices/slices.go
uablrek: The `Remove` function was taken from from "k8s.io/kubernetes/pkg/util/slice/". The original purpose of this issue was to replace that, but it was not very good in the first place so it is probably better to replace the few usages of Remove with a direct call to `filter`. The "empty" filter example shows how it can be done:
```go
func ExampleFilter_empty() {
	s := []string{"FOO", "", "", "BAR", "fd80::8888", ""}
	fmt.Println(Filter(nil, s, func(s string) bool { return s != "" }))
	// Output: [FOO BAR fd80::8888]
}
```
I remove the `Remove` function.

strings/validator/validator.go
uablrek: Me too. I did not know that you can use a `object.method()` in place of a function in `go`. I don't think there is something similar in Java or C++, probably why I did such an awkward first attempt.

uablrek:
Validators and `Remove()` removed.

strings/slices/slices_test.go
uablrek: You are absolutely right. I would not use them myself if I had not written them. Validators removed.

strings/validator/validator.go
uablrek: Agree. I re-wrote the Example with `IsIPv6String()` and it looks much better.

uablrek:
> We have k/k/pkg/util/slice - will this obviate that whole package?

Yes, that's the intention. It is in the description:

> The first goal is to replace k8s.io/kubernetes/pkg/util/slice/.

Fortunately it is used in very few places.

But IMO the scope has widened. Functions useful in K8s code are;

* Filter (or derivates such as "remove")
* Clone. See https://github.com/go101/go101/wiki/There-is-not-a-perfect-way-to-clone-slices-in-Go.
* SameElements. (please help me find a better name)

There is no `Clone()` in the `go` proposal. I asked about is but I think they took it as a joke (it isn't really) https://github.com/golang/go/issues/45955#issuecomment-873183426. There might not be a _perfect_ clone but the make+copy is good enough in 99.9% of the cases I think, so IMO it makes sense to have that as a function so nobody have to spend hours trying to find out how to clone a slice in the "correct" way (I did that).

thockin:
> There is no Clone() in the go proposal.

I expect it will come eventually.

> SameElements. (please help me find a better name)

Literature regarding sets would say `Equivalent` or `Equiv` ?  This is not quite a set (you can have dups) but it seems close enough to me.  Do we do this enough to warrant the function?  I mean, we have `sets.NewString()` which takes a variadic.  So I could say `sets.NewString(sl1...).Equal(sets.NewString(sl2...)` - that's more expensive, but is this super common?

Another word could be `Congruent`?

thockin (commented):
/hold

Remove the hold when naming is decided

/lgtm
/approve

uablrek:
> Do we do this enough to warrant the function?

Maybe not. I have only checked the `k8s.io/kubernetes/pkg/util/slice/` package and it is used in very few places and is does not even include a `Equiv` function. But the (unnecessary) `SortStrings` and the (misnamed) `CopyStrings` functions are used for that in a way that cries out for some code clean-up;

https://github.com/kubernetes/kubernetes/blob/1affd894cf5357559d64d6da7857b63ed760d3be/pkg/proxy/userspace/roundrobin.go#L266

https://github.com/kubernetes/kubernetes/blob/1affd894cf5357559d64d6da7857b63ed760d3be/pkg/proxy/userspace/roundrobin.go#L318-L327

But the usages seems limited to proxy/userspace which may be deprecaded(?).

To use the `sets` package was also proposed by @nikhita https://kubernetes.slack.com/archives/C09R23FHP/p1625039973123700?thread_ts=1625035208.123400&cid=C09R23FHP.

I propose we drop `Equiv` (SameElements) and if a similar function is seen in a review it should be taken as an indication that `sets` may be a better type to use than `[]string`?

I can easily refactor all places where `k8s.io/kubernetes/pkg/util/slice/` is used.

But there are other places where `filter` (and the other slices functions) are useful at other places. I have some in proxy/ipvs in mind.

thockin:
I'm ok with centralizing things like this, especially when the long-term answer is to trivially switch to a standard version.  We use `reflect.DeepEqual()` in a lot of places - I bet some of them could be replaced by this package

uablrek:
SameElements (Equiv) removed.

thockin (commented):
Thanks!

/lgtm
/approve

PR #209, opened 2021-07-06
LRU - seeded from github.com/golang/groupcache

dims:
Kubernetes has a lot of references to hashicorp's golang-lru package even after https://github.com/kubernetes/kubernetes/pull/95472 landed:
https://cs.k8s.io/?q=-lru%22&i=nope&files=&excludeFiles=&repos=kubernetes/kubernetes

Given that the golang-lru package came from:
https://github.com/golang/groupcache/tree/master/lru

We could just as well depend on that groupcache directly, however there is a catch, the hashicorp's golang-lru impl is thread safe, so essentially in this PR, we pick up LRU.go from groupcache and add a lock around it.

This will help us get rid of the last of the `Mozilla Public License` dependency from kubernetes LICENSES 

### DIFF
--- a/lru/lru.go
+++ b/lru/lru.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package lru
+
+import (
+	"sync"
+
+	groupcache "k8s.io/utils/internal/third_party/forked/golang/golang-lru"
+)
+
+type Key = groupcache.Key
+
+// Cache is a thread-safe fixed size LRU cache.
+type Cache struct {
+	cache *groupcache.Cache
+	lock  sync.RWMutex
+}
+
+// New creates an LRU of the given size.
+func New(size int) *Cache {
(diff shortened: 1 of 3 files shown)


thockin (commented):
1) I don't think we want to just copy the code in.  I don't want to maintain this forever.

2) You used RLock in write paths :)

Can we instead write this as a pure wrapper ?  E.g https://play.golang.org/p/tSTWKCDhJmt (untested).  We'll need to declare it a dep, so COULD be talked into forking this.  

If we are going to fork it, I'd fork it to internal/third_party and re-expose it similar to how net functions are being done.

dims:
> You used RLock in write paths :)

LOL - copy-pasta!

- the problem with pure wrapper (was my first instinct too), is that this repo is not maintained AND it does not have any tagged releases, which is why hashicorp folks forked it.

I totally forgot about the `third_party/forked/golang`, will do that instead with the wrapper around it.

thanks!

thockin:
How about moving it one level further into internal : https://github.com/kubernetes/utils/tree/master/internal/third_party/forked

dims:
> How about moving it one level further into internal : https://github.com/kubernetes/utils/tree/master/internal/third_party/forked

DONE!

thockin (commented):
Thanks!

/lgtm
/approve

lru/lru.go
```diff
@@ -0,0 +1,79 @@
...
+func (c *Cache) Add(key Key, value interface{}) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	c.cache.Add(key, value)
+}
+
+// Get looks up a key's value from the cache.
+func (c *Cache) Get(key Key) (value interface{}, ok bool) {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
```
liggitt, line 49: note for posterity, on an LRU, Get is a write path :)

fix in https://github.com/kubernetes/utils/pull/216

dims: 🙏🏾 thanks @liggitt !

PR #210, opened 2021-07-17
Ignore 'wait: no child processes' error when calling mount/umount

shyamjvs:
Ref https://github.com/kubernetes/kubernetes/issues/103753

I've only fixed the exec commands that are part of Mount() and Unmount() functions and that too in the linux mount helper. Not touching others, since I'm not sure about the implications. Let me know if I should.

/kind bug
/sig storage

### DIFF
--- a/mount/mount_linux.go
+++ b/mount/mount_linux.go
@@ -43,6 +43,8 @@ const (
 	fsckErrorsCorrected = 1
 	// 'fsck' found errors but exited without correcting them
 	fsckErrorsUncorrected = 4
+	// Error thrown by exec cmd.Run() when process spawned by cmd.Start() completes before cmd.Wait() is called (see - k/k issue #103753)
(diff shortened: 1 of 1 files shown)


mount/mount_linux.go
```diff
@@ -147,6 +149,10 @@ func (mounter *Mounter) doMount(mounterPath string, mountCmd string, source stri
 	command := exec.Command(mountCmd, mountArgs...)
 	output, err := command.CombinedOutput()
 	if err != nil {
+		if err.Error() == errNoChildProcesses {
```
liggitt, line 152: this package is unused now, see https://github.com/kubernetes/utils/blob/master/mount/README.md for the currently used location of this code

shyamjvs: Aah, I'll close this PR and open a new one there. Thanks!

PR #211, opened 2021-07-21
Introduce WithTicker clock

### LINKED ISSUE #196: Issues with Clock

MikeSpreitzer:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> 
/kind feature
> /kind flake

See linked issues. TL;DR adds a missing method to allow safer usage and to bridge the gap between two copies of the clock package so that the one in apimachinery can be deleted.

Fixes #196

Updates kubernetes/kubernetes#94738

This is a copy of #197 and is entirely due to @ash2k .  I am making a duplicate PR because the original one seems to be stuck due to an un-fixable testing misconfiguration.

**Release note**:
```
NONE
```

### DIFF
--- a/clock/testing/fake_clock.go
+++ b/clock/testing/fake_clock.go
@@ -25,8 +25,8 @@ import (
 
 var (
 	_ = clock.PassiveClock(&FakePassiveClock{})
-	_ = clock.Clock(&FakeClock{})
-	_ = clock.Clock(&IntervalClock{})
+	_ = clock.WithTicker(&FakeClock{})
+	_ = clock.WithTicker(&IntervalClock{})
 )
 
 // FakePassiveClock implements PassiveClock, but returns an arbitrary time.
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
Looks like #197 is going to make progress, closing this duplicate.

clock/testing/fake_clock.go
```diff
@@ -231,6 +249,12 @@ func (*IntervalClock) Tick(d time.Duration) <-chan time.Time {
 	panic("IntervalClock doesn't implement Tick")
 }
 
+// NewTicker is currently unimplemented, will panic.
+// TODO: make interval clock use FakeClock so this can be implemented.
+func (*IntervalClock) NewTicker(d time.Duration) clock.Ticker {
```
apelisse, line 254: Why do we need this function again? If we didn't, IntervalClock wouldn't implement the "WithTicker" interface, which means that it would fail at compile time. By adding it, it now implements WithTicker, but will fail at runtime?

PR #213, opened 2021-07-22
Introduced SimpleIntervalClock

MikeSpreitzer:
This is a more honest alternative to IntervalClock.

**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature
> /kind flake

This PR introduces clock/testing.SimpleIntervalClock, which is a better alternative to clock/testing.IntervalClock because the latter has additional methods that just panic.  Using the former makes for checking at compile time while the latter does the corresponding checking at runtime.

See https://github.com/kubernetes/utils/pull/197#discussion_r674361869 for the start of the discussion leading here.

**Release note**:
```
NONE
```

### DIFF
--- a/clock/testing/fake_clock.go
+++ b/clock/testing/fake_clock.go
@@ -196,7 +196,9 @@ func (f *FakeClock) Sleep(d time.Duration) {
 	f.Step(d)
 }
 
-// IntervalClock implements clock.Clock, but each invocation of Now steps the clock forward the specified duration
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
/cc @apelisse 
@ash2k 
@lavalamp

MikeSpreitzer:
@derekwaynecarr : the tests in client-go/tools/record/ are the only uses of IntervalClock that I found.

PR #219, opened 2021-08-30
keymutex: simplify unlocking

pohly:
**What type of PR is this?**

/kind api-change

The common pattern is:
```
  mutex.LockKey(key)
  defer mutex.UnlockKey(key)
```
However, linters point out the the error return of UnlockKey is not checked.
It is possible and okay to suppress that because it is known that the right key
is getting passed into that function:
```
  mutex.LockKey(key)
  defer func() {
     _ = mutex.UnlockKey(key)
  }()
```

But this is rather ugly. By returning an unlock function, locking and unlocking
can be done in a single line with less code:
```
  defer mutex.LockKey(key)()
```

Strictly speaking, this is an API break because it changes the LockKey signature. But in practice, existing code continues to compile and work as intended.

**Release note**:
```
simpler keymutex: locking and unlocking in a single line
```

### DIFF
--- a/keymutex/hashed.go
+++ b/keymutex/hashed.go
@@ -41,8 +41,11 @@ type hashedKeyMutex struct {
 }
 
 // Acquires a lock associated with the specified ID.
-func (km *hashedKeyMutex) LockKey(id string) {
+func (km *hashedKeyMutex) LockKey(id string) func() {
 	km.mutexes[km.hash(id)%uint32(len(km.mutexes))].Lock()
+	return func() {
+		km.UnlockKey(id)
+	}
 }
 
 // Releases the lock associated with the specified ID.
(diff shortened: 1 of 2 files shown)


liggitt:
/remove-kind api-change

this label is intended for REST / config / cli API changes, not go API changes

because this repo does not maintain release branches and tags, it has a strict forward compatibility requirement on the exposed go API surface (checked by the apidiff presubmit)

liggitt:
adding a new function to the interface would also not be API-compatible

it's not clear to me that the _ = ... workaround is incorrect. any time an unlock is separated in from from a lock, something else could have unlocked the id, and depending on the implementation, removed knowledge of the id from the locker.

pohly:
> adding a new function to the interface would also not be API-compatible

True. Then it's really not worth it for a minor convenience enhancement.

> it's not clear to me that the _ = ... workaround is incorrect.

It isn't incorrect, just boilerplate code that I was trying to avoid.

PR #220, opened 2021-09-15
fix: read small files or files that show zero size once as a large file.

manugupt1:
/proc/mounts zero as disk usage size

```
ubuntu@ip-172-31-49-2:~/go/src/k8s.io/kubernetes/vendor/k8s.io/utils/io$
du -h /proc/mounts
0	/proc/mounts
ubuntu@ip-172-31-49-2:~/go/src/k8s.io/kubernetes/vendor/k8s.io/utils/io$
```

As a result of which between multiple read calls, the content of
/proc/mounts might change, this change introduceces reading 0 or small
files to read the entire file in 1 go.

**What type of PR is this?**
/kind bug

Reduces the occurrence of https://github.com/kubernetes/kubernetes/issues/104976

When multiple pods with many subpaths are trying to mount / umount the subPaths, then the /proc/mounts keeps on consistently changing. To avoid this scenario, we should read the entire `/proc/mounts` in one go and also not do consistentRead check.

This PR introduces the change in the consistentRead to demonstrate the change and gather feedback from the community.

Fixes #  https://github.com/kubernetes/kubernetes/issues/104976

* This is to gather feedback and we can change the code specific to subPath itself.
* Should we introduce unix specific build tags in the utils library.
* Should we extract this out to a separate function itself in subPath rather than rely on this library
* Tested this change with steps given in: https://github.com/kubernetes/kubernetes/issues/104976 for a 1.5 hours; at 150 pods per 5 min. I have not seen any issue so far.

**Release note**:
```
N/A
```

/sig storage

### DIFF
--- a/io/read.go
+++ b/io/read.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"os"
 )
 
 // ErrLimitReached means that the read limit is reached.
@@ -41,6 +42,23 @@ func ConsistentRead(filename string, attempts int) ([]byte, error) {
 // introduces a sync callback that can be used by the tests to mutate the file
 // from which the test data is being read
 func consistentReadSync(filename string, attempts int, sync func(int)) ([]byte, error) {
+
+	// get the size of the file.
+	size, err := getFileSize(filename)
+	if err != nil {
+		return nil, err
+	}
+
+	// if the size of the file is less than 512 bytes, assume that reading from
+	// /proc or /sys like cases for Linux.
+	if size < 512 {
+		if contents, err := consistentReadFileNoStat(filename); err == nil {
+			return contents, nil
+		}
+		// fall back to consistent read because this is now a utility and an error here can break
+		// expected behavior.
+	}
+
 	oldContent, err := ioutil.ReadFile(filename)
 	if err != nil {
 		return nil, err
@@ -60,6 +78,52 @@ func consistentReadSync(filename string, attempts int, sync func(int)) ([]byte, 
 		oldContent = newContent
 	}
 	return nil, InconsistentReadError{filename, attempts}
+}
+
+// get filesize returns the filesize
+func getFileSize(filename string) (int, error) {
+	f, err := os.Open(filename)
+	if err != nil {
+		return 0, err
+	}
+
+	var size int
+	// Calling stat should be okay as it supports both Windows and Linux
+	// Ref (unix and posix compliant): https://cs.opensource.google/go/go/+/refs/tags/go1.17.1:src/os/stat_unix.go;l=16
+	// Ref (windows): https://cs.opensource.google/go/go/+/refs/tags/go1.17.1:src/os/stat_windows.go
+	if info, err := f.Stat(); err == nil {
+		size64 := info.Size()
+		if int64(int(size64)) == size64 {
+			size = int(size64)
+		}
+	}
+	size++ // one byte for final read at EOF
+	return size, nil
+}
+
+// consistentReadFileNoStat is inspired from prometheus code base for /proc and /sys
+func consistentReadFileNoStat(filename string) ([]byte, error) {
+	f, err := os.Open(filename)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+
+	const maxBufferSize = 1024 * 512
+	var contents = make([]byte, maxBufferSize)
+
+	// use f.Read as it points to File.Read which will directly call os dependent functions
+	// For Linux: https://cs.opensource.google/go/go/+/master:src/syscall/syscall_unix.go;l=188;bpv=1;bpt=1?q=syscall.Read&ss=go%2Fgo
+	// Also, do not prefer ReadAtMost as it internally results in a loop for read depending on the buffer capacity.
+	// This will result in multiple syscall.Read
+	n, err := f.Read(contents)
+	if err != nil {
+		// not returning contents to ensure that the bytes are not used up the stack as they occupy memory.
+		return []byte{}, err
+	}
+	// slice the byte before returning to the x number of bytes so that the maximum buffer is not reached
+	return contents[:n], nil
 }
 
 // InconsistentReadError is returned from ConsistentRead when it cannot get


dims:
@manugupt1 can you please add a test (would it make sense to add one?)?

/assign @thockin

manugupt1:
@dims I am not even sure right now if this is the right place to make the change; but this PR demonstrates the problem and of one way to fix it.

1. For example, if we go this route, the attempts do not make sense for some scenarios and not for others. This leads to an inconsistent API for consumers.
2. The other option could be this change could be specific to subpath on linux only rather than being OS agnostic. 

I would like to get feedback before writing tests, but if you still prefer tests, I can add a test and fix those.

dims:
@manugupt1 where is the `ConsistentRead` called from?

dims:
there's some fun stuff going on in https://github.com/golang/go/blob/master/src/os/file.go#L685-L690 as well.

manugupt1:
@dims Yep :). There is a lot of fun stuff. We need to be aware of the fact that ioutil changed its code between 1.16 and 1.17 I believe and that is yet another thing that I tried to avoid using;  I considered that while writing this PR but double checking won't hurt.

Right! the subpath mount hits the code path that you pointed out.. I am right now conflicted; what will be a good place to place this particular piece of code at. The more I think, the more I am inclined towards having it here: https://github.com/kubernetes/utils/blob/d34e5cb4466e533a699c4cdbd76f3a8220b21d54/mount/mount_linux.go#L457 as that utils is os specfic itself.

Let me know what do you think?

dims:
let me chew on this a bit. 

/unassign @thockin

dims:
@manugupt1 can things be way simpler ... say something like this? will this work for your case?

```
diff --git a/io/read.go b/io/read.go
index f0af3c8..fd48ce7 100644
--- a/io/read.go
+++ b/io/read.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"os"
 )
 
 // ErrLimitReached means that the read limit is reached.
@@ -49,7 +50,7 @@ func consistentReadSync(filename string, attempts int, sync func(int)) ([]byte,
 		if sync != nil {
 			sync(i)
 		}
-		newContent, err := ioutil.ReadFile(filename)
+		newContent, err := readFile(filename)
 		if err != nil {
 			return nil, err
 		}
@@ -62,6 +63,30 @@ func consistentReadSync(filename string, attempts int, sync func(int)) ([]byte,
 	return nil, InconsistentReadError{filename, attempts}
 }
 
+func readFile(path string) ([]byte, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+	fi, err := f.Stat()
+	if err != nil {
+		return nil, err
+	}
+	size := fi.Size()
+	if size < 512 {
+		var contents = make([]byte, 512)
+		n, err := f.Read(contents)
+		if err != nil {
+			return nil, err
+		} else {
+			return contents[:n], nil
+		}
+	} else {
+		return ioutil.ReadAll(f)
+	}
+}
+
 // InconsistentReadError is returned from ConsistentRead when it cannot get
 // a consistent read in given nr. of attempts. Caller should retry, kernel is
 // probably under heavy mount/unmount load.
```

manugupt1:
Thank you for reviewing it so deeply. I have a few things that we should.

Hmm! I think this breaks:

https://github.com/kubernetes/utils/blob/master/io/read.go#L48-L61. Someone in sig-storage should have more information on the past as I am reading it for the first time.

The below buffer will lead to multiple syscall.Read as the /proc/mounts has more content than 512 bytes in reality but displays 0 size while using du.
```
+		var contents = make([]byte, 512)
```

I checked this:

https://github.com/kubernetes/kubernetes/search?q=ConsistentRead (there are only two calls from here); so maybe we can just get rid of the retries (but I would like to still understand why this function was introduced in the first place).

manugupt1:
This PR also affects: https://github.com/kubernetes/utils/pull/220 

cc @jsafrane

dims:
how about this?

```
diff --git a/io/read.go b/io/read.go
index f0af3c8..876a7b4 100644
--- a/io/read.go
+++ b/io/read.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"os"
 )
 
 // ErrLimitReached means that the read limit is reached.
@@ -49,7 +50,7 @@ func consistentReadSync(filename string, attempts int, sync func(int)) ([]byte,
 		if sync != nil {
 			sync(i)
 		}
-		newContent, err := ioutil.ReadFile(filename)
+		newContent, err := readFile(filename)
 		if err != nil {
 			return nil, err
 		}
@@ -62,6 +63,28 @@ func consistentReadSync(filename string, attempts int, sync func(int)) ([]byte,
 	return nil, InconsistentReadError{filename, attempts}
 }
 
+func readFile(path string) ([]byte, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+	fi, err := f.Stat()
+	if err != nil {
+		return nil, err
+	}
+	size := fi.Size()
+	if size < 512 {
+		const maxBufferSize = 1024 * 512
+		var contents = make([]byte, maxBufferSize)
+		n, err := f.Read(contents)
+		if err == nil {
+			return contents[:n], nil
+		}
+	}
+	return ioutil.ReadAll(f)
+}
+
 // InconsistentReadError is returned from ConsistentRead when it cannot get
 // a consistent read in given nr. of attempts. Caller should retry, kernel is
 // probably under heavy mount/unmount load.

```

manugupt1:
@dims yep! this works. 

The key question though is that do we want attempts in this function or not as shown in the issue that I linked @jsafrane  with. I do not want to inadvertently break other drivers. I had to write this convoluted code for this reason itself.

dims:
@manugupt1 i'd vote to keep the attempts as is. if you can rework this PR with the patch you liked and figure out how to add a test case. then we can get @thockin and @jsafrane to review.

manugupt1:
@dims thanks for the help! I will submit a PR in the next couple of days.

dims:
@manugupt1 one more tip, you can do a `f.Seek(0, io.SeekEnd)` to get the length of the file to compute the length of the buffer, then do a SeekBegin to start from the top. Then we don't need magic number i think. give it a shot and see if that helps.

manugupt1:
@dims that is a really cool suggestion; I think it will work.
Reason being this syscall: https://man7.org/linux/man-pages/man2/lseek.2.html
which go calls via the seek call. https://cs.opensource.google/go/go/+/master:src/internal/poll/fd_unix.go;l=528;drc=master

I will try and let you know.

jsafrane:
`read()` on /proc/mounts does not return the whole file, even when it fits the provided buffer. On 4.18.0 (RHEL8) and 5.13.12 (Fedora 34), `read()` always returns +/- one page:

```
$ strace cat /proc/mounts
...
read(3, "proc /proc proc rw,nosuid,nodev,"..., 131072) = 4085
...
read(3, "xxx /tmp/11 tmpfs rw,seclabel,re"..., 131072) = 4080
...
read(3, "xxx /tmp/91 tmpfs rw,seclabel,re"..., 131072) = 511
```

`cat` provided buffer with 131072 bytes, but it got <4096 bytes in a single read. `consistentReadFileNoStat` proposed here does the same, `Read()` returns only the first ~4000 bytes.

manugupt1:
Hmm! that is a very good catch.

I wrote this small program and tried on Ubuntu (5.11) and AL2 (4.14). I got a single read call in my strace. Seems like this call is dependent on the page size.  I am not an expert in reading strace; so I will attach it. 
[strace_ubuntu.txt](https://github.com/kubernetes/utils/files/7178814/strace_ubuntu.txt)
[strace_al2.txt](https://github.com/kubernetes/utils/files/7178815/strace_al2.txt)


I have this set up `/sys/kernel/mm/transparent_hugepage/hpage_pmd_size` as `2097152` in both ubuntu and AL2 and my suspicion is that this is leading to the single read. I am curious to know what does your output look like as well.

```
ubuntu@ip-172-31-49-2:~/tmp$ uname -r
5.11.0-1017-aws
```
```
ec2-user@ip-172-31-17-14:~/tmp
▶ uname -r
4.14.243-185.433.amzn2.x86_64
```


```
package main

import (
	"os"
)

func main() {
	filename := "/proc/mounts"
	f, err := os.Open(filename)
	if err != nil {
		os.Exit(2)
	}
	defer f.Close()


	const maxBufferSize = 1024 * 512
	var contents = make([]byte, maxBufferSize)
	_, err = f.Read(contents)
	if err != nil {
		os.Exit(1)
	}

}
```

manugupt1:
@jsafrane Sorry! I did not understand the last bit. I got it now; so read only returns  upto a page and not everything is returned.

I tried util-linux as well; but that also did not work.

```
[ec2-user@ip-192-168-30-44 ~]$ strace findmnt > /dev/null
execve("/usr/bin/findmnt", ["findmnt"], 0x7ffef7a947a0 /* 21 vars */) = 0
brk(NULL)                               = 0x23b1000
```

```
read(3, "t-storage-1/shane-test-55 /var/l"..., 1024) = 1024
read(3, "/var/lib/kubelet/pods/f1cd7449-b"..., 1024) = 1024
read(3, "49-bc43-45a4-871f-364fdea4800b/v"..., 1024) = 1024
read(3, "0b/volume-subpaths/persistent-st"..., 1024) = 1024
read(3, "t-storage-1/app/27 rw,noatime sh"..., 1024) = 1024
read(3, "shared:1 - xfs /dev/nvme0n1p1 rw"..., 1024) = 1024
read(3, "ttr2,inode64,logbufs=8,logbsize="..., 1024) = 1024
read(3, "32k,noquota\n1423 41 259:1 /var/l"..., 1024) = 1024
read(3, "lib/kubelet/pods/7d26106f-77a7-4"..., 1024) = 1024
read(3, "7-4e1a-8ef1-75bb74be93a9/volumes"..., 1024) = 1024
read(3, "umes/kubernetes.io~empty-dir/per"..., 1024) = 1024
read(3, "/persistent-storage-1/shane-test"..., 1024) = 1024
read(3, "test-48 /var/lib/kubelet/pods/7d"..., 1024) = 1024
read(3, "s/7d26106f-77a7-4e1a-8ef1-75bb74"..., 1024) = 1024
read(3, "bb74be93a9/volume-subpaths/persi"..., 1024) = 956
read(3, "", 1024)                       = 0
```

manugupt1:
Closing this based on last comment.

PR #221, opened 2021-09-23
Implement AfterFunc for Clock, FakeClock and IntervalClock

MadhavJivrajani:
**What type of PR is this?**
/kind cleanup

This PR implements the `AfterFunc` method for the clocks in `k8s.io/utils/clock`. Having this implemented, `k8s.io/utils/clock` should be able to be a functional replacement of `k8s.io/apimachinery/pkg/util/clock` and we'll be one step closer to solving the divergent clock packages issue https://github.com/kubernetes/kubernetes/issues/94738.

Fixes https://github.com/kubernetes/kubernetes/issues/105173

None

**Release note**:
```
This PR introduces a new interface for the clock package: WithDelayedExecution - this extends the Clock interface by further including the AfterFunc method.
```

### DIFF
--- a/clock/testing/fake_clock.go
+++ b/clock/testing/fake_clock.go
@@ -49,6 +49,7 @@ type fakeClockWaiter struct {
 	skipIfBlocked bool
 	destChan      chan time.Time
 	fired         bool
+	afterFunc     func()
 }
 
 // NewFakePassiveClock returns a new FakePassiveClock.
@@ -116,6 +117,25 @@ func (f *FakeClock) NewTimer(d time.Duration) clock.Timer {
 	return timer
 }
 
+// AfterFunc is the Fake version of time.AfterFunc(d, cb).
+func (f *FakeClock) AfterFunc(d time.Duration, cb func()) clock.Timer {
+	f.lock.Lock()
+	defer f.lock.Unlock()
+	stopTime := f.time.Add(d)
+	ch := make(chan time.Time, 1) // Don't block!
+
+	timer := &fakeTimer{
+		fakeClock: f,
+		waiter: fakeClockWaiter{
+			targetTime: stopTime,
+			destChan:   ch,
+			afterFunc:  cb,
+		},
+	}
+	f.waiters = append(f.waiters, &timer.waiter)
+	return timer
+}
+
 // Tick constructs a fake ticker, akin to time.Tick
 func (f *FakeClock) Tick(d time.Duration) <-chan time.Time {
 	if d <= 0 {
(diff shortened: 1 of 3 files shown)


clock/testing/fake_clock.go
```diff
@@ -116,6 +117,25 @@ func (f *FakeClock) NewTimer(d time.Duration) clock.Timer {
 	return timer
 }
 
+// AfterFunc is the Fake version of time.AfterFunc(d, cb).
```
wojtek-t, line 120: Add a unit test for AfterFunc:

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/clock/clock_test.go#L150

wojtek-t (commented):
One comment

clock/testing/fake_clock.go
MadhavJivrajani: Will do, thanks!

@wojtek-t do you know how the test failure can be rectified? The failure is occurring because the `Clock` interface is no longer compatible with the previous version. 

cc @liggitt

wojtek-t:
Actually - the API diff is a good question.

I can see a motivation for not allowing extending interfaces.
In such case - the only thing I can imagine is to create a new interface that is extending clock
(similarly as we have WithTicker: https://github.com/kubernetes/utils/blob/master/clock/clock.go#L50)

[all the rest of the code remains as is]

/assign @dims

MadhavJivrajani:
/hold
until apidiff can be figured out

jonyhy96:
@MadhavJivrajani  hi~ sorry to bother, i'm woring on the same thing because i didn't found related issue(forgot to check prs),my implements is quit the same #223 to you but the testFunc maybe more clear. do u mind to merge my commit in your pr if u accept the testFunc? thx.

liggitt:
> In such case - the only thing I can imagine is to create a new interface that is extending clock
> (similarly as we have WithTicker: https://github.com/kubernetes/utils/blob/master/clock/clock.go#L50)

Something like that sounds like what I would expect.

MadhavJivrajani:
> @MadhavJivrajani hi~ sorry to bother, i was working on the same thing because i didn't found related issue(forgot to check prs) in the first place. My implements is quit the same #223 as yours but the testFunc maybe more clear. do u mind to merge my commit in your pr if u accept the testFunc? thx.

Hi @jonyhy96 - no issues. Thanks for creating the PR. Atleast imo the tests in this commit are more appropriate because firstly they align with the existing tests upstream, while the tests you've written look clearer - I'm not sure its complete because we need to check if trigger happens for the right intervals in order and not just based off a count (`called` in your case).

MadhavJivrajani:
@wojtek-t @liggitt I've made the changes. PTAL :)

wojtek-t:
/lgtm

@dims  - can you help with approving this?

MadhavJivrajani:
Will vendor in k/k after this is merged and migrate the last bit of code :)

dims:
no change in API, has test cases. @MadhavJivrajani FTW! this looks good to me.

/approve
/lgtm

PR #223, opened 2021-09-28
feat: make clock support AfterFunc

### LINKED ISSUE #222: make clock support AfterFunc()
[kubernetes/kubernetes](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/nodelifecycle/scheduler/timed_workers.go#L25) still using `k8s.io/apimachinery/pkg/util/clock` because of clock does't implements AfterFunc yet.

We could implement `AfterFunc` in clock and replace `k8s.io/apimachinery/pkg/util/clock` with `k8s.io/utils/clock` in main repository.

jonyhy96:
**What type of PR is this?**

/kind feature

Support AfterFunc in clock for third-party use.

Fixes #222

### DIFF
--- a/clock/clock.go
+++ b/clock/clock.go
@@ -34,6 +34,8 @@ type Clock interface {
 	// This method does not allow to free/GC the backing timer before it fires. Use
 	// NewTimer instead.
 	After(d time.Duration) <-chan time.Time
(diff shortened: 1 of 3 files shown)


jonyhy96:
PTAL @deads2k

jonyhy96:
dulp #221

PR #224, opened 2021-10-11
Prepare clock package to be compatible replacement of apimachinery/util/clock

MadhavJivrajani:
**What type of PR is this?**
/kind cleanup
Commit 1:
The reason for doing this is that `utils/clock` is supposed
to be a full replacement for `k8s.io/apimachinery/pkg/util/clock`.

Currently, `WithDelayedExecution` extends Clock - in order to deprecate
apimachinery/clock, `WithDelayedExecution` would be missing the NewTicker
method if gradual deprecation via type aliasing was to be done.

By extending `WithTicker` instead of `Clock`, we solve this issue.

Commit 2:
Eventhough the implementation is just a panic,
re-implementing the `NewTicker` method helps
in migration of the apimachinery/util/clock pkg
to the utils/clock pkg in terms of having identical
clients for `IntervalClock` in both packages as
well as the `IntervalClock` implementing the
`WithTickerAndDelayedExcution` interface.

Related to https://github.com/kubernetes/kubernetes/issues/94738

The next step after this would be to implement the `Tick` method in apimachinery clock, type aliasing the interfaces and marking the package as deprecated.

**Release note**:
```
NONE
```
/assign @liggitt @dims 
/cc @wojtek-t

### DIFF
--- a/clock/clock.go
+++ b/clock/clock.go
@@ -63,6 +63,16 @@ type WithDelayedExecution interface {
 	AfterFunc(d time.Duration, f func()) Timer
 }
 
+// WithTickerAndDelayedExecution allows for injecting fake or real clocks
+// into code that needs Ticker and AfterFunc functionality
+type WithTickerAndDelayedExecution interface {
+	WithTicker
+	// AfterFunc executes f in its own goroutine after waiting
+	// for d duration and returns a Timer whose channel can be
+	// closed by calling Stop() on the Timer.
+	AfterFunc(d time.Duration, f func()) Timer
+}
+
 // Ticker defines the Ticker interface.
 type Ticker interface {
 	C() <-chan time.Time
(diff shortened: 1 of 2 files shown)


MadhavJivrajani:
Yep :)

wojtek-t:
/lgtm

@liggitt @dims - can you take a look?

MadhavJivrajani:
/hold
(would like to double check once again post approval)

liggitt:
can you open a WIP PR against kubernetes/kubernetes that bumps k8s.io/utils to this commit and ensures we can completely switch over to this package and get green CI on that PR?

you can do that in kubernetes/kubernetes like this:
```
hack/pin-dependency.sh k8s.io/utils=github.com/MadhavJivrajani/utils rectify-interface-extension
hack/update-vendor.sh
```

MadhavJivrajani:
@liggitt @wojtek-t fyi - https://github.com/kubernetes/kubernetes/pull/106850 :)

wojtek-t:
That LGTM

MadhavJivrajani:
Thank you, both!
Removing hold now seeing that the CI in the test PR is green :)
/hold cancel

PR #225, opened 2021-10-31
fix panic error cause by index out of range when people use multi thr…

Zheaoli:
…eading reading

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature
> /kind flake

Fixes https://github.com/kubernetes/kubernetes/issues/106003

**Release note**:
```

```

### DIFF
--- a/buffer/ring_growing.go
+++ b/buffer/ring_growing.go
@@ -28,10 +28,10 @@ type RingGrowing struct {
 }
 
 // NewRingGrowing constructs a new RingGrowing instance with provided parameters.
-func NewRingGrowing(initialSize int32) *RingGrowing {
+func NewRingGrowing(initialSize int) *RingGrowing {
 	return &RingGrowing{
 		data: make([]interface{}, initialSize),
-		n:    initialSize,
+		n:    int32(initialSize),
 	}
 }
 
--- a/buffer/ring_growing.go
+++ b/buffer/ring_growing.go
@@ -16,17 +16,19 @@ limitations under the License.
 
 package buffer
 
+import "sync/atomic"
+
 // RingGrowing is a growing ring buffer.
 // Not thread safe.
 type RingGrowing struct {
 	data     []interface{}
-	n        int // Size of Data
-	beg      int // First available element
-	readable int // Number of data items available
+	n        int32 // Size of Data
+	beg      int32 // First available element
+	readable int32 // Number of data items available
 }
 
 // NewRingGrowing constructs a new RingGrowing instance with provided parameters.
-func NewRingGrowing(initialSize int) *RingGrowing {
+func NewRingGrowing(initialSize int32) *RingGrowing {
 	return &RingGrowing{
 		data: make([]interface{}, initialSize),
 		n:    initialSize,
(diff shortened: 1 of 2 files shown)


dims:
@Zheaoli will need a test to ensure we don't regress? (looks like this just touches up the existing test to make it compile)

Zheaoli:
@dims I think we need extra e2e to test this. the index out range error is not reproduced every time.

buffer/ring_growing.go
```diff
@@ -16,57 +16,72 @@ limitations under the License.
...
 
 // ReadOne reads (consumes) first item from the buffer if it is available, otherwise returns false.
 func (r *RingGrowing) ReadOne() (data interface{}, ok bool) {
-	if r.readable == 0 {
+	oldReadable := atomic.LoadInt32(&r.readable)
+	oldN := atomic.LoadInt32(&r.n)
+	if oldReadable == 0 {
+		return nil, false
+	}
+	if !atomic.CompareAndSwapInt32(&r.readable, oldReadable, oldReadable-1) {
```
cyfdecyf, line 45: Suppose `r.readable` is 100, it's possible that two goroutines are trying to execute this `compareAndSwapInt32` at the same time and only one gorouting will succeed. This means even we will return `nil` even if `r.readable > 0`.

I haven't write go code for a long time, but can we use `atomic.Xadd` here which should be much easier.

cyfdecyf, line 45: Suppose `r.readable` is 100, it's possible that two goroutines are trying to execute `atomic.compareAndSwapInt32` at the same time and only one gorouting succees. This means on gorouting will return `nil` even if `r.readable > 0`.

buffer/ring_growing.go
```diff
@@ -16,57 +16,72 @@ limitations under the License.
...
 		}
-		r.beg = 0
+		atomic.StoreInt32(&r.n, newN)
 		r.data = newData
-		r.n = newN
+		atomic.StoreInt32(&r.beg, 0)
 	}
 	r.data[(r.readable+r.beg)%r.n] = data
-	r.readable++
+	atomic.StoreInt32(&r.readable, oldReadable+1)
```
cyfdecyf, line 86: I'm not sure about the use case of `RingGrowing`, but for general cases (multi-producer and multi-consumer) this is still not thread safe.

It's possible that two goroutines execute line `66-68` at the same time thus get the same `oldReadable`. They may try to execute `atomic.StoreInt32` with exactly the same `oldReadable` which means we lost one item that should stay in the ring.

Lock-free unbounded queue is not so easy to get right. I haven't been doing this style of programming for a long time. This is the [last paper](https://www.cs.tau.ac.il/~mad/publications/ppopp2013-x86queues.pdf) I read about concurrent queues.

buffer/ring_growing.go
Zheaoli: Yes It will change the default behavior

buffer/ring_growing.go
Zheaoli: Yes, it's also not thread safe enough in this circumstance.

I think should we just simple use a atomic integer to replace default int for r.beg and check the margin before use it as a index for slice?

cyfdecyf: As I don't know the intended use case of `RingGrowing`, I don't know if it's appropriate to change this to a thread safe implementation.

If changing this to a thread safe version is the way to go, I'd suggest just use locks for simplicity. If performance is indeed a problem, consider use some proven implementation of concurrent queue.

deads2k:
I don't think that changing the method contract due to an improper usage is a good choice for this repo.  I'm not necessarily against providing a ThreadSafeRingGrowing wrapper.  But I'm also not entirely clear on where the shared_informer code went wrong.  I'm having some trouble getting the line numbers aligned.  See my comment in https://github.com/kubernetes/kubernetes/issues/106003#issuecomment-959959445

PR #226, opened 2021-11-01
Add wrapper func for spew.Sprintf in utils

haoruan:
**What type of PR is this?**
/kind cleanup

A wrapper of spew.Sprintf, so to pretty print object with a util function

Fixes kubernetes/kubernetes#8976

With existing usage of spew.Sprintf, the spew.Sdump is a better choice because it has the pretty format with full newlines, indentation and type.

**Release note**:
```

```

### DIFF
--- a/dump/dump.go
+++ b/dump/dump.go
@@ -1,0 +1,34 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package dump
+
+import (
+	"github.com/davecgh/go-spew/spew"
+)
+
+var prettyPrintConfig = &spew.ConfigState{
+	Indent:                  "  ",
+	DisableMethods:          true,
+	DisablePointerAddresses: true,
+	DisableCapacities:       true,
+}
+
+// PrettyPrintObject wrap the spew.Sdump with Indent, and disabled methods like error() and String()
+// The output may change over time, so for guaranteed output please take more direct control
+func PrettyPrintObject(a interface{}) string {
+	return prettyPrintConfig.Sdump(a)
+}
--- a/dump/dump_test.go
+++ b/dump/dump_test.go
@@ -1,0 +1,152 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package dump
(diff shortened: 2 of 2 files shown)


apelisse:
The original issue is 5 years old, and I don't know how much sense that makes today, is this still something we want? @thockin

thockin (commented):
Thanks for picking this up!  Unfortunately, I don't think this quite meets needs.

https://go.dev/play/p/Gvyo85V6YdA

First, `%+v` and `%#v` are not "pretty".  They are the opposite :)

Second, this doesn't disable methods, so somewhat less useful (would need this covered in test): https://go.dev/play/p/qciduo52P1q

Compare to litter (spew has more config, so does kr/pretty): https://go.dev/play/p/Z4Mz93Ip6Su

So I think the first step is to go look at k/k and see the places this would be used and figure out what most of them want.  Then come back to this and explain why whichever impl you think is right is the best choice.  Lastly, a rich test with maps, String() and Error() methods, etc

haoruan:
> Thanks for picking this up! Unfortunately, I don't think this quite meets needs.
> 
> https://go.dev/play/p/Gvyo85V6YdA
> 
> First, `%+v` and `%#v` are not "pretty". They are the opposite :)
> 
> Second, this doesn't disable methods, so somewhat less useful (would need this covered in test): https://go.dev/play/p/qciduo52P1q
> 
> Compare to litter (spew has more config, so does kr/pretty): https://go.dev/play/p/Z4Mz93Ip6Su
> 
> So I think the first step is to go look at k/k and see the places this would be used and figure out what most of them want. Then come back to this and explain why whichever impl you think is right is the best choice. Lastly, a rich test with maps, String() and Error() methods, etc

Thanks the advise, seems we're not limited by the format/lib, I will check again

haoruan:
spew.Sdump() has the pretty format with full newlines, indentation, type, and pointer information:
```
(main.Foo) {
 unexportedField: (*main.Bar)(0xf84002e210)({
  flag: (main.Flag) flagTwo,
  data: (uintptr) <nil>
 }),
 ExportedField: (map[interface {}]interface {}) {
  (string) "one": (bool) true
 }
}
```
 Also we can set Config.DisableMethods to disable String(), or Error().

spew.Sdump() has been widely used in kubernetes to dump object, or show the output diff in failed unit test, and spew.Sprintf() did the same thing with the same purpose. I'm not sure if they're reserved by purpose or we just missed them.

So my opinion is to make it uniform across the whole project, and replace Sprintf() with Sdump() instead of creating a wrapper. Later we can close the original issue, and create a new one to do the replacement.

thockin:
The reason I opened the issue was because it's all over the map.  Most places that want to dump something don't ever look at the logs, so don't notice that it uses .String() or prints pointer values or whatever.

thockin:
I realized this never got completed, sorry.

I don't love Sdump(), but it seems like least bad.  If we're going to do this, we should add `DisableCapacities`.  Spew seems unmaintained, maybe I'll fork it :)

We should probably call this package "dump" or "pretty" or similar.

We should document that the output of pretty.Print() may change over time and if anyone needs guaranteed output, they should take more direct control

haoruan:
Just update the package name, comments and remove the capacity in output, PTAL if we're going to take it.

thockin:
@BenTheElder @jpbetz @tallclair @cici37 @msau42 As discussed earlier, I recall now why JSON is not a great choice.

The GO JSON lib is very ...limited.

https://go.dev/play/p/ef_FY4KyrV-

Spew is unmaintained and large.  I am sure we could fork it and make it do what we want but....ugh

kr.Pretty is already used trasitively, though I can't see how to make it's output not include newlines.

litter is nicer output, but is a net new dep.  If we could get 3 or 4 deps which use spew to switch, we could justify this as a one-for-one.

@haoruan if you are still up for it, how about this:

1) move this kubernetes/kubernetes/pkg/util
2) convert all callsites in k/k to use this new function (same PR)
== We should have only one place that uses spew
3) we work with a few deps to kill off spew
4) we convert it to litter or something else
== No deps on spew
5) Move to kubernetes/utils

What do you think?

thockin:
yeah, let's do that, but please xref it so I don't forget :)

PR #227, opened 2021-11-08
IPv6 const should be type IPFamily

tyler-lloyd:
This causes unnecessary conversions between
string and IPFamily when using IPv4 and
IPv6.

Signed-off-by: Tyler Lloyd <tylerlloyd928@gmail.com>

**What type of PR is this?**

/kind cleanup

Consistency between IPv4 and IPv6 IP family types.


tyler-lloyd:
I see the change is failing the apidiff test I assume as a breaking change. I think this change, while potentially breaking in some cases, is welcome since if you're dealing with `net.IPv6` and `net.IPv4` then there is some awkward conversion happening somewhere since one could be used as an IPFamily type and the other a string.

PR #228, opened 2021-11-12
fix inotify memory leak

Dingshujie:
**What type of PR is this?**
 /kind bug

when delete file or dir , os will delete watch descriptor related to this path, file descriptor or watch descriptor not found, InotifyRmWatch syscall return EINVAL error, if return error, it may lead this path remain in watches and paths map, and no other event can trigger remove action. 

Fixes [#100241](https://github.com/kubernetes/kubernetes/issues/100241)

**Release note**:
```

```

### DIFF
--- a/inotify/inotify_linux.go
+++ b/inotify/inotify_linux.go
@@ -120,7 +120,11 @@ func (w *Watcher) RemoveWatch(path string) error {
 	}
 	success, errno := syscall.InotifyRmWatch(w.fd, watch.wd)
 	if success == -1 {
-		return os.NewSyscallError("inotify_rm_watch", errno)
+		// when file descriptor or watch descriptor not found, InotifyRmWatch syscall return EINVAL error
+		// if return error, it may lead this path remain in watches and paths map, and no other event can trigger remove action.
+		if errno != syscall.EINVAL {
+			return os.NewSyscallError("inotify_rm_watch", errno)
+		}
 	}
 	delete(w.watches, path)
 	// Locking here to protect the read from paths in readEvents.


thockin:
Is there a way we can test this?  It SEEMS like a reasonably plausible change, but how do I know?

Dingshujie:
@thockin, i wrote a test function
``` golang
package main

import (
	"flag"
	"fmt"
	"strings"
	"time"

	"k8s.io/utils/inotify"
)

func main() {
	var path string
	flag.StringVar(&path, "path", "", "path want to watch")
	flag.Parse()
	path = strings.TrimSpace(path)
	if len(path) == 0 {
		fmt.Println("error input path")
		return
	}
	watcher, err := inotify.NewWatcher()
	if err != nil {
		fmt.Println("create new watcher failed: ", err)
		return
	}
	err = watcher.Watch(path)
	if err != nil {
		fmt.Println("add watcher failed: ", err)
		return
	}
	for {
		select {
		case ev := <-watcher.Event:
			fmt.Println("event:", ev)
			eventType := ""
			switch {
			case ev.Mask&inotify.InDelete > 0:
				eventType = "delete"
			case ev.Mask&inotify.InDeleteSelf > 0:
				eventType = "delete"
			}
			if eventType == "delete" {
				time.Sleep(1 * time.Second)
                                err = watcher.RemoveWatch(path)
                                if err != nil {
                                        fmt.Println("remove watcher failed: ", err)
                                        return
                                }
			}
		case err := <-watcher.Error:
			fmt.Println("error:", err)
			return
		}
	}
}
``` 
watched input path, and i deleted this path,  test function watched InDelete or InDeleteSelf call RemoveWatch, you can see function return error

image

thockin:
I really meant an automated test, but I'll approve for now.

thockin:
Will you also update this in k/k ? I think we can make a case for this bugfix after code freeze, if we need to.

Dingshujie:
ok, i will submit a PR in k/k.

jonyhy96:
found that there are some projects may also needs update https://grep.app/search?q=k8s.io/utils/inotify

inotify/inotify_linux.go
```diff
@@ -120,7 +120,11 @@ func (w *Watcher) RemoveWatch(path string) error {
 	}
 	success, errno := syscall.InotifyRmWatch(w.fd, watch.wd)
 	if success == -1 {
-		return os.NewSyscallError("inotify_rm_watch", errno)
+		// when file descriptor or watch descriptor not found, InotifyRmWatch syscall return EINVAL error
+		// if return error, it may lead this path remain in watches and paths map, and no other event can trigger remove action.
+		if errno != syscall.EINVAL {
+			return os.NewSyscallError("inotify_rm_watch", errno)
+		}
```
gjkim42, line 127: Are we sure that the inotify object is actually removed whenever the EINVAL error occurs? Otherwise, this may introduce a kernel side memory leak.

PR #231, opened 2022-01-06
[WIP] Fork code from distribution/distribution

dims:
we had a dependency in kubernetes namely `github.com/docker/distribution`, this code from docker made its way to github.com/distribution/distribution on its way it picked up go modules and had a `v3` go module as well. the problem was we could not move to the newer thing as it was bringing in a whole boatload of transitive dependencies. please https://github.com/kubernetes/kubernetes/pull/107211#issuecomment-1006698414 for context.

So we fork directory from https://github.com/distribution/distribution/ trying to preserve the history

This brings some newer code than what's currently in kubernetes, they are:
- https://github.com/distribution/distribution/pull/2786/commits/0ac367fd6bee057d404c405a298b4b7aedf301ec
- https://github.com/distribution/distribution/pull/3265/commits/c26fa61ec9234e3b52181c6d6dc1204f6713329f
- https://github.com/distribution/distribution/pull/2979/commits/a19e1847a3a815894bdc9ffdc064392c96fa38c6

Note that the copy in k/k did not have any test cases as well. Here we do. the LICENSE file from the original repo was preserved as well in the `distribution/` directory.

There is a precedent for this, we did the same with `inotify` in https://github.com/kubernetes/utils/pull/96

Also we could not use the `kubernetes/kubernetes/third_party/` directory as there are things in the staging/ directory that need this code.

### DIFF
--- a/distribution/LICENSE
+++ b/distribution/LICENSE
@@ -1,0 +1,202 @@
+Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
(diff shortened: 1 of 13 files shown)


liggitt:
a few initial questions/thoughts:

1. were you preserving history for review and planning to squash?
2. can we prune everything unused by k/k? if so, does that let us drop the digestset package and go-units dependency? also, I think that would let us reduce to a single package, which we might want to name more descriptively (`imagespec` or something)
3. does the change in distribution/distribution@a19e184 modify behavior or validation of image specs? if so, I'm not sure we want to automatically pick that up (or would need to at least consider the impact to existing API data)
4. since changes to this parsing will impact API validation and behavior, it needs an owners file aligned with k/k API approval

dims:
> a few initial questions/thoughts:
> 
> 1. were you preserving history for review and planning to squash?

Not planning to squash, the history will provide enough context and more importantly attribution/copyright etc. which gets lost if we dont. We can layer additional commits to slim things down for sure. (either in the same PR or in follow ups)

> 2. can we prune everything unused by k/k? if so, does that let us drop the digestset package and go-units dependency? also, I think that would let us reduce to a single package, which we might want to name more descriptively (`imagespec` or something)

YES! i believe it's possible.

> 3. does the change in [distribution/distribution@a19e184](https://github.com/distribution/distribution/commit/a19e184) modify behavior or validation of image specs? if so, I'm not sure we want to automatically pick that up (or would need to at least consider the impact to existing API data)

Looking at the test case in https://github.com/distribution/distribution/pull/2979 it seems like the right thing to do. But yes, let's ask sig-node folks

> 4. since changes to this parsing will impact API validation and behavior, it needs an owners file aligned with k/k API approval

Agree will add sig-node approvers and reviewers

thockin:
I wonder if we should just have an org for these, so we can put them in their own repos.
 
I have been squatting on https://github.com/kubernetes-util and https://github.com/kubernetes-utils and https://github.com/kubernetes-x and https://github.com/k8s-internal.  Someone else has https://github.com/kubernetes-internal but it seems dead, maybe we could claw it back.

We could even (maybe?) use Go's `internal` semantics.  E.g. redirect `go get k8s.io/internal/*` to github.com/kubernetes-internal/* ?

dims:
@thockin if we have an org and make things easy, there will be a proliferation of these i am afraid ...

dims:
closing in favor of https://github.com/kubernetes/utils/pull/232

PR #232, opened 2022-01-07
[WIP] Fork code from docker/distribution and opencontainers/digest - take #2

dims:
we had a dependency in kubernetes namely `github.com/docker/distribution`, this code from docker made its way to github.com/distribution/distribution on its way it picked up go modules and had a `v3` go module as well. the problem was we could not move to the newer thing as it was bringing in a whole boatload of transitive dependencies. please https://github.com/kubernetes/kubernetes/pull/107211#issuecomment-1006698414 for context.

So we start with the code from the 2 projects at the version specified in kubernetes go.mod/sum:
- `github.com/opencontainers/go-digest` (v1.0.0)
- `github.com/docker/distribution` (v2.7.1+incompatible)

NOTE that we are NOT preserving git history. If we need history, folks will have to go to the original repositories.

And then we prune unnecessary files/code and adjust imports etc to have just a small subset of things used in k/k and delete anything we dont. Note that we still have what we copied over from k/k in a separate commit so if needed.

There is a precedent for this, we did the same with `inotify` in https://github.com/kubernetes/utils/pull/96

Also we could not use the `kubernetes/kubernetes/third_party/` directory as there are things in the staging/ directory that need this code.


smarterclayton:
This aligns similar to what openshift did (to focus on the stable "api" parts of distribution and image references that shouldn't be coupled to serving or client behavior) in https://github.com/openshift/library-go/tree/master/pkg/image/internal, so it looks good from that perspective.

I may suggest to upstream distribution/distribution folks that splitting that library and doing symbol refs might be the right thing to do long term.

PR #233, opened 2022-01-12
pointer: Add helper funcs for time.Duration

ialidzhikov:
/kind feature

The PR adds helper funcs to the pointer pkg for working with `time.Duration`.

**Release note**:
```
NONE
```

### DIFF
--- a/pointer/pointer_test.go
+++ b/pointer/pointer_test.go
@@ -19,6 +19,7 @@ package pointer
 import (
 	"fmt"
 	"testing"
+	"time"
 )
 
 func TestAllPtrFieldsNil(t *testing.T) {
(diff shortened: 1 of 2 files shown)


MadhavJivrajani (commented):
Could you please add unit tests?

thockin:
All the other types have a "simple" func name, eg. `pointer.Float64` - can you do the same with `Duration`?  The longer `Float64Ptr` is a back-compat thing.

IOW - there's a pattern, is there a reason not to follow it?

ialidzhikov:
> Could you please add unit tests?

Done.

> All the other types have a "simple" func name, eg. `pointer.Float64` - can you do the same with `Duration`? The longer `Float64Ptr` is a back-compat thing.
> 
> IOW - there's a pattern, is there a reason not to follow it?

Good point, I now changed from `DurationPtr` to `Duration`.

ialidzhikov:
@MadhavJivrajani @thockin did you had a chance to check the recent changes?

ialidzhikov:
Friendly ping @MadhavJivrajani @thockin

pointer/pointer_test.go
```diff
@@ -372,3 +373,49 @@ func TestFloat64Equal(t *testing.T) {
...
+	if DurationEqual(nil, Duration(42)) {
+		t.Errorf("expected false (nil != val)")
+	}
+	if DurationEqual(Duration(42), nil) {
+		t.Errorf("expected false (val != nil)")
+	}
+	if DurationEqual(Duration(42), Duration(43)) {
+		t.Errorf("expected false (val != val)")
+	}
+}
```
MadhavJivrajani, line 421: These look good!
Would it be possible to add these tests in a tabular form?

To be clear, I mean something like:
```go
func TestDuration(t *testing.T) {
    cases := []struct{
        name string // if applicable, optional
        val time.Duration
        ptr *time.Duration
    } {
        {...},
    }
    for _, case := range cases {
        ...
    }
}
```

MadhavJivrajani (commented):
Apologies for the late reply, suggested a minor change, thanks for raising the PR!

pointer/pointer_test.go
ialidzhikov: These tests are "inspired" by the existing tests that all follow the same approach. For consistency reasons, I would rather keep it as it is. I guess you can open issue to rework all existing tests to table tests. WDYT?

MadhavJivrajani: Let's keep it as is :)

MadhavJivrajani:
/assign @thockin 
in case you're around to take a final look and approve :)
Thanks!

thockin (commented):
Thanks!

/lgtm
/approve

PR #234, opened 2022-01-25
Add IPFamilyOf() / IPFamilyOfString() / IPFamilyOfCIDR() / IPFamilyOfCIDRString()

danwinship:
**What type of PR is this?**
/kind api-change
/kind feature

    Often you want to test whether two IPs/CIDRs are of the same family,
    without actually caring which family they are. While you can do this
    by checking:
    
      if utilsnet.IsIPv6(ip[0]) == utilsnet.IsIPv6(ip[1]) {
    
    this is somewhat confusing to read and can look like it is checking
    that they are both IPv6 if you read it quickly.
    
    So add new functions so that you can do
    
      if utilsnet.IPFamilyOf(ip[0]) == utilsnet.IPFamilyOf(ip[1]) {

Some examples:
- [kubelet](https://github.com/kubernetes/kubernetes/blob/ab69524f795c42094a6630298ff53f3c3ebab7f4/cmd/kubelet/app/server.go#L1154) (though that should be using `netutils.IsDualStackIPs()` anyway)
- [endpoints controller](https://github.com/kubernetes/kubernetes/blob/ab69524f795c42094a6630298ff53f3c3ebab7f4/pkg/controller/endpoint/endpoints_controller.go#L259)
- [iptables proxy](https://github.com/kubernetes/kubernetes/blob/ab69524f795c42094a6630298ff53f3c3ebab7f4/pkg/proxy/iptables/proxier.go#L1455)
- [ipvs proxy](https://github.com/kubernetes/kubernetes/blob/ab69524f795c42094a6630298ff53f3c3ebab7f4/pkg/proxy/ipvs/proxier.go#L559)
- tons of examples in [ovn-kubernetes](https://github.com/ovn-org/ovn-kubernetes/blob/362f2eb43dffd1a48135b3c7fd48c4ec88f45bed/go-controller/pkg/ovn/egressip.go#L2151)...

none

~I had thought utils wasn't API-stable, but I see in the README.md now that "stable, or backward compatible, API" is listed among the "criteria for adding code here".~

~I broke API by renaming "IPFamily" to "IPFamilyType" so I could use "IPFamily" as a function name. If I can't do that then I need alternate name suggestions...~

**Release note**:
```
Added IPFamilyOf() / IPFamilyOfString() / IPFamilyOfCIDR() /
IPFamilyOfCIDRString() functions to make it easier to check "are these two
IPs/CIDRs the same family?"

The value `IPv6 = "6"` from `k8s.io/utils/net` is now correctly typed as a value
of type `IPFamily` rather than being an untyped string.
```

/sig network
/cc @aojea


net/ipfamily.go
```diff
@@ -0,0 +1,178 @@
...
+
+package net
+
+import (
+	"fmt"
+	"net"
+)
+
+// IPFamilyType refers to a specific family if not empty, i.e. "4" or "6".
+type IPFamilyType string
```
aojea, line 25: I don't see the old type imported in k/k

```
$ grep -r netutils.IPFamily 
$ grep -r utilnet.IPFamily 
```

net/ipfamily_test.go
```diff
@@ -0,0 +1,418 @@
...
+		{
+			desc:   "ipv6 CIDR 2",
+			cidr:   "2000::/10",
+			family: IPv6,
+		},
+		{
+			desc:   "ipv6 CIDR 3",
+			cidr:   "2001:db8::/32",
+			family: IPv6,
+		},
```
aojea, line 394: test case for unknown?
```suggestion
		},{
			desc:   "Unknown IP family",
			cidr:   "invalid.cidr/mask",
			family: IPFamilyUnknown,
		},
		
```

aojea:
minor comment for adding a test case
LGTM but with the same doubt about the IPFamily -> IPFamilyType change
/assign @thockin 
for approval and try to clarify the doubts

net/ipfamily.go
thockin: I don't think we can do this.  We do claim a stable API.

net/ipfamily.go
```diff
@@ -106,50 +106,73 @@ func IsDualStackCIDRStrings(cidrs []string) (bool, error) {
 	return IsDualStackCIDRs(parsedCIDRs)
 }
 
-// IsIPv6 returns if netIP is IPv6.
-func IsIPv6(netIP net.IP) bool {
-	return netIP != nil && netIP.To4() == nil
+// IPFamily returns the IP family of netIP.
+func IPFamily(netIP net.IP) IPFamilyType {
```
thockin, line 110: `GetIPFamily()` or `IPFamilyOf()` or `FamilyOf()` or something?

net/ipfamily_test.go
danwinship: sure. (the tests that are in the PR now are just a consolidation of the pre-existing tests, which did not have any unknown/nil tests for CIDR)

thockin (commented):
Eventually we'll need to adapt to https://github.com/golang/go/issues/46518

Maybe the medium-term answer is something like 

```
type IP struct {
    net netip.Addr  // always set
    std net.IP  // set on first use
    str string  // set on first use
    valid bool
    err  error
}

func IPFromNetIP(in net.IP) IP { ... }
func IPFromString(in string) IP { ... }

func  (ip *IP) String() string {
    // need a lock
    if ip.str == "" {
        ip.str = ip.toString()
    }
    return ip.str
}

func  (ip *IP) Family IPFamily { ... }
```

That's a much bigger topic, of course.  Made more complicated by trying to unmarshal strings -> IP and retain the invalid value for helpful errors.

danwinship:
redone with `IPFamilyOf(net.IP)`, `IPFamilyOfString(string)`, `IPFamilyOfCIDR(*net.IPNet)`, `IPFamilyOfCIDRString(string)`

danwinship:
LOL. OK, apparently in the current code, `utilsnet.IPv4` is an `IPFamily`, but `utilsnet.IPv6` is an untyped string. I'm going to assume that's obviously a mistake and we should break API to fix it.

net/port.go
```diff
@@ -23,15 +23,6 @@ import (
 	"strings"
 )
 
-// IPFamily refers to a specific family if not empty, i.e. "4" or "6".
-type IPFamily string
-
-// Constants for valid IPFamilys:
-const (
-	IPv4 IPFamily = "4"
-	IPv6          = "6"
```
aojea, line 32: :)

thockin (commented):
Thanks!

/lgtm
/approve

danwinship:
/unassign thockin
so it will show up as needing an API reviewer...
/remove-lifecycle stale

danwinship:
API review needed because:
 
> This PR fixes an obvious bug where `netutils.IPv4` had the type `netutils.IPFamily`, but `netutils.IPv6` was an untyped string. Though this is an API break, it is unlikely to break any existing code. Among other things, it appears that k/k never got ported to use the utils-ified version of this API anyway, nor is it used by sigs.k8s.io/kpng.

Issue #236, opened 2022-01-28
ebtables: Fails to get ebtables version


defo89:
**What happened**:
Kubelet fails to get `ebtables` version. Apparently some versions do not contain `v` before the version number.

```
# ebtables --version
ebtables 1.8.7 (nf_tables)
```

```
kubelet-wrapper[3583]: I0128 10:17:28.917365    3583 kubenet_linux.go:765] "Failed to get ebtables version. Skip syncing ebtables dedup rules" err="no ebtables version found in string: ebtables 1.8.3 (nf_tables)\n"
```

**What you expected to happen**:
[versionMatcher](https://github.com/kubernetes/utils/blob/master/net/ebtables/ebtables.go#L124) should account for it.

**How to reproduce it**:
I reused the code snippet.

```
package main

import (
	"fmt"
	"regexp"
)

func getEbtablesVersionString(version string) (string, error) {
	versionMatcher := regexp.MustCompile(`v([0-9]+\.[0-9]+\.[0-9]+)`)
	match := versionMatcher.FindStringSubmatch(version)
	if match == nil {
		return "", fmt.Errorf("no ebtables version found in string: %s", version)
	}
	return match[1], nil
}

func main() {
	//version := "ebtables v2.0.10-4 (December 2011)"
	version := "ebtables 1.8.7 (nf_tables)"
	fmt.Printf(getEbtablesVersionString(version))
}
```

If `version := "ebtables 1.8.7 (nf_tables)"` then check fails
```
› ./main
%!(EXTRA *errors.errorString=no ebtables version found in string: ebtables 1.8.7 (nf_tables))
```

If `version := "ebtables v2.0.10-4 (December 2011)"`, it goes through.
```
› ./main
2.0.10%!(EXTRA <nil>)
```

**Anything else we need to know?**:
Issue does not occur on Flatcar 2905.2.3 5.10.61-flatcar as the version string is "compliant" :) 
```
$ ebtables --version
ebtables v2.0.10-4 (December 2011)
```

**Environment**:
- Kubernetes version (use `kubectl version`): 
```
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.5", GitCommit:"aea7bbadd2fc0cd689de94a54e5b7b758869d691", GitTreeState:"clean", BuildDate:"2021-09-15T21:10:45Z", GoVersion:"go1.16.8", Compiler:"gc", Platform:"linux/amd64"}
```
[...]

PR #237, opened 2022-01-28
ebtables: versionMatcher remove v before version

defo89:
**What type of PR is this?**
/kind bug

Do not expect that `ebtables --version` has v string before the version.

Fixes https://github.com/kubernetes/utils/issues/236

### DIFF
--- a/net/ebtables/ebtables.go
+++ b/net/ebtables/ebtables.go
@@ -121,7 +121,7 @@ func getEbtablesVersionString(exec utilexec.Interface) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	versionMatcher := regexp.MustCompile(`v([0-9]+\.[0-9]+\.[0-9]+)`)
+	versionMatcher := regexp.MustCompile(`([0-9]+\.[0-9]+\.[0-9]+)`)
 	match := versionMatcher.FindStringSubmatch(string(bytes))
 	if match == nil {
 		return "", fmt.Errorf("no ebtables version found in string: %s", bytes)


aojea:
yeah, I can confirm this is the case in 

Centos8
```
ebtables --version
ebtables 1.8.4 (nf_tables)
```

Fedora
```
ebtables --version
ebtables 1.8.7 (nf_tables)
```
maybe that was changed with the migration to nftables

aojea:
can we cover this with some unit tests to check that it works for both formats?

net/ebtables/ebtables.go
```diff
@@ -121,7 +121,7 @@ func getEbtablesVersionString(exec utilexec.Interface) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	versionMatcher := regexp.MustCompile(`v([0-9]+\.[0-9]+\.[0-9]+)`)
+	versionMatcher := regexp.MustCompile(`([0-9]+\.[0-9]+\.[0-9]+)`)
```
thockin, line 124: Maybe clearer with `v?` at the beginning?  And a comment explaining this is actually seen in the wild in both forms.

Agree with the point about a unit test.  Maybe factor this to a `parseVersion(string) (string, error)` function and test that with a few interesting inputs.

thockin: https://go.dev/play/p/C5_zp5HI8uX

thockin:
Hi @defo89 is this still something you are working on?

thockin:
Last ping before code freeze...

thockin:
Labelled this good-first-issue in case someone wants to take it over.

PR #238, opened 2022-02-23
Add path and args to exec cmd

eddiezane:
**What type of PR is this?**

> /kind feature

This PR adds Path and Args to the exec command proxy. I need access to the os.Command.Args for another PR and cannot type cast/assert.

Ref: https://github.com/kubernetes/kubernetes/pull/108199

**Release note**:
```
NONE
```


exec/testing/fake_exec.go
```diff
@@ -240,6 +240,14 @@ func (fake *FakeCmd) Stop() {
 	// no-op
 }
 
+func (fake *FakeCmd) GetPath() string {
+	return fake.Argv[0]
```
eddiezane, line 244: This would give the first arg which is the command name but does not do a realpath on it. Probably fine unless someone really needs to use this in a test scenario?

eddiezane:
/assign @lavalamp

I am not sure what to do about the apidiff failing. Does this need a release note for people when they upgrade to this so they can update their mocks?

liggitt:
Because we need changes to this repo to be compatible with a wide variety of Kubernetes versions, we don't allow breaking go API changes, and apidiff helps avoid merging accidental breaks. Adding methods to the interface would mean someone bumping to this version who had their own impl would no longer compile correctly. 

You can create a second interface that in lines the first and adds these two new methods.

eddiezane:
Going to switch to using plain old os/exec.

PR #239, opened 2022-03-15
Pointer: Add generic Pointer func

alvaroaleman:
This adds a new, generic pointer func to the pointer package.

**What type of PR is this?**
>
/kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

**Release note**:
```

```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,16 +4,16 @@ jobs:
   build:
     strategy:
       matrix:
-        go-versions: [1.13.x]
+        go-versions: [1.18.x]
         platform: [windows-latest]
     runs-on: ${{ matrix.platform }}
     steps:
       - name: Install Go
-        uses: actions/setup-go@v1
+        uses: actions/setup-go@v3
         with:
-          go-version: ${{ matrix.go-version }}
+          go-version: ${{ matrix.go-versions }}
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Build
         run: |
           go build ./...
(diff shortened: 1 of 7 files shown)


MadhavJivrajani (commented):
I think we'll need to update the `golangci-lint` version and explicitly mention `1.18`:
```
docker run --rm -v `pwd`:/go/src/k8s.io/klog -w /go/src/k8s.io/klog \
            golangci/golangci-lint:v1.45.0 golangci-lint run --go 1.18--disable-all -v \
            -E govet -E misspell -E gofmt -E ineffassign -E golint
```
See https://github.com/golangci/golangci-lint/issues/2649 for more details.

liggitt:
does this require all consumers of k8s.io/utils to set `go1.18` in their go.mod files, or just compile with go 1.18?

alvaroaleman:
@thockin the tests in ./internal/third_party/forked/golang/net do not pass on go versions > 1.16. Is this known? Can you have a look?

liggitt:
> Out of curiosity, why do we care? So importing projects that do not use this package can still use an older compiler?

changing go.mod to 1.17+ messes up the require directives in ways that break our dependency analysis tooling (xref https://github.com/golang/go/issues/47648)

liggitt:
> So importing projects that do not use this package can still use an older compiler?

also, we don't tag versions of k8s.io/utils and maintain multiple branches, so I probably wouldn't expect us to start using go1.18 features until go1.18 was the oldest supported go version (when go1.19 is released)

alvaroaleman:
I am sorry the PR has so many unrelated changes now, all of them are fallout from updating the go version:
* The github workflow never used the configured go version, because it referred an undefined variable from the matrix (`go-version` vs `go-versions`)
* pkg/inotify doesn't pass go vet on go 1.17+ because it calls t.Fatalf from a goroutine
* forked/golang/net doesn't pass tests on go 1.17+, precicely because of the change in behavior it is supposed to prevent
* gofmt needed updating in a couple of places to add the `go:build` directives added in go .17

euroelessar:
Just passing by, but can you consider the following changes, please?
1. Use `// +build go1.18` build tag in a file with generic implementations to keep compatibility with lower golang versions
2. In case of golang 1.18, implement existing methods using generic one, e.g. `var Int = Pointer[int]`
3. Keep an existing implementation for versions below 1.17 behind `!go1.18` build tag, this way an older implementation can be deleted once 1.18 is the lowest supported version
4. Provide an implementation for `Deref` and `Equal` methods as well

PR #241, opened 2022-04-04
IPv6 was not of IPFamily type

MushuEE:
See https://goplay.tools/snippet/xzIT8x5I_EZ for example.

There was a floating untyped string.  Should not impact much but safer to type enforce.

/kind bug
/kind cleanup

### DIFF
--- a/net/port.go
+++ b/net/port.go
@@ -28,7 +28,8 @@ type IPFamily string
 
 // Constants for valid IPFamilys:
 const (
-	IPv4, IPv6 IPFamily = "4", "6"
+	IPv4 IPFamily = "4"
+	IPv6 IPFamily = "6"
 )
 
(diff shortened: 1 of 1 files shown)


net/port.go
```diff
@@ -28,8 +28,7 @@ type IPFamily string
 
 // Constants for valid IPFamilys:
 const (
-	IPv4 IPFamily = "4"
-	IPv6          = "6"
+	IPv4, IPv6 IPFamily = "4", "6"
```
MushuEE, line 31:
```suggestion
	IPv4 IPFamily = "4"
	IPv6 IPFamily = "6"
```

MushuEE:
Unsure if we can get around
```
FAIL: k8s.io/utils/net contains incompatible changes:
- IPv6: changed from untyped string to IPFamily
 ```

PR #242, opened 2022-04-06
Ineffective break statements triggering SA4011

MushuEE:
These trigger https://staticcheck.io/docs/checks#SA4011

I am assuming we do not run static checks in CI but this one is a little dangerous because breaks are commonly misused to indicate breaking from the outer loop. Here it seems we just want to skip. I added "Do Nothing" to be explicit.

/kind cleanup

**Release note**:
```

```


MadhavJivrajani:
/lgtm
/remove-lifecycle rotten
/assign @dims 
Thanks!

PR #248, opened 2022-06-28
update github CI

aojea:
/kind failing-test

Update github actions testing jobs:

-  Update golang versions
- Update golangci-lint version
- Update actions dependencies

**Release note**:
```
NONE
```

### DIFF
--- a/internal/third_party/forked/golang/net/ip_test.go
+++ b/internal/third_party/forked/golang/net/ip_test.go
@@ -76,14 +76,6 @@ func TestParseIP(t *testing.T) {
 		if out := ParseIP(tt.in); !reflect.DeepEqual(out, tt.out) {
 			t.Errorf("ParseIP(%q) = %v, want %v", tt.in, out, tt.out)
 		}
-		if tt.in == "" {
-			// Tested in TestMarshalEmptyIP below.
-			continue
-		}
-		var out IP
-		if err := out.UnmarshalText([]byte(tt.in)); !reflect.DeepEqual(out, tt.out) || (tt.out == nil) != (err != nil) {
-			t.Errorf("IP.UnmarshalText(%q) = %v, %v, want %v", tt.in, out, err, tt.out)
-		}
 	}
 }
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,12 +4,12 @@ jobs:
   build:
     strategy:
       matrix:
-        go-versions: [1.13.x]
+        go-versions: [1.17.x, 1.18.x]
         platform: [windows-latest]
     runs-on: ${{ matrix.platform }}
     steps:
       - name: Install Go
-        uses: actions/setup-go@v1
+        uses: actions/setup-go@v3
         with:
           go-version: ${{ matrix.go-version }}
       - name: Checkout code
@@ -24,7 +24,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-versions: [1.12.x, 1.13.x, 1.14.x]
+        go-versions: [1.17.x, 1.18.x]
         platform: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.platform }}
     steps:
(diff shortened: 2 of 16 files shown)


liggitt:
```
=== RUN   TestParseIP
Error:     ip_test.go:85: IP.UnmarshalText("127.001.002.003") = <nil>, invalid IP address: 127.001.002.003, want 127.1.2.3
Error:     ip_test.go:85: IP.UnmarshalText("127.007.008.009") = <nil>, invalid IP address: 127.007.008.009, want 127.7.8.9
Error:     ip_test.go:85: IP.UnmarshalText("127.010.020.030") = <nil>, invalid IP address: 127.010.020.030, want 127.10.20.30
Error:     ip_test.go:85: IP.UnmarshalText("::ffff:127.001.002.003") = <nil>, invalid IP address: ::ffff:127.001.002.003, want 127.1.2.3
Error:     ip_test.go:85: IP.UnmarshalText("::ffff:127.007.008.009") = <nil>, invalid IP address: ::ffff:127.007.008.009, want 127.7.8.9
Error:     ip_test.go:85: IP.UnmarshalText("::ffff:127.010.020.030") = <nil>, invalid IP address: ::ffff:127.010.020.030, want 127.10.20.30
```

aojea:
ok, found it

aojea:
so, the net.IP.Unmarshall() method uses Parse

```
// UnmarshalText implements the encoding.TextUnmarshaler interface.
// The IP address is expected in a form accepted by ParseIP.
func (ip *IP) UnmarshalText(text []byte) error {
	if len(text) == 0 {
		*ip = nil
		return nil
	}
	s := string(text)
	x := ParseIP(s)
	if x == nil {
		return &ParseError{Type: "IP address", Text: s}
	}
	*ip = x
	return nil
}
```

aojea:
> so, the net.IP.Unmarshall() method uses Parse
> 
> ```
> // UnmarshalText implements the encoding.TextUnmarshaler interface.
> // The IP address is expected in a form accepted by ParseIP.
> func (ip *IP) UnmarshalText(text []byte) error {
> 	if len(text) == 0 {
> 		*ip = nil
> 		return nil
> 	}
> 	s := string(text)
> 	x := ParseIP(s)
> 	if x == nil {
> 		return &ParseError{Type: "IP address", Text: s}
> 	}
> 	*ip = x
> 	return nil
> }
> ```

we are not using it in k/k, so it may be ok to not override it https://grep.app/search?q=UnmarshalText&filter[lang][0]=Go&filter[repo][0]=kubernetes/kubernetes&filter[path][0]=staging/src/k8s.io/

aojea:
/assign @liggitt @thockin 

ok, this is working now

internal/third_party/forked/golang/net/ip_test.go
```diff
@@ -76,14 +76,6 @@ func TestParseIP(t *testing.T) {
 		if out := ParseIP(tt.in); !reflect.DeepEqual(out, tt.out) {
 			t.Errorf("ParseIP(%q) = %v, want %v", tt.in, out, tt.out)
 		}
-		if tt.in == "" {
-			// Tested in TestMarshalEmptyIP below.
-			continue
-		}
-		var out IP
-		if err := out.UnmarshalText([]byte(tt.in)); !reflect.DeepEqual(out, tt.out) || (tt.out == nil) != (err != nil) {
```
aojea, line 84: this only worked before because was tested with golang < 1.17

aojea:
Ping @thockin 

We have to update the ci jobs or we can risk to merge something without proper testing

thockin (commented):
Do we have an org-wide policy on using GH actions vs Prow?

/approve
/lgtm

aojea:
> Do we have an org-wide policy on using GH actions vs Prow?
> 
> /approve /lgtm

Let's ask @BenTheElder

BenTheElder:
> Do we have an org-wide policy on using GH actions vs Prow?

No, we use both.

Entirely too much detail on actions vs prow:
<details>
GitHub actions can be pretty nice. The main unfortunate downside is that the merge robot cannot guarantee linearized test state.  Prow testing + merge robot ensures that prior to merge tests pass with CI locally merging then testing (HEAD of target branch, HEAD of PR), whereas other CI platforms tend to just checkout and test the PR branch without any sort of merge / rebase.

For a repo like Kubernetes/Kubernetes, I consider not being able to enforce this a bit of a deal breaker.

For other repos with lower PR frequency etc this is less of an issue, my remaining complaint is less access for PR authors (no `/retest` though theoretically someone could fix this https://github.com/kubernetes/test-infra/issues/25831).

The upshot is we don't pay for / staff it, there are pretty good docs, and it's possible to do some things we can't do easily on Prow like run on mac, or (ab)use mac runners to run arbitrary linux VMs without worrying about cleanup (e.g. to test cgroupsv2 in KIND).

Actions also spook me because people tend to use pre-written "actions" by referencing repo@tag which we don't control, that's mitigate-able though by either requiring people to use repo@commit-hash, or by not using third party actions or only using inline shell steps / invoking local scripts.
<details>

PR #249, opened 2022-07-13
Properly deprecate the funcs from the pointer pkg

ialidzhikov:
**What type of PR is this?**
/kind cleanup

This PR properly deprecates the funcs from the pointer pkg that are kept only for backwards compatibility reasons. In this way linters will report usage of deprecated var when a consumer package tries to use the deprecated funcs.

Fixes #

**Release note**:
```
NONE
```

### DIFF
--- a/pointer/pointer.go
+++ b/pointer/pointer.go
@@ -52,6 +52,8 @@ func Int(i int) *int {
 	return &i
 }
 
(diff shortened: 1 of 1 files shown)


thockin (commented):
Thanks!

/lgtm
/approve

PR #253, opened 2022-08-08
add uint64 and uint to pointer pkg

kschoche:
**What type of PR is this?**
/kind feature

This PR adds support for interacting with 32 and 64bit unsigned ints in the `pointer` package.
We wanted to standardize on the pointer package but found that it was lacking unsigned int support which is commonly used in k8s and applications interacting with k8s. 

Fixes #

I'm not 100% sure if it makes sense to add the back-compatibility pieces which I copied from the existing `Int`and `Int64` implementations. If not I can back them out and just leave the new pieces.

**Release note**:
```

```

### DIFF
--- a/pointer/pointer.go
+++ b/pointer/pointer.go
@@ -103,6 +103,62 @@ func Int32Equal(a, b *int32) bool {
 	return *a == *b
 }
 
+// Uint returns a pointer to an uint
+func Uint(i uint) *uint {
+	return &i
+}
+
+// UintPtr is a function variable referring to Uint.
+// Deprecated: Use Uint instead.
+var UintPtr = Uint // for back-compat
+
+// UintDeref dereferences the uint ptr and returns it if not nil, or else
+// returns def.
+func UintDeref(ptr *uint, def uint) uint {
+	if ptr != nil {
+		return *ptr
+	}
+	return def
+}
+
+// UintPtrDerefOr is a function variable referring to UintDeref.
+// Deprecated: Use UintDeref instead.
+var UintPtrDerefOr = UintDeref // for back-compat
+
+// Uint32 returns a pointer to an uint32.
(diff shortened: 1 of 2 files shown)


apelisse:
I'd sort of expect all these things to now be defined by golang since I'm assuming they can be implemented with generics?

kschoche:
> I'd sort of expect all these things to now be defined by golang since I'm assuming they can be implemented with generics?

Hi @apelisse, thanks for the reply!
I had seen that there have been some PRs already that attempted to introduce generics to the pointer package [PR](https://github.com/kubernetes/utils/pull/239) but it seems like there were some blockers with the requirements for using go 1.18. Is that still the case? If so, I'm wondering what to do in the meantime for those of us who may also have the same concerns?

The use case I'm trying to address is to provide coverage in the pointer package for Uint64 as we are trying to use it with some CRD code that uses it for things like versions and timestamps.

I do understand that an obvious+easy answer is for users to BYO functionality, which is what we already do, and so this PR is more of a "nice to have" vs a necessity.

Thoughts?
Thanks!
~Kyle

apelisse:
I guess I don't see any problem with this PR being merged in the mean time.

Where do you need this code? Or is it strictly hypothetical?

kschoche:
Cool.
I ran into this while doing some code cleanup on `hashicorp/consul-k8s` where we've to date just been rolling our own version of the `pointer` pkg for interacting with CR specs in our CRD controllers. Right now if I could get this merged then I could fully move over to `pointer`.

kschoche:
Hi @apelisse - sorry for the ping here, but I just wanted to double-check if there were any additional changes you might see needed for this PR while it's fresh in my mind. Cheers.

apelisse:
Sorry, I missed your reply.

/lgtm
/approve

PR #261, opened 2022-09-22
Drop test-only dependencies

liggitt:
**What type of PR is this?**
/kind cleanup

```release-note
NONE
```

/cc @dims


dims:
ah nice! thanks @liggitt reviewing now

PR #262, opened 2022-10-12
feat: add netip utils

saltbo:
**What type of PR is this?**
/kind feature

step to step replace the `netutils` by the new pkg `netiputils`

https://github.com/kubernetes/kubernetes/issues/106161

**Release note**:
```

```

### DIFF
--- a/net/netip/ipnet.go
+++ b/net/netip/ipnet.go
@@ -1,0 +1,224 @@
+/*
+Copyright 2016 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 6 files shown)


saltbo:
test failed because of the error gha, wait for pr #263 to merged

jpbetz:
@saltbo Can we put this in https://github.com/kubernetes/kubernetes/tree/master/pkg/util instead?

A couple motivations:
- We'd like to get some evidence of adoption before promoting things into kubernetes/util
- With go generics on the horizon, per-type custom sets in kubernetes/util are not going to be as valuable as they were in the past

thockin:
I like this idea - show me how you will use this lib across k/k and
that this is the right API still, and then we can promote it.  k/utils
is the "for real, we support this long-term" repo.  k/k/pkg/utils has
a lower bar and more freedom to make some mistakes in things like API
design.

saltbo:
got it

PR #264, opened 2022-10-19
[interface-breaking] Add optional OpenTelemetry tracing to the utils/trace package

dashpole:
Part of https://github.com/kubernetes/kubernetes/issues/113170, which is a beta requirement for the API Server tracing KEP: https://github.com/kubernetes/enhancements/issues/647

**What type of PR is this?**

> /kind feature

This PR allows libraries which already use OpenTelemetry for tracing to have spans and span events recorded when this library is used.

If the user does not use OpenTelemetry, the OpenTelemetry calls are a No-Op.  **This does NOT modify existing logging-based behavior.**

This will allow OpenTelemetry-based tracing to achieve parity with text-based tracing without maintaining duplicate instrumentation code.

**Notes to reviewers**:

I would encourage reviewing commit-by-commit.

**Release note**:
```
utils/trace now requires Context as an argument, and accepts opentelemetry attributes, rather than Field structs.
```


dashpole:
cc @logicalhan @dgrisonnet

dashpole:
/hold
To make sure others have a chance to review

logicalhan (approved):
/lgtm
/approve

liggitt:
Hmm, both the interface break and the dependency additions aren't in line with how we run this repo. Is this possible to accomplish with local interfaces and an additive interface or constructor?

/hold

dashpole:
> Hmm, both the interface break and the dependency additions aren't in line with how we run this repo. Is this possible to accomplish with local interfaces and an additive interface or constructor?

It is possible to accomplish it with new functions instead of breaking the existing ones.  But it isn't possible to accomplish without the additional dependency.

As an alternative, I could define the new set of functions for tracing in component-base, and have those functions call into utils/traces and otel's tracing functions.  Let me know if you'd prefer that

dashpole:
See https://github.com/kubernetes/kubernetes/pull/113183 for the alternative approach.

dashpole:
Closing in favor of https://github.com/kubernetes/kubernetes/pull/113183

PR #269, opened 2022-11-02
Add generic Ref, Deref and Equal to the pointer package

alculquicondor:
**What type of PR is this?**

/kind feature

Add Ref, Deref and Equal that use generics. This simplifies implementation and also allow to have pointers of enums or arbitrary types.

Fixes #

**Release note**:
```
Add Ref, Deref and Equal to pointer package to handle pointers generically
```

### DIFF
--- a/pointer/pointer.go
+++ b/pointer/pointer.go
@@ -47,57 +47,20 @@ func AllPtrFieldsNil(obj interface{}) bool {
 	return true
 }
 
-// Int returns a pointer to an int
-func Int(i int) *int {
-	return &i
+// Ref returns a pointer to the value.
+func Ref[T any](v T) *T {
+	return &v
 }
 
-// IntPtr is a function variable referring to Int.
-//
-// Deprecated: Use Int instead.
-var IntPtr = Int // for back-compat
-
-// IntDeref dereferences the int ptr and returns it if not nil, or else
-// returns def.
-func IntDeref(ptr *int, def int) int {
+// Deref dereferences the ptr and returns it if not nil, or else returns def.
+func Deref[T any](ptr *T, def T) T {
 	if ptr != nil {
 		return *ptr
 	}
 	return def
 }
 
-// IntPtrDerefOr is a function variable referring to IntDeref.
-//
-// Deprecated: Use IntDeref instead.
-var IntPtrDerefOr = IntDeref // for back-compat
-
-// Int32 returns a pointer to an int32.
-func Int32(i int32) *int32 {
-	return &i
-}
-
-// Int32Ptr is a function variable referring to Int32.
-//
-// Deprecated: Use Int32 instead.
-var Int32Ptr = Int32 // for back-compat
-
-// Int32Deref dereferences the int32 ptr and returns it if not nil, or else
-// returns def.
-func Int32Deref(ptr *int32, def int32) int32 {
-	if ptr != nil {
-		return *ptr
-	}
-	return def
-}
-
-// Int32PtrDerefOr is a function variable referring to Int32Deref.
-//
-// Deprecated: Use Int32Deref instead.
-var Int32PtrDerefOr = Int32Deref // for back-compat
-
-// Int32Equal returns true if both arguments are nil or both arguments
-// dereference to the same value.
-func Int32Equal(a, b *int32) bool {
+func Equal[T comparable](a, b *T) bool {
 	if (a == nil) != (b == nil) {
 		return false
 	}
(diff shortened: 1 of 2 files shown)


alculquicondor:
Ahhh, actually, this seems to be against the policy https://github.com/kubernetes/community/blob/master/sig-architecture/generics.md#generics-policy

alculquicondor:
/hold
until 2023-07-28, as per https://kubernetes.io/releases/patch-releases/#1-24

/lifecycle frozen

thockin:
Are we doing generics in utils now?


@logicalhan @dims @BenTheElder

BenTheElder:
I think the comments above already note that and are holding the PR until that time though?

thockin:
I missed that.  Mea culpa.

alculquicondor:
The guidance is now adjusted to the EoL for 1.23, which is on `2023-02-28`

pointer/pointer.go
```diff
@@ -47,57 +47,20 @@ func AllPtrFieldsNil(obj interface{}) bool {
 	return true
 }
 
-// Int returns a pointer to an int
-func Int(i int) *int {
-	return &i
+// Ref returns a pointer to the value.
+func Ref[T any](v T) *T {
```
thockin, line 51: Why not "New"  `ptr.New[int64](12345)` is as close as we can get to `&int64(12345)` or `new(int64, 12345)`

alculquicondor: I liked how complementary `Ref` and `Deref` sounded. But `New` would work too.

alculquicondor:
FWIIW, 1.23 is using golang 1.19 now, but I don't think there was a release with it.

alculquicondor:
For sure there has been a release with 1.18 https://github.com/kubernetes/kubernetes/pull/113983

@BenTheElder, can we lift the restriction?

alculquicondor:
cc @liggitt

liggitt:
> For sure there has been a release with 1.18 [kubernetes/kubernetes#113983](https://github.com/kubernetes/kubernetes/pull/113983)

Actually, there hasn't... 1.23 jumped from go1.17 to go1.19 and is not yet released on that.

I still think we should still wait until 1.23/1.24 are out of support since library consumers can still be on the original go minor versions those were released with

alculquicondor:
> since library consumers can still be on the original go minor versions those were released with

But if they upgrade the libraries they would need to upgrade golang, or not necessarily?

liggitt:
> But if they upgrade the libraries they would need to upgrade golang, or not necessarily?

No, building / testing 1.23 with go1.17 and 1.24 with go1.18 still works.

thockin:
So we're NOT accelerating this?  Just confirming.

liggitt:
> So we're NOT accelerating this? Just confirming.

That would be my vote. I think we should stick to the original timeline, so we don't put ourselves in a position where a bump of k8s.io/utils in the 1.23 or 1.24 release branch would force a library consumer to update go versions on a patch bump.

alculquicondor:
But 1.24 was originally released with go=1.18, which supports generics. So we should good to start using generics when 1.23 goes EoL.

shaneutt:
Still holding until July, to my understanding?

/remove-lifecycle stale

pointer/pointer.go
skitt: How about `To`? `pointer.To(12345)` seems even nicer to me...

pointer/pointer.go
```diff
@@ -107,304 +70,218 @@ func Int32Equal(a, b *int32) bool {
...
-// returns def.
-func UintDeref(ptr *uint, def uint) uint {
-	if ptr != nil {
-		return *ptr
-	}
-	return def
-}
+	// IntPtr is a function variable referring to Int.
+	//
+	// Deprecated: Use Int instead.
```
skitt, line 80: This might as well point to `Ref` directly since that’s what people updating deprecated code should end up using.

pointer/pointer.go
```diff
@@ -107,304 +70,218 @@ func Int32Equal(a, b *int32) bool {
...
+	// returns def.
+	IntDeref = Deref[int]
 
-// Uint32 returns a pointer to an uint32.
-func Uint32(i uint32) *uint32 {
-	return &i
-}
+	// IntPtrDerefOr is a function variable referring to IntDeref.
+	//
+	// Deprecated: Use IntDeref instead.
```
skitt, line 89: Same here: this might as well point to `Deref` directly. (Same goes for all the other instances.)

skitt (changes requested):

pointer/pointer.go
thockin: `pointer.New()`, `pointer.To()`, `pointer.P()` would all be fine to me.  We could even rename the pkg to `ptr`.

alculquicondor: can we? I guess we would need to duplicate the package first.

But we are still waiting for 1.24 to go EoL next month.

skitt: It wouldn’t have to be duplicated — you could add a generic `ptr` package, and change `pointer` to be full of vars pointing to `ptr` functions.

skitt: One benefit of introducing a new generic `ptr` package instead of updating the existing `pointer` package would be to simplify migration tracking: any references to `utils/pointer` would point (hah) to code that should be updated.

alculquicondor: I agree.

@skitt would you like to take over? The time for adding these generics is coming next month, and I honestly have limited time to do so.

I can simply close this PR and you can start over with the `ptr` package approach.

skitt: I can take care of this, yes!

PR #270, opened 2022-11-07
Fix trace panic when concurrently adding children

dashpole:
**What type of PR is this?**

/kind failing-test

Attempts to fix a panic found in integration tests.  The panics were introduced after nested traces were introduced in https://github.com/kubernetes/kubernetes/pull/113172.

This adds a test which easily caused a panic without the locking.

Fixes https://github.com/kubernetes/kubernetes/issues/113651

**Release note**:
```
NONE
```

cc @liggitt @aojea @logicalhan

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"sync"
 	"time"
 
 	"k8s.io/klog/v2"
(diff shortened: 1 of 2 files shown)


aojea:
great catch

one question, the only thing that I can see this can deadlock is if a nested trace can point to itself as a parent, that can not happen, right?

dashpole:
That cannot happen because parent is only set during `Nest`, and is only set on a trace that is also created within `Nest`

aojea:
/lgtm

Thanks

PR #271, opened 2022-11-09
Add IPFamilyOf() / IPFamilyOfString() / IPFamilyOfCIDR() / IPFamilyOfCIDRString()

danwinship:
**What type of PR is this?**
/kind api-change
/kind feature

    Often you want to test whether two IPs/CIDRs are of the same family,
    without actually caring which family they are. While you can do this
    by checking:
    
      if utilsnet.IsIPv6(ip) == utilsnet.IsIPv6CIDR(cidr) {
    
    this is somewhat confusing to read and can look like it is checking
    that they are both IPv6 if you read it quickly.
    
    So add new functions so that you can do
    
      if utilsnet.IPFamilyOf(ip) == utilsnet.IPFamilyOfCIDR(cidr) {

Some examples:
- [kubelet](https://github.com/kubernetes/kubernetes/blob/ab69524f795c42094a6630298ff53f3c3ebab7f4/cmd/kubelet/app/server.go#L1154) (though that should be using `netutils.IsDualStackIPs()` anyway)
- [endpoints controller](https://github.com/kubernetes/kubernetes/blob/ab69524f795c42094a6630298ff53f3c3ebab7f4/pkg/controller/endpoint/endpoints_controller.go#L259)
- [iptables proxy](https://github.com/kubernetes/kubernetes/blob/ab69524f795c42094a6630298ff53f3c3ebab7f4/pkg/proxy/iptables/proxier.go#L1455)
- [ipvs proxy](https://github.com/kubernetes/kubernetes/blob/ab69524f795c42094a6630298ff53f3c3ebab7f4/pkg/proxy/ipvs/proxier.go#L559)
- tons of examples in [ovn-kubernetes](https://github.com/ovn-org/ovn-kubernetes/blob/362f2eb43dffd1a48135b3c7fd48c4ec88f45bed/go-controller/pkg/ovn/egressip.go#L2151)...

none

Re-post of #234 now that the API issue around `net.IPv6` being an untyped string has been fixed separately (#265).

**Release note**:
```
Added IPFamilyOf() / IPFamilyOfString() / IPFamilyOfCIDR() /
IPFamilyOfCIDRString() functions to make it easier to check "are these two
IP strings / net.IPs / CIDR strings / net.IPNets the same family?"
```

/sig network
/cc @aojea @thockin
(who both reviewed it before in #234 and approved it; the endpoint here is the same, but I refactored from 2 commits to 4)


net/ipfamily.go
```diff
@@ -0,0 +1,178 @@
...
+		if v4Found && v6Found {
+			continue
+		}
+
+		if IsIPv6(ip) {
+			v6Found = true
+			continue
+		}
+
+		v4Found = true
```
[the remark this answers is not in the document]
thockin:
```diff
diff --git a/net/net.go b/net/net.go
index b7c08e2..b500f4c 100644
--- a/net/net.go
+++ b/net/net.go
@@ -22,6 +22,7 @@ import (
 	"math"
 	"math/big"
 	"net"
+	"net/netip"
 	"strconv"
 )
 
@@ -113,49 +114,61 @@ func IsDualStackCIDRStrings(cidrs []string) (bool, error) {
 	return IsDualStackCIDRs(parsedCIDRs)
 }
 
-// IsIPv6 returns if netIP is IPv6.
-func IsIPv6(netIP net.IP) bool {
-	return netIP != nil && netIP.To4() == nil
+// IsIPv6 tests whether ip is IPv6.  Not being IPv6 does not imply that it is
+// necessarily IPv4.
+func IsIPv6(ip net.IP) bool {
+	nip, ok := netip.AddrFromSlice(ip)
+	if !ok {
+		return false
+	}
+	nip = nip.Unmap()
+	return nip.Is6()
 }
 
-// IsIPv6String returns if ip is IPv6.
+// IsIPv6String tests whether ip is IPv6.  Not being IPv6 does not imply that
+// it is necessarily IPv4.
 func IsIPv6String(ip string) bool {
-	netIP := ParseIPSloppy(ip)
-	return IsIPv6(netIP)
+	return IsIPv6(ParseIPSloppy(ip))
 }
 
-// IsIPv6CIDRString returns if cidr is IPv6.
-// This assumes cidr is a valid CIDR.
+// IsIPv6CIDRString tests whether cidr is IPv6.  Not being IPv6 does not imply
+// that it is necessarily IPv4.
 func IsIPv6CIDRString(cidr string) bool {
 	ip, _, _ := ParseCIDRSloppy(cidr)
 	return IsIPv6(ip)
 }
 
-// IsIPv6CIDR returns if a cidr is ipv6
+// IsIPv6CIDR tests whether cidr is IPv6.  Not being IPv6 does not imply
+// that it is necessarily IPv4.
 func IsIPv6CIDR(cidr *net.IPNet) bool {
-	ip := cidr.IP
-	return IsIPv6(ip)
+	return IsIPv6(cidr.IP)
 }
 
-// IsIPv4 returns if netIP is IPv4.
-func IsIPv4(netIP net.IP) bool {
-	return netIP != nil && netIP.To4() != nil
+// IsIPv4 tests whether ip is IPv6.  Not being IPv6 does not imply that it is
+// necessarily IPv6.
+func IsIPv4(ip net.IP) bool {
+	nip, ok := netip.AddrFromSlice(ip)
+	if !ok {
+		return false
+	}
+	nip = nip.Unmap()
+	return nip.Is4()
 }
 
-// IsIPv4String returns if ip is IPv4.
+// IsIPv4String tests whether ip is IPv4.  Not being IPv4 does not imply that it is
+// necessarily IPv4.
 func IsIPv4String(ip string) bool {
-	netIP := ParseIPSloppy(ip)
-	return IsIPv4(netIP)
+	return IsIPv4(ParseIPSloppy(ip))
 }
 
-// IsIPv4CIDR returns if a cidr is ipv4
+// IsIPv4CIDR tests whether cidr is IPv4.  Not being IPv4 does not imply that it is
+// necessarily IPv4.
 func IsIPv4CIDR(cidr *net.IPNet) bool {
-	ip := cidr.IP
-	return IsIPv4(ip)
+	return IsIPv4(cidr.IP)
 }
 
-// IsIPv4CIDRString returns if cidr is IPv4.
-// This assumes cidr is a valid CIDR.
+// IsIPv4CIDRString tests whether cidr is IPv4.  Not being IPv4 does not imply
+// that it is necessarily IPv4.
 func IsIPv4CIDRString(cidr string) bool {
 	ip, _, _ := ParseCIDRSloppy(cidr)
 	return IsIPv4(ip)
```

net/ipfamily.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	return IsDualStackCIDRs(parsedCIDRs)
+}
+
+// IPFamilyOf returns the IP family of netIP.
+func IPFamilyOf(netIP net.IP) IPFamily {
+	switch {
+	case netIP == nil:
+		return IPFamilyUnknown
+	case netIP.To4() != nil:
```
thockin, line 114: As above - this doesn't hold up under testing.

net/ipfamily.go
```diff
@@ -0,0 +1,178 @@
...
+
+package net
+
+import (
+	"fmt"
+	"net"
+)
+
+// IPFamily refers to a specific family if not empty, i.e. "4" or "6".
+type IPFamily string
```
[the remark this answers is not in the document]
danwinship: because this is just moving existing code. I don't know why they chose string originally

net/ipfamily.go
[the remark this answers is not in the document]
danwinship: OK, yes, `IsIPv6(net.IP([]byte{}))` will return `true`. But I believe:
- `IsIPv4()` correctly handles all inputs
- `IsIPv6()` correctly handles all valid `net.IP`s, and `nil`
- `IsIPv4String()` and `IsIPv6String()` correctly handle all inputs (returning `false` for anything that can't be converted to a valid `net.IP`)

we can fix `IsIPv6()` to DTRT with invalid `net.IPs` too if you want. I don't think there's any need to drag in `netip`.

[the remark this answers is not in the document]
thockin: Yeah, my little hack test was using `ip.To16()` rather than `!ip.To4()` which was poor fidelity.  Better testing shows I was wrong.

https://go.dev/play/p/K9H0pJx8eCg

The only one that is odd (and both impls get it wrong) is  "::FFFF:C0A8:1" and I am not sure how much we care?

I also wrote a benchmark for this, and it seems netip is BARELY faster:

```go
// You can edit this code!
// Click here and start typing.
package main

import (
	"net"
	"net/netip"
	"testing"
)

//go:noinline
func curIsIPv6(netIP net.IP) bool {
	return netIP != nil && netIP.To4() == nil
}

//go:noinline
func curIsIPv4(netIP net.IP) bool {
	return netIP != nil && netIP.To4() != nil
}

func addrFromIP(ip net.IP) netip.Addr {
	nip, ok := netip.AddrFromSlice(ip)
	if !ok {
		return netip.Addr{}
	}
	return nip.Unmap()
}

//go:noinline
func newIsIPv4(ip net.IP) bool {
	return addrFromIP(ip).Is4()
}

//go:noinline
func newIsIPv6(ip net.IP) bool {
	return addrFromIP(ip).Is6()
}

func do(b *testing.B, fam int) {
	cases := []struct {
		str string
		is4 bool
		is6 bool
	}{
		{"", false, false},
		{"0.0.0.0", true, false},
		{"127.0.0.1", true, false},
		{"::FFFF:C0A8:1", false, true},
		{"::1", false, true},
	}
	for _, tc := range cases {
		ip := net.ParseIP(tc.str)

		for i := 0; i < b.N; i++ {
			if fam == 4 {
				_ = curIsIPv4(ip)
				_ = curIsIPv6(ip)
			} else if fam == 6 {
				_ = newIsIPv4(ip)
				_ = newIsIPv6(ip)
			}
		}
	}
}

func BenchmarkCur(b *testing.B) {
	b.ReportAllocs()
	do(b, 4)
}

func BenchmarkNew(b *testing.B) {
	b.ReportAllocs()
	do(b, 6)
}
```

```
BenchmarkCur-6   	18189759	        57.13 ns/op	       0 B/op	       0 allocs/op
BenchmarkNew-6   	19459976	        53.26 ns/op	       0 B/op	       0 allocs/op
```

net/ipfamily.go
```diff
@@ -0,0 +1,180 @@
...
+
+// IsDualStackIPStrings returns true if:
+// - all elements of ips can be parsed as IPs
+// - at least one IP from each family (v4 and v6) is present
+func IsDualStackIPStrings(ips []string) (bool, error) {
+	parsedIPs := make([]net.IP, 0, len(ips))
+	for _, ip := range ips {
+		parsedIP := ParseIPSloppy(ip)
+		if parsedIP == nil {
+			return false, fmt.Errorf("ip %q is invalid", ip)
```
thockin, line 63: `return false, fmt.Errorf("invalid IP[%d]: %v", i, ip)`

net/ipfamily.go
```diff
@@ -0,0 +1,180 @@
...
+	v4Found := false
+	v6Found := false
+	for _, cidr := range cidrs {
+		switch IPFamilyOfCIDR(cidr) {
+		case IPv4:
+			v4Found = true
+		case IPv6:
+			v6Found = true
+		default:
+			return false, fmt.Errorf("list contains an invalid CIDR")
```
thockin, line 83: `return false, fmt.Errorf("invalid CIDR[%d]: %v", i, cidr)`

net/ipfamily.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	return (v4Found && v6Found), nil
+}
+
+// IsDualStackCIDRStrings returns if
+// - all elements of cidrs can be parsed as CIDRs
+// - at least one CIDR from each family (v4 and v6) is present
+func IsDualStackCIDRStrings(cidrs []string) (bool, error) {
+	parsedCIDRs, err := ParseCIDRs(cidrs)
```
thockin, line 94: check for good error message there, too?

net/ipfamily.go
```diff
@@ -0,0 +1,180 @@
...
+	if err != nil {
+		return false, err
+	}
+	return IsDualStackCIDRs(parsedCIDRs)
+}
+
+// IPFamilyOf returns the IP family of netIP.
```
thockin, line 101: ...or else IPFamilyUnknown

net/ipfamily.go
```diff
@@ -0,0 +1,180 @@
...
+	case netIP.To4() != nil:
+		return IPv4
+	case netIP.To16() != nil:
+		return IPv6
+	default:
+		return IPFamilyUnknown
+	}
+}
+
+// IPFamilyOfString returns the IP family of ip.
```
thockin, line 113: ...or else IPFamilyUnknown

net/ipfamily.go
```diff
@@ -0,0 +1,180 @@
...
+func IPFamilyOfString(ip string) IPFamily {
+	return IPFamilyOf(ParseIPSloppy(ip))
+}
+
+// IPFamilyOfCIDR returns the IP family of cidr.
```
thockin, line 118: ...or else IPFamilyUnknown

net/ipfamily.go
```diff
@@ -0,0 +1,180 @@
...
+func IPFamilyOfCIDR(cidr *net.IPNet) IPFamily {
+	if cidr == nil {
+		return IPFamilyUnknown
+	}
+	return IPFamilyOf(cidr.IP)
+}
+
+// IPFamilyOfCIDRString returns the IP family of cidr.
```
thockin, line 126: ...or else IPFamilyUnknown

net/ipfamily.go
```diff
@@ -0,0 +1,180 @@
...
+func IsDualStackCIDRStrings(cidrs []string) (bool, error) {
+	parsedCIDRs, err := ParseCIDRs(cidrs)
+	if err != nil {
+		return false, err
+	}
+	return IsDualStackCIDRs(parsedCIDRs)
+}
+
+// IPFamilyOf returns the IP family of netIP.
+func IPFamilyOf(netIP net.IP) IPFamily {
```
thockin, line 102: super nit: s/netIP/ip/ all over?

net/ipfamily.go
```diff
@@ -0,0 +1,180 @@
...
+	ip, _, _ := ParseCIDRSloppy(cidr)
+	return IPFamilyOf(ip)
+}
+
+// IsIPv6 returns true if netIP is IPv6 (and false if it is IPv4, nil, or invalid).
+func IsIPv6(netIP net.IP) bool {
+	return IPFamilyOf(netIP) == IPv6
+}
+
+// IsIPv6String returns true if ip contains a single IPv6 address and nothing else. (It
```
thockin, line 137: The second part of all of these does not need to be in parens.  It's just as important as the first part.

thockin (commented):
just nits!

net/ipfamily.go
```diff
@@ -0,0 +1,181 @@
...
+
+// IsDualStackIPStrings returns true if:
+// - all elements of ips can be parsed as IPs
+// - at least one IP from each family (v4 and v6) is present
+func IsDualStackIPStrings(ips []string) (bool, error) {
+	parsedIPs := make([]net.IP, 0, len(ips))
+	for i, ip := range ips {
+		parsedIP := ParseIPSloppy(ip)
+		if parsedIP == nil {
+			return false, fmt.Errorf("invalid IP[%d]: %v", i, ip)
```
thockin, line 63: super nit: s/%v/%q - if they pass "" the error is poor.

net/net.go
```diff
@@ -29,138 +29,16 @@ import (
 // order is maintained
 func ParseCIDRs(cidrsString []string) ([]*net.IPNet, error) {
 	cidrs := make([]*net.IPNet, 0, len(cidrsString))
-	for _, cidrString := range cidrsString {
+	for i, cidrString := range cidrsString {
 		_, cidr, err := ParseCIDRSloppy(cidrString)
 		if err != nil {
-			return nil, fmt.Errorf("failed to parse cidr value:%q with error:%v", cidrString, err)
+			return nil, fmt.Errorf("invalid CIDR[%d]: %v (%v)", i, cidr, err)
```
thockin, line 35: %q for `cidr`

net/ipfamily.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// IPFamilyOfCIDRString returns the IP family of cidr.
+func IPFamilyOfCIDRString(cidr string) IPFamily {
+	ip, _, _ := ParseCIDRSloppy(cidr)
+	return IPFamilyOf(ip)
+}
+
+// IsIPv6 returns true if netIP is IPv6 (and false if it is IPv4, nil, or invalid).
+func IsIPv6(netIP net.IP) bool {
```
thockin, line 134: s/netIP/ip

net/ipfamily.go
```diff
@@ -0,0 +1,181 @@
...
+
+// IsIPv6CIDRString returns true if cidr contains a single IPv6 CIDR and nothing else. It
+// returns false if cidr is an empty string, an IPv4 CIDR, or anything else that is not a
+// single valid IPv6 CIDR.
+func IsIPv6CIDRString(cidr string) bool {
+	return IPFamilyOfCIDRString(cidr) == IPv6
+}
+
+// IsIPv4 returns true if netIP is IPv4 (and false if it is IPv6, nil, or invalid).
+func IsIPv4(netIP net.IP) bool {
```
thockin, line 159: s/netIP/ip

thockin (commented):
/approve
/lgtm

nits can be followups - minor

PR #274, opened 2023-02-02
Make FakeExec safe for concurrency

artemvmin:
**What type of PR is this?**
/kind feature

This PR adds a mutex to the FakeExec struct.

The real executor shares no state between [osexec.Command()](https://github.com/kubernetes/utils/blob/391b47cb40296348795a867f38db2827a3099ce0/exec/exec.go#L102) calls. Meanwhile, [FakeExec](https://github.com/kubernetes/utils/blob/391b47cb40296348795a867f38db2827a3099ce0/exec/testing/fake_exec.go#L48) shares all the state stored on the struct.

I am working with a CSI driver that uses a single object for all provisioning calls. This change will allow me to test its behavior when calls are made concurrently.

This change is backwards compatible, since the existing behavior is undefined (and nondeterministic) for concurrent calls.

**Release note**:
```
The FakeExec test double is now safe for concurrency.
```

### DIFF
--- a/exec/testing/fake_exec.go
+++ b/exec/testing/fake_exec.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"sync"
 
 	"k8s.io/utils/exec"
 )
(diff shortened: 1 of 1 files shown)


exec/testing/fake_exec.go
```diff
@@ -81,6 +87,9 @@ func (fake *FakeExec) CommandContext(ctx context.Context, cmd string, args ...st
 
 // LookPath is for finding the path of a file
 func (fake *FakeExec) LookPath(file string) (string, error) {
+	fake.mu.Lock()
```
apelisse, line 90: Can this lead to a lock ordering inversion or even just a recursive lock?

artemvmin: `Command()` and `LookPath()` don't explicitly call each other. This means that once either function acquires a lock, it will be able to finish and unlock. Still, the caller could embed one method inside the other, which would cause deadlock.

I went ahead and tightened the lock to only the section violating the invariant, i.e. when CommandCalls is incremented and CommandScript accessed non-atomically.

There's still the implication that the fields in the fake object shouldn't be modified while it's being used in a concurrent context, but there are easier ways to make a brittle test.

apelisse: Thanks, I would also at least mention this in the documentation for the function that the risk exists.

artemvmin: Good idea. Done.

apelisse:
/approve

@msau42 do you mind re-applying lgtm?

Thanks!

PR #280, opened 2023-05-10
add generic methods for integer package

logicalhan:
**What type of PR is this?**

> /kind cleanup

Generics were introduced in go 1.18, we're at 1.20 so we can start using them in the utils package.

**Release note**:
```
IntMax,IntMin,Int32Max,Int32Min,Int64Max,Int64Min are now deprecated, please use `Max` or `Min` instead.
```

### DIFF
--- a/integer/integer.go
+++ b/integer/integer.go
@@ -16,7 +16,29 @@ limitations under the License.
 
 package integer
 
+// number is a constraint that permits any signed type, or any unsigned type
+type number interface {
+	~int | ~int8 | ~int16 | ~int32 | ~int64 | ~uint | ~uint8 | ~uint16 | ~uint32 | ~uint64 | ~uintptr
+}
+
+// Max returns the maximum of the params
+func Max[E number](a, b E) E {
+	if b > a {
+		return b
+	}
+	return a
+}
+
+// Min returns the minimum of the params
+func Min[E number](a, b E) E {
+	if b < a {
+		return b
+	}
+	return a
+}
+
 // IntMax returns the maximum of the params
+// Deprecated: please use Max
 func IntMax(a, b int) int {
 	if b > a {
 		return b
@@ -25,6 +47,7 @@ func IntMax(a, b int) int {
 }
 
 // IntMin returns the minimum of the params
+// Deprecated: please use Min
 func IntMin(a, b int) int {
 	if b < a {
 		return b
@@ -33,6 +56,7 @@ func IntMin(a, b int) int {
 }
 
 // Int32Max returns the maximum of the params
+// Deprecated: please use Max
 func Int32Max(a, b int32) int32 {
 	if b > a {
 		return b
@@ -41,6 +65,7 @@ func Int32Max(a, b int32) int32 {
 }
 
 // Int32Min returns the minimum of the params
+// Deprecated: please use Min
 func Int32Min(a, b int32) int32 {
 	if b < a {
 		return b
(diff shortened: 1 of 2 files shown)


logicalhan:
/retest

(I'm a bit confused by this test failure)

liggitt:
this may be the first time apidiff has run since https://github.com/kubernetes/utils/pull/243 merged, and I'm not sure it is handling diffing of generic APIs correctly :-/

it was happy when the generic APIs were new, but looks like it fails to understand a no-op to them

liggitt:
looks like apidiff added generics support in https://go-review.googlesource.com/c/exp/+/411076 but I'm not sure if this is exposing an edge it didn't handle

logicalhan:
> this may be the first time apidiff has run since https://github.com/kubernetes/utils/pull/243 merged, and I'm not sure it is handling diffing of generic APIs correctly :-/

Ha, I was just looking at apidiff code, it must be a bug.

integer/integer.go
```diff
@@ -16,7 +16,29 @@ limitations under the License.
 
 package integer
 
+// number is a constraint that permits any floating-point type, any signed type, or any unsigned type
+type number interface {
+	~float32 | ~float64 | ~int | ~int8 | ~int16 | ~int32 | ~int64 | ~uint | ~uint8 | ~uint16 | ~uint32 | ~uint64 | ~uintptr
```
thockin, line 21: Why do we handle float types in package `integer` ?

"numeric" ?  "mathutil"?  Something else?  Or don't handle float?

thockin: Another option is to make a new package and just alias - `var IntMax = numeric.Max[int]` if you want to keep floats

integer/integer_test.go
```diff
@@ -242,3 +245,27 @@ func TestRoundToInt32(t *testing.T) {
 		}
 	}
 }
+
+func TestMin(t *testing.T) {
+	if min := Min(math.MinInt, 0); min != math.MinInt {
```
thockin, line 250: Should we have Uint test cases?  I know it's kind of silly...

thockin (commented):
Thanks!

/lgtm
/approve

/hold 

if you want to add uint cases, else remove hold

integer/integer_test.go
logicalhan: Eh, doesn't hurt. Added.

integer/integer.go
logicalhan: Nah, I think this is probably fine. If we needed it, I imagine someone would have created a float library by now.

thockin:
Thanks!

/lgtm
/approve

integer/integer.go
```diff
@@ -16,7 +16,29 @@ limitations under the License.
 
 package integer
 
+// number is a constraint that permits any signed type, or any unsigned type
+type number interface {
+	~int | ~int8 | ~int16 | ~int32 | ~int64 | ~uint | ~uint8 | ~uint16 | ~uint32 | ~uint64 | ~uintptr
+}
```
MadhavJivrajani, line 22: Introducing a user-defined constraint here will continue breaking CI till https://github.com/golang/go/issues/60911 is fixed.
See: https://github.com/kubernetes/utils/issues/284

MadhavJivrajani (changes requested):
/hold

MadhavJivrajani:
@logicalhan -- I see that GH workflows can't be retriggered after 30 days of its initial run, I think you'll have to push an empty commit to retrigger a run here :/
The job is back on its feet fwiw.

logicalhan:
This PR is going to be obsolete actually due to https://tip.golang.org/ref/spec#Min_and_max.

PR #281, opened 2023-05-31
feat: Add LockWithContext method to keymutex

DerekTBrown:
**What type of PR is this?**
 /kind feature

- As implemented, `KeyMutex` doesn't provide cancellation/timeout functionality. This results in poor behavior (unnecessary lock contention), and increases the complexity of the required caller implementation (caller must determine whether to immediately return lock if context is cancelled, etc). 
- This PR introduces a `LockWithContext` method to `KeyMutex`, which allows callers to more gracefully implement acquisition cancellation/timeouts.
- In order to implement this behavior, this PR migrates the underlying synchronization primitive from a `Mutex` to a `chan`.  This is required because `Mutex` doesn't have support for `select` or other conditional acquisition (timeouts) short of polling.

This is a new feature. Please let me know if I need to create a corresponding issue.

**Release note**:
```

```

### DIFF
--- a/keymutex/hashed.go
+++ b/keymutex/hashed.go
@@ -56,7 +56,7 @@ func (km *hashedKeyMutex) LockKey(id string) {
 
 // Tries to acquire the associated lock within the timeframe,
 // returns true if lock is acquired, false otherwise.
-func (km *hashedKeyMutex) LockKeyWithContext(id string, ctx context.Context) bool {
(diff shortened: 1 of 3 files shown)


DerekTBrown:
@MadhavJivrajani I updated the PR to pass the linter. `apidiff` is failing only because I am adding a public API:

```
FAIL: k8s.io/utils/keymutex contains incompatible changes:
- KeyMutex.LockKeyWithContext: added
```

What are the next steps for reviewing/approving this change?

Issue #284, opened 2023-06-21
Broken CI job: apidiff


MadhavJivrajani:
The apidiff job is broken since the merge of https://github.com/kubernetes/utils/pull/243

I don't think its anything to do with the PR itself, but more to do with how apidiff handles things.
I've filed an issue for this with a potential root cause: https://github.com/golang/go/issues/60911

One of 2 can be done here:
- We wait for the release branches to be bumped to go1.20 (which I think is in progress https://github.com/kubernetes/release/issues/2815#issuecomment-1560435556) and then bump `utils` to go1.20 and then update `sets` to use `comparable`.
- Revert https://github.com/kubernetes/utils/pull/243 till we bump to go1.20 or apidiff gets a fix that we can consume (whichever happens first).

IMO we should revert for now to ensure the CI is in good shape for when we need to fix something here to mitigate a test flake/failure in something like k/k.

/cc @logicalhan 
/cc @cpanato (in case you have an estimate yet on the go1.20 bumps of release branches)

MadhavJivrajani:
There is a fix open for `apidiff`: https://go-review.googlesource.com/c/exp/+/505276
On installing `apidiff` with this patch, it seems to fix the problem 🎉 
```
❯ make verify-apidiff
GO111MODULE=on ./hack/verify-apidiff.sh -r master
Inspecting API of master...
Comparing with master...
Ignoring internal package k8s.io/utils/internal/third_party/forked/golang/golang-lru
Ignoring internal package k8s.io/utils/internal/third_party/forked/golang/net
❯ echo $?
0
```

dims:
thanks for chasing this @MadhavJivrajani

MadhavJivrajani:
https://github.com/golang/go/issues/60911 is resolved now 🎉 

Created https://github.com/kubernetes/test-infra/pull/30048
Once merged, we can start retriggering the workflows here.

MadhavJivrajani:
/close
Job is back on its feet.

PR #285, opened 2023-06-21
Revert "add generic sets to utils"

MadhavJivrajani:
Reverts kubernetes/utils#243
On account of https://github.com/kubernetes/utils/issues/284

/hold


MadhavJivrajani:
/close 
just realised can't revert since we've merged the API in already

PR #286, opened 2023-07-07
Ensure set.Equal() compares the two set lengths

skitt:
**What type of PR is this?**

/kind bug

set.Equal() currently checks the length of the owning set against itself, so a.Equal(b) returns true if a is a superset of b, with no further restriction.

This fixes the length check and adds unit tests to catch this case.

Fixes #

**Release note**:
```

```


logicalhan (approved):
/lgtm
/approve

PR #294, opened 2023-11-23
Align set with k/k apimachinery

skitt:
**What type of PR is this?**

/kind cleanup

This aligns the features of generic Set with the version in apimachinery. The main change is that sets can be instantiated with any comparable type, not only ordered types; as a result, SortedList() can no longer be implemented as a method, and is replaced by the List() function to match apimachinery.

Fixes #

**Release note**:
```
sets.Set accepts any comparable type; the sets.SortedList method is replaced with a sets.List function, to match the apimachinery implementation, and to allow unordered sets to be converted to ordered slices.
```


logicalhan (approved):
/lgtm
/approve

skitt:
Thanks @logicalhan! There’s a failure in apidiff but I don’t think it’s significant — the changes are expected (or bugs in apidiff).

logicalhan:
> Thanks @logicalhan! There’s a failure in apidiff but I don’t think it’s significant — the changes are expected (or bugs in apidiff).

It is significant because we don't know who's using this.

skitt:
> It is significant because we don't know who's using this.

The changes are listed as follows:

```
- KeySet: changed from func(map[E]A) Set[E] to func(map[E]A) Set[E]
- New: changed from func(...E) Set[E] to func(...E) Set[E]
- Set: changed from Set[E ordered] to Set[E comparable]
```

The first two aren’t changes. The last one is fine because comparable is a superset of ordered, so any existing `Set` will still build.

skitt:
The `apidiff` failure is

```
FAIL: k8s.io/utils/set contains incompatible changes:
- KeySet: changed from func(map[E]A) Set[E] to func(map[E]A) Set[E]
- New: changed from func(...E) Set[E] to func(...E) Set[E]
- Set: changed from Set[E ordered] to Set[E comparable]
```

The change it’s complaining about is the last one, which percolates up through the stack. But in practice I don’t think it’s a breaking API change: current code using `k/u` `Set` with an `ordered` type will compile fine with the new one, since an `ordered` type is also `comparable`.

set/set.go
```diff
@@ -17,19 +17,21 @@ limitations under the License.
 package set
```
pohly, line 17: The main obstacle for transitioning from k8s.io/apimachinery/pkg/util/sets to k8s.io/utils/set: the former uses plural and thus code has `sets.New`, the latter singular and thus would have to change imports to `sets "k8s.io/utils/set"` (ugly!) or change the code (also ugly!).

"set" is arguably the better name (there's only one set type here). Also, these files have all the commit history. Maybe we can create a mirror package `k8s.io/utils/sets` with a `typedef Set = set.Set` and function wrappers for the few functions here (`New`, `List`, `KeySet`)?

set/set.go
```diff
@@ -43,23 +45,24 @@ func (s Set[T]) Clear() Set[T] {
 }
 
 // KeySet creates a Set[E] from a keys of a map[E](? extends interface{}).
-func KeySet[E ordered, A any](theMap map[E]A) Set[E] {
+// If the value passed in is not actually a map, this will panic.
```
pohly, line 48: Okay, I bite: how can `theMap` not be a map?

PR #295, opened 2023-11-23
On Go 1.21+, use cmp.Ordered instead of a private type

skitt:
**What type of PR is this?**

/kind cleanup

Go 1.21 introduces a language-maintained constraint permitting any ordered types, use that instead of the private version defined in sets.

Fixes #

**Release note**:
```
NONE
```


logicalhan (approved):
/lgtm
/approve

skitt:
apidiff is failing with

```
FAIL: k8s.io/utils/set contains incompatible changes:
- KeySet: changed from func(map[E]A) Set[E] to func(map[E]A) Set[E]
- New: changed from func(...E) Set[E] to func(...E) Set[E]
- Set: changed from Set[E ordered] to Set[E cmp.Ordered]
```

The first two aren’t changes, and the latter is OK because `cmp.Ordered` is equivalent to `ordered`.

skitt:
@logicalhan what needs to be done regarding the apidiff failure?

PR #300, opened 2024-01-29
add  github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510 to shlex

pacoxu:
**What type of PR is this?**
https://github.com/google/shlex is archived, so we may maintain it here.

/kind cleanup

xref https://github.com/kubernetes-sigs/kustomize/issues/4963 and https://github.com/kubernetes/kubernetes/pull/114912

Since https://github.com/google/shlex is archived, we added it to unwanted dep list.
However, before we have requirements to change it, we may keep it as is.

This may be a fix later if we have new requirement on shlex.

**Release note**:
```
Add google/shlex that is archived
```

### DIFF
--- a/shlex/LICENSE
+++ b/shlex/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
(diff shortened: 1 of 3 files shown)


pacoxu:
/close
Since https://github.com/google/shlex is archived, we added it to unwanted dep list.
However, before we have requirements to change it, we may keep it as is.

This may be a fix later if we have new requirement on shlex.

PR #303, opened 2024-03-03
add generic implementation of bidirectional map

logicalhan:
furthermore, this PR extracts out the generic interfaces we use in the generic set implementation so that we can reuse them in the bidirectional map.

**What type of PR is this?**

/kind feature

Generics provide a path to implementing generic implementations of arbitrary (but useful) data-structures. This PR creates a bidirectional map.

Fixes #

**Release note**:
```
k8s/utils now supports a generic implementation of a bidirectional map.
```

### DIFF
--- a/bidirectionalmap/map.go
+++ b/bidirectionalmap/map.go
@@ -1,0 +1,107 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package bidirectionalmap
+
+import (
+	"k8s.io/utils/genericinterfaces"
+	"k8s.io/utils/set"
+)
+
+// BidirectionalMap is a bidirectional map.
+type BidirectionalMap[X genericinterfaces.Ordered, Y genericinterfaces.Ordered] struct {
+	right map[X]set.Set[Y]
+	left  map[Y]set.Set[X]
+}
+
+// NewBidirectionalMap creates a new BidirectionalMap.
+func NewBidirectionalMap[X genericinterfaces.Ordered, Y genericinterfaces.Ordered]() *BidirectionalMap[X, Y] {
+	return &BidirectionalMap[X, Y]{
+		right: make(map[X]set.Set[Y]),
+		left:  make(map[Y]set.Set[X]),
+	}
+}
+
+// InsertRight inserts a new item into the right map, return true if the key-value was not already
+// present in the map, false otherwise
+func (bdm *BidirectionalMap[X, Y]) InsertRight(x X, y Y) bool {
+	if bdm.right[x] == nil {
+		bdm.right[x] = set.New[Y]()
+	}
+	if bdm.right[x].Has(y) {
+		return false
+	}
+	if bdm.left[y] == nil {
+		bdm.left[y] = set.New[X]()
+	}
+	bdm.right[x].Insert(y)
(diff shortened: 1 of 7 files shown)


logicalhan:
I don't actually think the apidiff compatibility check is valid here. Any concrete usage of the set library should continue to work, since the interface itself has not changed it's underlying signature.

bidirectionalmap/map.go
```diff
@@ -0,0 +1,116 @@
...
+*/
+
+package bidirectionalmap
+
+import (
+	"k8s.io/utils/genericinterfaces"
+	"k8s.io/utils/set"
+)
+
+// BidirectionalMap is a bidirectional map.
```
thockin, line 24: I would flesh this out more.  What is the semantic, what guarantees does it offer or not offer, what is it for (a real example)?  Also explaion which is the left-key and right-value vs. the left-value and right-key.

Naively I would expect BiDiMap[X, Y] to mean map[X]Y + Map[Y]X, but this seems to be the opposite

bidirectionalmap/map.go
```diff
@@ -0,0 +1,116 @@
...
+
+// NewBidirectionalMap creates a new BidirectionalMap.
+func NewBidirectionalMap[X genericinterfaces.Ordered, Y genericinterfaces.Ordered]() *BidirectionalMap[X, Y] {
+	return &BidirectionalMap[X, Y]{
+		right: make(map[X]set.Set[Y]),
+		left:  make(map[Y]set.Set[X]),
+	}
+}
+
+// InsertRight inserts a new item into the right map.
```
thockin, line 38: Explain the return value?

bidirectionalmap/map_test.go
```diff
@@ -0,0 +1,46 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package bidirectionalmap
+
+import "testing"
+
+func TestMultipleInserts(t *testing.T) {
+	bidimap := NewBidirectionalMap[string, string]()
```
thockin, line 22: Examples with different types would better illustrate.

You could add an example_test.go to get a nice godoc example

thockin (commented):
Without a concrete use case in k/k I don't know if we want to check this in.  Do we have one?

I'm not sure on names.  Having a pkg per type feels bad, but it's not clear what package this should be in.

Go has `strings` and `slices`.  We have `set` (which should probably be `sets`).  Maybe `maps` ?

Would `BiMap` be sufficient?  Just looking for shorter.  Not critical

bidirectionalmap/map.go
logicalhan: I'm not sure what you mean by opposite, `map[X]Y + Map[Y]X` can be invalidated by have multiple keys in `map[x]y` map to the same y value. e.g.:

```
a = map[string]string
a["a"] = "1"
a["b"] = "1"
```
The reverse mapping must return a list of the keys of map `a` otherwise we have a lossy map.

Therefore, a bidirectional map must be two maps, each of which map to a list (or set) of values.

bidirectionalmap/map.go
logicalhan: Will do.

set/set.go
```diff
@@ -18,24 +18,26 @@ package set
...
-func New[E ordered](items ...E) Set[E] {
+func New[E genericinterfaces.Ordered](items ...E) Set[E] {
 	ss := Set[E]{}
 	ss.Insert(items...)
 	return ss
 }
 
 // KeySet creates a Set[E] from a keys of a map[E](? extends interface{}).
-func KeySet[E ordered, A any](theMap map[E]A) Set[E] {
+func KeySet[E genericinterfaces.Ordered, A any](theMap map[E]A) Set[E] {
```
liggitt, line 40: What's the motivation for exporting the type? are we expecting other things to reference the interface? They couldn't previously, which had the benefit of only allowing people to declare KeySet with their own types, which had to be compatible with our constraints... that seemed reasonable to me.

logicalhan: I'd need to redeclare the exact same type in the maps package otherwise, since the output of the value of a map is a generic set.

skitt: How about waiting for the bump to Go 1.21 in utils and relying on `cmp.Ordered`? See also #295.

PR #304, opened 2024-03-05
.: bump go version to go1.21

MadhavJivrajani:
**What type of PR is this?**
/kind cleanup

go1.20 is out of support as of ~3 weeks ago and all supported k/k release branches are on go1.21.
This PR bumps version of Go to 1.21.

This PR also cleans up the build tag directives and consistently uses only `//go:build` now

Fixes #

@dims @liggitt - I'm wondering if we put in `go1.21.x` in the `go.mod` or an overall release version like `go1.21` should do like we do in k/k?

**Release note**:
```
NONE
```

/assign @dims @liggitt

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module k8s.io/utils
 
-go 1.18
+go 1.21
 
 require (
 	github.com/davecgh/go-spew v1.1.1
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,7 +4,7 @@ jobs:
   build:
     strategy:
       matrix:
-        go-version: [1.20.x, 1.21.x]
+        go-version: [1.21.x, 1.22.x]
         platform: [windows-latest]
     runs-on: ${{ matrix.platform }}
     steps:
@@ -24,7 +24,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.20.x, 1.21.x]
+        go-version: [1.21.x, 1.22.x]
         platform: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.platform }}
     steps:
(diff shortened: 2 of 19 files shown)


MadhavJivrajani:
@logicalhan FYI ^
for your work on the set implementations using built in constraints rather than defining our own.

dims:
cc @xdu31

go.mod
```diff
@@ -1,6 +1,6 @@
 module k8s.io/utils
 
-go 1.18
+go 1.21
```
liggitt, line 3: /hold

This doesn't do what you think it does :-/

This forces anything importing this module to also bump its own `go` module directive to go1.21, and makes go change the defaults for godebug compatibility switch defaults. https://github.com/golang/go/issues/65573 is an open proposal to improve this in go1.23, but until that lands, we need to avoid bumping our go directives past go1.20 in our release-from-HEAD repos like utils/kube-openapi/gengo/etc.

Because this repo only releases from HEAD, if we find a bug we need to fix and backport to older kubernetes versions (like 1.27), this would bump the go.mod version in 1.27 from go1.20 to go1.21, which we don't want. Repos that have release branches (like k/k) are more reasonable to keep pushing the go.mod to latest go at HEAD.

golang.org/x/... modules also keep their go directive at the oldest language version you can build with (e.g. https://cs.opensource.google/go/x/sys/+/master:go.mod has 1.18).

liggitt:
> all supported k/k release branches are on go1.21.

they *build* with go1.21 now, but 1.27 and 1.28 are relying on go1.20-compatible behavior via their `go` directive in their go.mod file, which this would force higher if we ever had to pick up a fix from this repo in those branches.

MadhavJivrajani:
Ahhh right
I see what you mean, the decoupling godebug switches proposal that's ongoing. Makes sense.

We can revisit this once we know what happens there, thanks @liggitt!
/close

PR #306, opened 2024-03-06
.*: add presubmit check for verifying go directive changes

MadhavJivrajani:
This commit adds a script that checks the changed version in the go.mod file with a certain maximum version that the go directive can have.

We set the maximum version of the go directive as 1.20 here because the oldest go directive that exists on our supported release branches in k/k is 1.20.

This commit additionally changes the requisite GH action to incorporate this check.

Fixes https://github.com/kubernetes/utils/issues/305
xref https://github.com/kubernetes/kubernetes/issues/123744

/sig architecture 
/area code-organization
/kind cleanup
/assign @dims @liggitt

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,6 +37,22 @@ jobs:
     - name: Test
       run: |
         make test
+  verify-go-directive:
+    strategy:
+      matrix:
+        go-version: [1.20.x, 1.21.x]
+        platform: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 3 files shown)


MadhavJivrajani:
Successful run in the normal case: https://github.com/kubernetes/utils/actions/runs/8169957204/job/22335081807?pr=306
Failing run in invalid case when I try and change go.mod to 1.21: https://github.com/kubernetes/utils/actions/runs/8169981509/job/22335158126?pr=306

PR #307, opened 2024-04-23
Add dependabot

lucacome:
**What type of PR is this?**

/kind cleanup

Adds dependabot to keep dependencies up to date. Most of the GitHub Actions are out of date and even have warnings in all the runs.

Fixes #

**Release note**:
```

```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily
+
+  - package-ecosystem: gomod
+    directory: /
+    schedule:
+      interval: daily


lucacome:
@liggitt what do you think about this?

liggitt:
I don't really think we need to add it for this repo

* we don't build binaries out of this repository
* we only have two direct dependencies, which are both slow moving and expected to be API stable, so any downstream consumer should be able to use whatever version they want
* this repo has an explicit goal of not picking up any new dependencies: "Very few (ideally zero) external dependencies."

lucacome:
Let me know if I should close the PR, or change it to only update GitHub Actions. 
Personally I don't think it hurts to keep everything up to date (even just to get a PR so we know there's an update), but I'm happy to change it.

PR #308, opened 2024-04-23
Update Go versions in workflows

lucacome:
**What type of PR is this?**

/kind cleanup

Updates the version of go the minimum supported.

Fixes #

**Release note**:
```

```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,88 +4,108 @@ jobs:
   build:
     strategy:
       matrix:
-        go-version: [1.20.x, 1.21.x]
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module k8s.io/utils
 
-go 1.18
+go 1.21.0
```
liggitt, line 3: leave this at go1.18, bumping here has side effects we don't want (https://github.com/kubernetes/kubernetes/issues/123744)

liggitt: cc @MadhavJivrajani

PR #310, opened 2024-05-20
add netip utils

aojea:
/kind feature

During the new feature added to the apiserver to handle multiple ServiceCIDR that new netip golang library was used , see https://github.com/kubernetes/kubernetes/pull/122888/files

There are two helper functions that are used in several places and that can be better be part of the utils library:

- IPToAddr: converts a net.IP to a netip.Addr
- BroadcastAddress: returns the broadcast address of a given netip.Prefix, useful to check if given an IP this is a HostIP, i.e. it is contained in the prefix and is not the subnet address or the broadcast address (for IPv4)

### DIFF
--- a/net/netip.go
+++ b/net/netip.go
@@ -1,0 +1,68 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package net
+
+import (
+	"fmt"
+	"net"
+	"net/netip"
+)
+
+// IPToAddr converts a net.IP to a netip.Addr
+// if the net.IP is not valid it returns an empty netip.Addr{}
+func IPToAddr(ip net.IP) netip.Addr {
+	// https://pkg.go.dev/net/netip#AddrFromSlice can return an IPv4 in IPv6 format
+	// so we have to check the IP family to return exactly the format that we want
+	// address, _ := netip.AddrFromSlice(net.ParseIPSloppy(192.168.0.1)) returns
+	// an address like ::ffff:192.168.0.1/32
+	bytes := ip.To4()
+	if bytes == nil {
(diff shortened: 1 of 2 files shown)


danwinship:
See also https://github.com/kubernetes/kubernetes/pull/122549

I was pushing for apimachinery rather than utils there, though I've been starting to lean away from that as I continue trying to port things to the new APIs.

I'm thinking that regardless of what happens, we will probably want to deprecate the IP-handling functions in `k8s.io/utils/net` (possibly in favor of `k8s.io/utils/net/v2`? or if not, `k8s.io/utils/ip`), because the existing functions/types are squatting on useful names that limit our ability to have a fully-thought-out and well-designed combined set of `net` and `netip` APIs. So from that POV, I guess merging this PR wouldn't really get in the way of my plans (it just adds more APIs to deprecate), but maybe we should take a step back and think about this all first instead?

Current thoughts:
- We want to start moving to the `netip` types, but we will not be able to get rid of the `net` types any time soon (especially because of vendored libraries that still use the `net` types). Thus, we need a full set of good/safe APIs for both sets of types, plus APIs to efficiently convert between them.
- We can use generics to hide a lot of the `net`-vs-`netip`, duplication, and even simplify some of the existing netutils APIs.
- `netutils.IPFamilyOf()` would be twice as useful if it returned a type that could be cast to `v1.IPFamily` (currently `netutils.IPv4 == "4"` but `v1.IPv4Protocol == "IPv4"`), and would be ten times as useful if it returned a type that actually typechecked as the same type as `v1.IPFamily`. (This is possible; see https://github.com/kubernetes/kubernetes/pull/122549/commits/cfc6ca6d9574135da68e467807a3e0c2c8a390f8.)
- In 122549 I was pushing for putting the utilities in apimachinery, but keeping them in utils actually works fine.

net/netip.go
```diff
@@ -0,0 +1,68 @@
...
+	address, _ := netip.AddrFromSlice(bytes)
+	return address
+}
+
+// BroadcastAddress returns the broadcast address of the subnet
+// The broadcast address is obtained by setting all the host bits
+// in a subnet to 1.
+// network 192.168.0.0/24 : subnet bits 24 host bits 32 - 24 = 8
+// broadcast address 192.168.0.255
+func BroadcastAddress(subnet netip.Prefix) (netip.Addr, error) {
```
danwinship, line 46: do we really do this enough in k8s to need a shared utility function for it?

aojea:
> See also [kubernetes/kubernetes#122549](https://github.com/kubernetes/kubernetes/pull/122549)

>  So from that POV, I guess merging this PR wouldn't really get in the way of my plans (it just adds more APIs to deprecate), but maybe we should take a step back and think about this all first instead?

let´s not  add more tech debt the

PR #311, opened 2024-05-25
net: add multi listener impl for net.Listener

aroradaman:
This adds an implementation of net.Listener which uses synchronous multiplexing to listen on and accept connections from multiple addrs.

**What type of PR is this?**
/kind feature

This PR adds an implementation of **net.Listener** which uses synchronous multiplexing to listen to and accept connections from multiple addresses.

Fixes #

**Release note**:
```
NONE
```

### DIFF
--- a/net/multi_listen.go
+++ b/net/multi_listen.go
@@ -1,0 +1,195 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package net
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"sync"
+)
+
+// connErrPair pairs conn and error which is returned by accept on sub-listeners.
+type connErrPair struct {
+	conn net.Conn
+	err  error
+}
+
+// multiListener implements net.Listener
+type multiListener struct {
+	listeners []net.Listener
+	wg        sync.WaitGroup
+
+	// connCh passes accepted connections, from child listeners to parent.
+	connCh chan connErrPair
+	// stopCh communicates from parent to child listeners.
+	stopCh chan struct{}
+}
+
+// compile time check to ensure *multiListener implements net.Listener
+var _ net.Listener = &multiListener{}
+
+// MultiListen returns net.Listener which can listen on and accept connections for
+// the given network on multiple addresses. Internally it uses stdlib to create
+// sub-listener and multiplexes connection requests using go-routines.
+// The network must be "tcp", "tcp4" or "tcp6".
+// It follows the semantics of net.Listen that primarily means:
+//  1. If the host is an unspecified/zero IP address with "tcp" network, MultiListen
+//     listens on all available unicast and anycast IP addresses of the local system.
+//  2. Use "tcp4" or "tcp6" to exclusively listen on IPv4 or IPv6 family, respectively.
+//  3. The host can accept names (e.g, localhost) and it will create a listener for at
+//     most one of the host's IP.
+func MultiListen(ctx context.Context, network string, addrs ...string) (net.Listener, error) {
+	var lc net.ListenConfig
+	return multiListen(
+		ctx,
+		network,
+		addrs,
+		func(ctx context.Context, network, address string) (net.Listener, error) {
+			return lc.Listen(ctx, network, address)
+		})
+}
+
+// multiListen implements MultiListen by consuming stdlib functions as dependency allowing
+// mocking for unit-testing.
+func multiListen(
+	ctx context.Context,
+	network string,
+	addrs []string,
+	listenFunc func(ctx context.Context, network, address string) (net.Listener, error),
+) (net.Listener, error) {
+	if !(network == "tcp" || network == "tcp4" || network == "tcp6") {
+		return nil, fmt.Errorf("network %q not supported", network)
+	}
+	if len(addrs) == 0 {
+		return nil, fmt.Errorf("no address provided to listen on")
+	}
+
+	ml := &multiListener{
+		connCh: make(chan connErrPair),
+		stopCh: make(chan struct{}),
+	}
+	for _, addr := range addrs {
+		l, err := listenFunc(ctx, network, addr)
+		if err != nil {
+			// close all the sub-listeners and exit
+			_ = ml.Close()
+			return nil, err
+		}
+		ml.listeners = append(ml.listeners, l)
+	}
+
+	for _, l := range ml.listeners {
+		ml.wg.Add(1)
+		go func(l net.Listener) {
+			defer ml.wg.Done()
+			for {
+				// Accept() is blocking, unless ml.Close() is called, in which
+				// case it will return immediately with an error.
+				conn, err := l.Accept()
+				// This assumes that ANY error from Accept() will terminate the
+				// sub-listener. We could maybe be more precise, but it
+				// doesn't seem necessary.
+				terminate := err != nil
+
+				select {
+				case ml.connCh <- connErrPair{conn: conn, err: err}:
+				case <-ml.stopCh:
+					// In case we accepted a connection AND were stopped, and
+					// this select-case was chosen, just throw away the
+					// connection.  This avoids potentially blocking on connCh
+					// or leaking a connection.
+					if conn != nil {
+						_ = conn.Close()
+					}
+					terminate = true
+				}
+				// Make sure we don't loop on Accept() returning an error and
+				// the select choosing the channel case.
+				if terminate {
+					return
+				}
+			}
+		}(l)
+	}
+	return ml, nil
+}
+
+// Accept implements net.Listener. It waits for and returns a connection from
+// any of the sub-listener.
+func (ml *multiListener) Accept() (net.Conn, error) {
+	// wait for any sub-listener to enqueue an accepted connection
+	connErr, ok := <-ml.connCh
+	if !ok {
+		// The channel will be closed only when Close() is called on the
+		// multiListener. Closing of this channel implies that all
+		// sub-listeners are also closed, which causes a "use of closed
+		// network connection" error on their Accept() calls. We return the
+		// same error for multiListener.Accept() if multiListener.Close()
+		// has already been called.
+		return nil, fmt.Errorf("use of closed network connection")
+	}
+	return connErr.conn, connErr.err
+}
+
+// Close implements net.Listener. It will close all sub-listeners and wait for
+// the go-routines to exit.
+func (ml *multiListener) Close() error {
+	// Make sure this can be called repeatedly without explosions.
+	select {
+	case <-ml.stopCh:
+		return fmt.Errorf("use of closed network connection")
+	default:
+	}
+
+	// Tell all sub-listeners to stop.
+	close(ml.stopCh)
+
+	// Closing the listeners causes Accept() to immediately return an error in
+	// the sub-listener go-routines.
+	for _, l := range ml.listeners {
+		_ = l.Close()
+	}
+
+	// Wait for all the sub-listener go-routines to exit.
+	ml.wg.Wait()
+	close(ml.connCh)
+
+	// Drain any already-queued connections.
+	for connErr := range ml.connCh {
+		if connErr.conn != nil {
+			_ = connErr.conn.Close()
+		}
+	}
+	return nil
+}
+
+// Addr is an implementation of the net.Listener interface.  It always returns
+// the address of the first listener.  Callers should  use conn.LocalAddr() to
+// obtain the actual local address of the sub-listener.
+func (ml *multiListener) Addr() net.Addr {
+	return ml.listeners[0].Addr()
+}
+
+// Addrs is like Addr, but returns the address for all registered listeners.
+func (ml *multiListener) Addrs() []net.Addr {
+	var ret []net.Addr
+	for _, l := range ml.listeners {
+		ret = append(ret, l.Addr())
+	}
+	return ret
+}
--- a/net/multi_listen_test.go
+++ b/net/multi_listen_test.go
@@ -1,0 +1,498 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package net
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"strconv"
+	"sync/atomic"
+	"testing"
+	"time"
+)
+
+type fakeCon struct {
+	remoteAddr net.Addr
+}
+
+func (f *fakeCon) Read(_ []byte) (n int, err error) {
+	return 0, nil
+}
+
+func (f *fakeCon) Write(_ []byte) (n int, err error) {
+	return 0, nil
+}
+
+func (f *fakeCon) Close() error {
+	return nil
+}
+
+func (f *fakeCon) LocalAddr() net.Addr {
+	return nil
+}
+
+func (f *fakeCon) RemoteAddr() net.Addr {
+	return f.remoteAddr
+}
+
+func (f *fakeCon) SetDeadline(_ time.Time) error {
+	return nil
+}
+
+func (f *fakeCon) SetReadDeadline(_ time.Time) error {
+	return nil
+}
+
+func (f *fakeCon) SetWriteDeadline(_ time.Time) error {
+	return nil
+}
+
+var _ net.Conn = &fakeCon{}
+
+type fakeListener struct {
+	addr         net.Addr
+	index        int
+	err          error
+	closed       atomic.Bool
+	connErrPairs []connErrPair
+}
+
+func (f *fakeListener) Accept() (net.Conn, error) {
+	if f.index < len(f.connErrPairs) {
+		index := f.index
+		connErr := f.connErrPairs[index]
+		f.index++
+		return connErr.conn, connErr.err
+	}
+	for {
+		if f.closed.Load() {
+			return nil, fmt.Errorf("use of closed network connection")
+		}
+	}
+}
+
+func (f *fakeListener) Close() error {
+	f.closed.Store(true)
+	return nil
+}
+
+func (f *fakeListener) Addr() net.Addr {
+	return f.addr
+}
+
+var _ net.Listener = &fakeListener{}
+
+func listenFuncFactory(listeners []*fakeListener) func(_ context.Context, network string, address string) (net.Listener, error) {
+	index := 0
+	return func(_ context.Context, network string, address string) (net.Listener, error) {
+		if index < len(listeners) {
+			host, portStr, err := net.SplitHostPort(address)
+			if err != nil {
+				return nil, err
+			}
+			port, err := strconv.Atoi(portStr)
+			if err != nil {
+				return nil, err
+			}
+			listener := listeners[index]
+			addr := &net.TCPAddr{
+				IP:   ParseIPSloppy(host),
+				Port: port,
+			}
+			if err != nil {
+				return nil, err
+			}
+			listener.addr = addr
+			index++
+
+			if listener.err != nil {
+				return nil, listener.err
+			}
+			return listener, nil
+		}
+		return nil, nil
+	}
+}
+
+func TestMultiListen(t *testing.T) {
+	testCases := []struct {
+		name          string
+		network       string
+		addrs         []string
+		fakeListeners []*fakeListener
+		errString     string
+	}{
+		{
+			name:      "unsupported network",
+			network:   "udp",
+			errString: "network \"udp\" not supported",
+		},
+		{
+			name:      "no host",
+			network:   "tcp",
+			errString: "no address provided to listen on",
+		},
+		{
+			name:          "valid",
+			network:       "tcp",
+			addrs:         []string{"127.0.0.1:12345"},
+			fakeListeners: []*fakeListener{{connErrPairs: []connErrPair{}}},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx := context.TODO()
+			ml, err := multiListen(ctx, tc.network, tc.addrs, listenFuncFactory(tc.fakeListeners))
+
+			if tc.errString != "" {
+				assertError(t, tc.errString, err)
+			} else {
+				assertNoError(t, err)
+			}
+			if ml != nil {
+				err = ml.Close()
+				if err != nil {
+					t.Errorf("Did not expect error: %v", err)
+				}
+			}
+		})
+	}
+}
+
+func TestMultiListen_Addr(t *testing.T) {
+	ctx := context.TODO()
+	ml, err := multiListen(ctx, "tcp", []string{"10.10.10.10:5000", "192.168.1.10:5000", "127.0.0.1:5000"}, listenFuncFactory(
+		[]*fakeListener{{}, {}, {}},
+	))
+	if err != nil {
+		t.Errorf("Did not expect error: %v", err)
+	}
+
+	if ml.Addr().String() != "10.10.10.10:5000" {
+		t.Errorf("Expected '10.10.10.10:5000' but got '%s'", ml.Addr().String())
+	}
+
+	err = ml.Close()
+	if err != nil {
+		t.Errorf("Did not expect error: %v", err)
+	}
+}
+
+func TestMultiListen_Addrs(t *testing.T) {
+	ctx := context.TODO()
+	addrs := []string{"10.10.10.10:5000", "192.168.1.10:5000", "127.0.0.1:5000"}
+	ml, err := multiListen(ctx, "tcp", addrs, listenFuncFactory(
+		[]*fakeListener{{}, {}, {}},
+	))
+	if err != nil {
+		t.Errorf("Did not expect error: %v", err)
+	}
+
+	gotAddrs := ml.(*multiListener).Addrs()
+	for i := range gotAddrs {
+		if gotAddrs[i].String() != addrs[i] {
+			t.Errorf("expected %q; got %q", addrs[i], gotAddrs[i].String())
+		}
+
+	}
+
+	err = ml.Close()
+	if err != nil {
+		t.Errorf("Did not expect error: %v", err)
+	}
+}
+
+func TestMultiListen_Close(t *testing.T) {
+	testCases := []struct {
+		name          string
+		addrs         []string
+		runner        func(listener net.Listener, acceptCalls int) error
+		fakeListeners []*fakeListener
+		acceptCalls   int
+		errString     string
+	}{
+		{
+			name:  "close",
+			addrs: []string{"10.10.10.10:5000", "192.168.1.10:5000", "127.0.0.1:5000"},
+			runner: func(ml net.Listener, acceptCalls int) error {
+				for i := 0; i < acceptCalls; i++ {
+					_, err := ml.Accept()
+					if err != nil {
+						return err
+					}
+				}
+				err := ml.Close()
+				if err != nil {
+					return err
+				}
+				return nil
+			},
+			fakeListeners: []*fakeListener{{}, {}, {}},
+		},
+		{
+			name:  "close with pending connections",
+			addrs: []string{"10.10.10.10:5001", "192.168.1.10:5002", "127.0.0.1:5003"},
+			runner: func(ml net.Listener, acceptCalls int) error {
+				for i := 0; i < acceptCalls; i++ {
+					_, err := ml.Accept()
+					if err != nil {
+						return err
+					}
+				}
+				err := ml.Close()
+				if err != nil {
+					return err
+				}
+				return nil
+			},
+			fakeListeners: []*fakeListener{{
+				connErrPairs: []connErrPair{{
+					conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("10.10.10.10"), Port: 50001}},
+				}}}, {
+				connErrPairs: []connErrPair{{
+					conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("192.168.1.10"), Port: 50002}},
+				},
+				}}, {
+				connErrPairs: []connErrPair{{
+					conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("127.0.0.1"), Port: 50003}},
+				}},
+			}},
+		},
+		{
+			name:  "close with no pending connections",
+			addrs: []string{"10.10.10.10:3001", "192.168.1.10:3002", "127.0.0.1:3003"},
+			runner: func(ml net.Listener, acceptCalls int) error {
+				for i := 0; i < acceptCalls; i++ {
+					_, err := ml.Accept()
+					if err != nil {
+						return err
+					}
+				}
+				err := ml.Close()
+				if err != nil {
+					return err
+				}
+				return nil
+			},
+			fakeListeners: []*fakeListener{{
+				connErrPairs: []connErrPair{
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("10.10.10.10"), Port: 50001}}},
+				}}, {
+				connErrPairs: []connErrPair{
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("192.168.1.10"), Port: 50002}}},
+				}}, {
+				connErrPairs: []connErrPair{
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("127.0.0.1"), Port: 50003}}},
+				},
+			}},
+			acceptCalls: 3,
+		},
+		{
+			name:  "close on close",
+			addrs: []string{"10.10.10.10:5000", "192.168.1.10:5000", "127.0.0.1:5000"},
+			runner: func(ml net.Listener, acceptCalls int) error {
+				for i := 0; i < acceptCalls; i++ {
+					_, err := ml.Accept()
+					if err != nil {
+						return err
+					}
+				}
+				err := ml.Close()
+				if err != nil {
+					return err
+				}
+
+				err = ml.Close()
+				if err != nil {
+					return err
+				}
+				return nil
+			},
+			fakeListeners: []*fakeListener{{}, {}, {}},
+			errString:     "use of closed network connection",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx := context.TODO()
+			ml, err := multiListen(ctx, "tcp", tc.addrs, listenFuncFactory(tc.fakeListeners))
+			if err != nil {
+				t.Errorf("Did not expect error: %v", err)
+			}
+			err = tc.runner(ml, tc.acceptCalls)
+			if tc.errString != "" {
+				assertError(t, tc.errString, err)
+			} else {
+				assertNoError(t, err)
+			}
+
+			for _, f := range tc.fakeListeners {
+				if !f.closed.Load() {
+					t.Errorf("Expeted sub-listener to be closed")
+				}
+			}
+		})
+	}
+}
+
+func TestMultiListen_Accept(t *testing.T) {
+	testCases := []struct {
+		name          string
+		addrs         []string
+		runner        func(listener net.Listener, acceptCalls int) error
+		fakeListeners []*fakeListener
+		acceptCalls   int
+		errString     string
+	}{
+		{
+			name:  "accept all connections",
+			addrs: []string{"10.10.10.10:3000", "192.168.1.103:4000", "127.0.0.1:5000"},
+			runner: func(ml net.Listener, acceptCalls int) error {
+				for i := 0; i < acceptCalls; i++ {
+					_, err := ml.Accept()
+					if err != nil {
+						return err
+					}
+				}
+				err := ml.Close()
+				if err != nil {
+					return err
+				}
+				return nil
+			},
+			fakeListeners: []*fakeListener{{
+				connErrPairs: []connErrPair{
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("10.10.10.10"), Port: 50001}}},
+				}}, {
+				connErrPairs: []connErrPair{
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("192.168.1.10"), Port: 50002}}},
+				}}, {
+				connErrPairs: []connErrPair{
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("127.0.0.1"), Port: 50003}}},
+				},
+			}},
+			acceptCalls: 3,
+		},
+		{
+			name:  "accept some connections",
+			addrs: []string{"10.10.10.10:3000", "192.168.1.103:4000", "172.16.20.10:5000", "127.0.0.1:6000"},
+			runner: func(ml net.Listener, acceptCalls int) error {
+
+				for i := 0; i < acceptCalls; i++ {
+					_, err := ml.Accept()
+					if err != nil {
+						return err
+					}
+
+				}
+				err := ml.Close()
+				if err != nil {
+					return err
+				}
+				return nil
+			},
+			fakeListeners: []*fakeListener{{
+				connErrPairs: []connErrPair{
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("10.10.10.10"), Port: 30001}}},
+				}}, {
+				connErrPairs: []connErrPair{
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("192.168.1.10"), Port: 40001}}},
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("192.168.1.10"), Port: 40002}}},
+				}}, {
+				connErrPairs: []connErrPair{
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("172.16.20.10"), Port: 50001}}},
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("172.16.20.10"), Port: 50002}}},
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("172.16.20.10"), Port: 50003}}},
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("172.16.20.10"), Port: 50004}}},
+				}}, {
+				connErrPairs: []connErrPair{
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("127.0.0.1"), Port: 60001}}},
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("127.0.0.1"), Port: 60002}}},
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("127.0.0.1"), Port: 60003}}},
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("127.0.0.1"), Port: 60004}}},
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("127.0.0.1"), Port: 60005}}},
+				},
+			}},
+			acceptCalls: 3,
+		},
+		{
+			name:  "accept on closed listener",
+			addrs: []string{"10.10.10.10:3001", "192.168.1.10:3002", "127.0.0.1:3003"},
+			runner: func(ml net.Listener, acceptCalls int) error {
+				err := ml.Close()
+				if err != nil {
+					return err
+				}
+				for i := 0; i < acceptCalls; i++ {
+					_, err := ml.Accept()
+					if err != nil {
+						return err
+					}
+				}
+				return nil
+			},
+			fakeListeners: []*fakeListener{{
+				connErrPairs: []connErrPair{
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("10.10.10.10"), Port: 50001}}},
+				}}, {
+				connErrPairs: []connErrPair{
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("192.168.1.10"), Port: 50002}}},
+				}}, {
+				connErrPairs: []connErrPair{
+					{conn: &fakeCon{remoteAddr: &net.TCPAddr{IP: ParseIPSloppy("127.0.0.1"), Port: 50003}}},
+				},
+			}},
+			acceptCalls: 1,
+			errString:   "use of closed network connection",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx := context.TODO()
+			ml, err := multiListen(ctx, "tcp", tc.addrs, listenFuncFactory(tc.fakeListeners))
+			if err != nil {
+				t.Errorf("Did not expect error: %v", err)
+			}
+
+			err = tc.runner(ml, tc.acceptCalls)
+			if tc.errString != "" {
+				assertError(t, tc.errString, err)
+			} else {
+				assertNoError(t, err)
+			}
+		})
+	}
+}
+
+func assertError(t *testing.T, errString string, err error) {
+	if err == nil {
+		t.Errorf("Expected error '%s' but got none", errString)
+	}
+	if err.Error() != errString {
+		t.Errorf("Expected error '%s' but got '%s'", errString, err.Error())
+	}
+}
+
+func assertNoError(t *testing.T, err error) {
+	if err != nil {
+		t.Errorf("Did not expect error: %v", err)
+	}
+}


aroradaman:
For the testing part should I directly add an integration test that listens on [net.InterfaceAddrs](https://pkg.go.dev/net#InterfaceAddrs) on multiple ports and then try to connect to server for validation?

aroradaman:
For windows, we might need a goroutine-based async multiplexer as select is not supported.

net/multi_listen_darwin.go
```diff
@@ -0,0 +1,94 @@
...
+		for _, fd := range ml.fds {
+			if fd > maxfd {
+				maxfd = fd
+			}
+			addFDToFDSet(fd, readFds)
+		}
+
+		// wait for any of the sockets to be ready for accepting new connection
+		timeout := syscall.Timeval{Sec: 1, Usec: 0}
+		err := syscall.Select(maxfd+1, readFds, nil, nil, &timeout)
```
uablrek, line 44: Will this not block the OS-thread? (sorry for asking before checking)

aroradaman: Yes, it will.
**Accept()** will be called inside **http.Serve** which is a blocking call. In kube-proxy we run healthz and metrics as a separate go-routine. 

https://github.com/golang/go/blob/go1.22.3/src/net/http/server.go#L3255

aroradaman: also **Accept** is supposed to be a blocking call, right?

uablrek: Then I think it can't be done in this way. In go, (very) many go routines are mapped on a few OS-threads (equals the number of cores by default), and the OS-threads should not be blocked. Compare to the [resolver](https://pkg.go.dev/net#hdr-Name_Resolution). Google writes it's own resolver just to avoid blocking the OS-thread

uablrek: Would it be possible to use the normal net.Listen in some way? In go-routines and a channel for instance? And have a "super-Listen" on top

aroradaman: > Then I think it can't be done in this way. In go, (very) many go routines are mapped on a few OS-threads (equals the number of cores by default), and the OS-threads should not be blocked. Compare to the [resolver](https://pkg.go.dev/net#hdr-Name_Resolution). Google writes it's own resolver just to avoid blocking the OS-thread

Maybe then there is scope for improvement in how kube-proxy handles http servers. All the servers healthz, metrics, health-check servers for services use stdlib listener which uses **Accept** which is a blocking call.

aroradaman: > Would it be possible to use the normal net.Listen in some way? In go-routines and a channel for instance? And have a "super-Listen" on top

I initially planned to do that only, using go-routines to listen on every addr but then using select to mulitplex without spawning any go-routine looked a better approach to me.

uablrek: > All the servers healthz, metrics, health-check servers for services use stdlib listener which uses Accept which is a blocking call.

Are they using `syscall.Accept`?! If so, that's a bug.

uablrek: > I initially planned to do that only, using go-routines to listen on every addr but then using select to mulitplex without spawning any go-routine looked a better approach to me.

Go-routines are cheap, and I think you can get away with just one channel which makes the code simpler. Also, I think your `Listen` should take a `context` as parameter. net.Listen uses [ListenConfig](https://pkg.go.dev/net#ListenConfig) which feels like a work-around for keeping backward compatibility.

uablrek: `net.Listener.Accept()` (simplified syntax) only block the go-routine, not the OS-thread

uablrek: `l.Accept()` will block the go-routine (but not any os-thread), when a client connects it will un-block and send the conn on the channel, which will un-block multiListener.Accept. Lgtm (except that l.Accept returns (conn, err) which I assume is a typo)

aroradaman: @uablrek I tried to learn more about the go-scheduler and here is my understanding:

> Then I think it can't be done in this way. In go, (very) many go routines are mapped on a few OS-threads (equals the number of cores by default),

The scheduler is not limited to `GOMAXPROCS`, it tries to keep `GOMAXPROCS` os-threads and parks any additional os-thread (puts it to sleep for reuse in future). The scheduler also has a "handoff" mechanism, when the os-thread is blocked by a syscall the scheduler unparks a thread or creates a new one if none is available and then the run-queue of the blocked os-thread is handed over to the un-parked/new thread.

> l.Accept() will block the go-routine (but not any os-thread)

l.Accept() will also block os-thread as internally it uses blocking accept(4) syscall, scheduler will unpark a thread and handoff the runqueue to the unparked thread and wait on accept(4) to complete.

https://github.com/golang/go/blob/go1.22.3/src/runtime/proc.go#L21-L31

I think calling select on all sockets will be less expensive as handoff will happen for only one os-thread compared to blocking on accept(4) on multiple sockets which will trigger multiple handoffs.

uablrek: No, a Listener.Accept() does not spawn a new OS-thread. Do an experiment:
```go
package main;
import (
	"net"
	"sync"
)
func main() {
	var wg sync.WaitGroup
	l, _ := net.Listen("tcp", "[::1]:5006")
	for i := 0; i < 100; i++ {
		wg.Add(1)
		go func() {
			_, _ = l.Accept()
		}()
	}
	wg.Wait()
}
```
Compile and run:
```
GOMAXPROCS=1 ./accept
```
Check /proc:
```
> ps awwwxxxx|grep accept
 186386 pts/5    Sl+    0:00 ./accept
 186391 pts/2    S+     0:00 grep accept
> ls /proc/186386/task/
186386/  186387/  186388/  186389/
```
OS-threads are quite resource demanding, and starting, say 10,000 OS-threads would probably crash your PC, or at least stall it. On the other hand, go is made to make go-routines cheap.

uablrek: The go "runtime" does asynchronous "accept" (and other syscall's) and wake up the appropriate go-routines. (I haven't actually read the source, but this is how it must work). Go-routines are what's called [green threads](https://en.wikipedia.org/wiki/Green_thread) back in my days.

aroradaman: @uablrek I ran your code with **strace** to understand what happens under the hood and made some interesting observations:

_socket(AF_INET6, SOCK_STREAM|SOCK_CLOEXEC|SOCK_NONBLOCK, IPPROTO_IP) = 3_
**net.Listen** creates TCP socket with **SOCK_NONBLOCK**

_bind(3, {sa_family=AF_INET6, sin6_port=htons(5006), sin6_flowinfo=htonl(0), inet_pton(AF_INET6, "::1", &sin6_addr), sin6_scope_id=0}, 28) = 0_
socket binding

_listen(3, 8192)_ 
socket listen

_epoll_ctl(5, EPOLL_CTL_ADD, 3, {events=EPOLLIN|EPOLLOUT|EPOLLRDHUP|EPOLLET, data={u32=3196059650, u64=9180667291895857154}}) = 0_
This looked unexpected in the beginning, looked into the source code and found **Accept()** on **net.TCPListener** calls this to interact with [netpoller](https://morsmachine.dk/netpoller)
https://github.com/golang/go/blob/go1.22.3/src/internal/poll/fd_unix.go#L598

_accept4(3, 0xc0000d3d14, [112], SOCK_CLOEXEC|SOCK_NONBLOCK) = -1 EAGAIN (Resource temporarily unavailable)_
expected **Accept4** to fail with **EAGAIN** as socket was created using **SOCK_NONBLOCK**

_epoll_pwait(5, [], 128, 0, NULL, 0)     = 0_
This also looked unexpected, figured out from the source code that when **Accept4** fails with **EAGAIN** go waits for the underlying file descriptor to be ready using **netpoller** and that blocks the go-routine instead of os-thread.
https://github.com/golang/go/blob/go1.22.3/src/internal/poll/fd_unix.go#L611

On successful connection with a client the server flow resumes with
_accept4(3, {sa_family=AF_INET6, sin6_port=htons(52764), sin6_flowinfo=htonl(0), inet_pton(AF_INET6, "::1", 
&sin6_addr), sin6_scope_id=0}, [112 => 28], SOCK_CLOEXEC|SOCK_NONBLOCK) = 4_
This time **Accept4** is successful as the **fd** is ready for read.

And then the loop **Accept4** failing with **EAGAIN** and **EpollPwait** continues for 100 times (100 is the number of go-routines spawned in the example). 
I was expecting 100 concurrent Accept calls but that didn't happen because Accept is protected by lock. For a given socket/fd go will internally serialize all the Accept calls.
https://github.com/golang/go/blob/go1.22.3/src/internal/poll/fd_unix.go#L593-L596

TLDR;
 Calling **Accept** on **net.Listerner** doesn't block os-thread(thanks to netpoller). The current impl using **Select** will block the os-thread which will lead to performance issues.
 I will rework on the impl to use stdlib **net.Listen** in multiple go-routines instead of **Select** for multiplexing.

aroradaman:
> do we really need this?

Currently, for v1alpha2 I spawn [multiple helathz and metrics servers in go-routines](https://github.com/kubernetes/kubernetes/pull/121830/commits/cf40dc6a9dd3ff6d03e93145e0a9015f02528e53#diff-01be127f21d82a5a515f93cd1bb612be2d93d749968b193a6722e44b2cb20137R181-R204), this could really help there.

Also, Tim suggested this [here](https://github.com/kubernetes/kubernetes/issues/125055#issuecomment-2127554800)

uablrek:
> do we really need this?

My vote is on a library function because of testability. I can imagine some "what if"s that should be tested in (good) unit-tests at one place.

I also want to emphasise my ["context" comment](https://github.com/kubernetes/utils/pull/311#discussion_r1614677089):

> Also, I think your NewMultiListener should take a context as parameter. net.Listen uses [ListenConfig](https://pkg.go.dev/net#ListenConfig) which feels like a work-around for keeping backward compatibility.

danwinship:
> > do we really need this?
> 
> Currently, for v1alpha2 I spawn [multiple helathz and metrics servers in go-routines](https://github.com/kubernetes/kubernetes/pull/121830/commits/cf40dc6a9dd3ff6d03e93145e0a9015f02528e53#diff-01be127f21d82a5a515f93cd1bb612be2d93d749968b193a6722e44b2cb20137R181-R204), this could really help there.
> 
> Also, Tim suggested this [here](https://github.com/kubernetes/kubernetes/issues/125055#issuecomment-2127554800)

[KEP-2438 Dual-Stack API Server](https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/2438-dual-stack-apiserver#kube-apiserver-internals) needs it too

danwinship:
But you really shouldn't be implementing this in terms of `syscall`. Use goroutines. Trust the runtime.

net/multi_listen.go
```diff
@@ -0,0 +1,147 @@
...
+	addresses []string,
+	resolveTCPAddrFunc func(network, address string) (*net.TCPAddr, error),
+	listenFunc func(network, address string) (net.Listener, error),
+) (net.Listener, error) {
+	ml := &multiListener{
+		connAndErrCh: make(chan connAndErrPair),
+		stopCh:       make(chan struct{}),
+	}
+
+	for _, address := range addresses {
```
aojea, line 64: we may need to filter duplicate addresses? or do we want to fail (I think it will fail to create a listener in the same address twice)?

net/multi_listen.go
```diff
@@ -0,0 +1,147 @@
...
+				ml.connAndErrCh <- connAndErrPair{conn: conn, err: err, index: index}
+			}
+		}()
+	}
+	return ml, nil
+}
+
+// Accept is part of net.Listener interface.
+func (ml *multiListener) Accept() (net.Conn, error) {
+	connAndErr := <-ml.connAndErrCh
```
aojea, line 128:
```suggestion
	connAndErr, ok := <-ml.connAndErrCh
	if !ok {
	  // close everything if the channel has been closed
```

net/multi_listen.go
aroradaman: I think the caller be responsible for that (not passing the same address twice).
I can't imagine a case where there is a need to listen on the same address twice, if it is passed it might be because of some bug and we should error out instead of silently ignoring the dup address, right?

net/multi_listen.go
aroradaman: update the code and unit test.

net/multi_listen.go
```diff
@@ -0,0 +1,151 @@
...
+					// Accept() will throw "use of closed network connection" when listener is closed,
+					// we can ignore that error and break out of this goroutine.
+					case <-ml.stopCh:
+						return
+					default:
+						ml.connAndErrCh <- connAndErrPair{conn: conn, err: err, index: index}
+					}
+				}
+				ml.connAndErrCh <- connAndErrPair{conn: conn, err: err, index: index}
+			}
```
aojea, line 120: I think you can remove the stopCh just using `connAndErrCh` 
```suggestion
				select {
				case ml.connAndErrCh <- connAndErrPair{conn: conn, err: err, index: index}
				default:
				         return
			}
```

uablrek:
You _must_ have support for `context`. Please see https://github.com/kubernetes/utils/pull/311#discussion_r1614677089. I will make a better review later.

net/multi_listen.go
aojea: fair

net/multi_listen.go
aroradaman: There can be some corner cases in that approach, if there is a genuine error in Accept() call (when listeners are not closed) then the worker go-routine will try consuming from `connAndErrCh` and if the channel already has something to read from then that connection will be dropped.

uablrek:
I think the multi listener must be backward compatible. That means that a call like:
```go
l,e := MultiListen(ctx, []string{":8080"})
```
must work, and listen on both families, and:
```go
l,e := MultiListen(ctx, []string{"0.0.0.0:8080"})
// or
l,e := MultiListen(ctx, []string{"[::]:8080"})
```
must also listen on both families.

net/multi_listen.go
aojea: don't completely follow, is an unbuffered channel and now you always send anything through the channel expect when the stopCh is closed, what is the different on using the stopCh to sync the goroutines vs using the connAndErrCh?

aroradaman:
> You _must_ have support for `context`. Please see [#311 (comment)](https://github.com/kubernetes/utils/pull/311#discussion_r1614677089). I will make a better review later.

net.Listener interface won't allow Accept to consume context, should I store the context with the structure to check context in the Accept call?

uablrek:
Actually, I don't think you have to store it. Just create the Listerners with something like:
```go
l,e := net.ListenConfig{}.Listen(ctx, network, address)
```
Then I suppose all Accept() will exit with an error if the ctx is canceled

net/multi_listen.go
aroradaman: accept() immediately fails with "use of closed network connection" when the listener is closed and I used stopCh to determine that and exit the goroutine. I can use atomic.Bool instead.

This error is not exported by stdlib, so I need some way to determine the exit condition for the goroutines.

net/multi_listen.go
```diff
@@ -0,0 +1,153 @@
...
+var _ net.Listener = &multiListener{}
+
+// MultiListen returns net.Listener which can listen for and accept
+// TCP connections on multiple addresses.
+func MultiListen(ctx context.Context, addresses []string) (net.Listener, error) {
+	return multiListen(
+		ctx,
+		addresses,
+		net.ResolveTCPAddr,
+		func(ctx context.Context, network, address string) (net.Listener, error) {
```
uablrek, line 56: I assume the "Func"s are there to simplify unit-test? If so, please comment that since it otherwise seems ... un-motivated

uablrek: Also, please verify that when net.ListenConfig is used, then the "KeepAlive" default is the golang-default (15s), and not the os-default (usually 2h on Linux). If it is the os-default, then set it to 15s explicitly to be in sync with golang standard, but that may of course change in the future (sigh),

net/multi_listen.go
uablrek: In the same area: return an error if the address slice is empty. I think it works like this now: (ml, nil) ia returned, and a subsequent Accept() hangs forever. While this is logical (sort of), it's probably not what the caller intended.

net/multi_listen.go
```diff
@@ -0,0 +1,153 @@
...
+	for i := range ml.listeners {
+		_ = ml.listeners[i].Close()
+	}
+	ml.wg.Wait()
+	return nil
+}
+
+// Addr is part of net.Listener interface.
+func (ml *multiListener) Addr() net.Addr {
+	return ml.addrs[ml.latestAcceptedIndex.Load()]
```
uablrek, line 152: IMO the "latestAcceptedIndex" should be removed. It is unreliable (race) and may lure a caller to actually think it can be used for something sensible. Make a comment instead that the Addr() call _is_ unreliable, and the caller should use conn.LocalAddr() to get the _actual_ local address.
Just return ml.addrs[0].

net/multi_listen.go
```diff
@@ -0,0 +1,153 @@
...
+			// close all listeners
+			_ = ml.Close()
+			return nil, err
+		}
+
+		ml.addrs = append(ml.addrs, addr)
+		ml.listeners = append(ml.listeners, l)
+	}
+
+	for i := range ml.listeners {
```
uablrek, line 109: Use
```go
for i, l := range ml.listeners {
```
or, if you accept to drop latestAcceptedIndex:
```go
for _, l := range ml.listeners {
```

net/multi_listen.go
uablrek: The atomic bool doesn't prevent the race at all. The accept go-routine may be scheduled out _right before_:
```go
				ml.connErrIndexCh <- connErrIndexTriplet{conn: conn, err: err, index: index}
```
_then_ Close() is called, and the go-routine is re-scheduled, and the send will be on a closed channel -> panic().

IMHO you should keep it simple. Close the channel _after_ `ml.wg.Wait()`, and just send all errors from Accept() on the channel. Let the user sort it out. They should be prepared that Accept() returns an error if the listener is closed (and will likely ignore them)

aroradaman: > The atomic bool doesn't prevent the race at all. The accept go-routine may be scheduled out right before:
> then Close() is called, and the go-routine is re-scheduled, and the send will be on a closed channel -> panic().

When **Close()** is called it sets the atomic bool to true, and in the accept go-routine I simply exit/return when atomic bool is true instead of writing on the channel and no panic would occur.
```go
			for {
				conn, err := ml.listeners[index].Accept()
				if err != nil && ml.closed.Load() {
					return
				}
				ml.connErrIndexCh <- connErrIndexTriplet{conn: conn, err: err, index: index}
			}
```



> IMHO you should keep it simple. Close the channel after ml.wg.Wait(), and just send all errors from Accept() on the channel. Let the user sort it out. They should be prepared that Accept() returns an error if the listener is closed (and will likely ignore them)

In that case I would need to exit the accept for loop whenever any error occurs, right? Other wise the go-routine would never end and **ml.Accept()** would be blocked on **ml.wg.Wait()** forever. 
But we should only exit on "use of closed network connection" which occurs when listener is closed, right? Other errors are genuine and should be passed to user.

Right now this atomic bool is being used to determine the exit condition.

net/multi_listen.go
aroradaman: I can't directly simplify to 
```go
for _, l := range ml.listeners {
```
as the variable `l` captured by the gorutine will always point to the last listener. 
https://go.dev/wiki/CommonMistakes

net/multi_listen.go
aroradaman: https://github.com/golang/go/blob/go1.22.4/src/net/dial.go#L806-L809
net.Listen() starts with empty ListenConfig, all the default values will besame.

net/multi_listen.go
uablrek:
```go
			for {
				conn, err := ml.listeners[index].Accept()
				if err != nil && ml.closed.Load() {
					return
				}
// if the go-routine is suspended *right here*, after reading the "closed" boolean, but before sending on the
// channel, and *then* another go-routine calling close and thus set the "closed" boolean (which is already read
// as "true" by this go-routine) and closing the channel, then you will get a panic() when this go-routine is
// re-scheduled and sends on a closed channel! This is true regardless if the boolean is atomic, or not.
				ml.connErrIndexCh <- connErrIndexTriplet{conn: conn, err: err, index: index}
			}
```

uablrek: > In that case I would need to exit the accept for loop whenever any error occurs, right?

Um, yes you are right. If you use a "closed" boolean (which should not be atomic), you must use a Mutex, e.g.
```go
			for {
				conn, err := ml.listeners[index].Accept()
				ml.mutex.Lock()
				if ml.closed {
					ml.mutex.Unlock()
					return
				} else {
					ml.connErrIndexCh <- connErrIndexTriplet{conn: conn, err: err, index: index}
				}
				ml.mutex.Unlock()
			}
// And
func (ml *multiListener) Close() error {
	ml.mutex.Lock()
	ml.closed = true
	close(ml.connErrIndexCh)
	ml.mutex.Unlock()
	for i := range ml.listeners {
		_ = ml.listeners[i].Close()
	}
	ml.wg.Wait()
	return nil
}
```

net/multi_listen.go
aroradaman: how a mutex instead of atmoic.Bool ?
```go
			for {
			       conn, err := ml.listeners[index].Accept()
			       ml.mux.Lock()
			       if err != nil && ml.closed {
					return
			       }
			       ml.connErrIndexCh <- connErrIndexTriplet{conn: conn, err: err, index: index}
			       ml.mux.Unlock()
			}
```

net/multi_listen.go
aroradaman: how a mutex instead of atmoic.Bool ?
```go
			for {
			       conn, err := ml.listeners[index].Accept()
			       ml.mux.Lock()
			       if err != nil && ml.closed {
					return
			       }
			       ml.connErrIndexCh <- connErrIndexTriplet{conn: conn, err: err, index: index}
			       ml.mux.Unlock()
			}
```

uablrek: The Accept() is "semantically weird". Since the multi-listener _always_ have Accept's running on all addresses, stale entries will linger on the channel until Accept() is called the next time. For instance if the multi-listener is created with one IPv4 and one IPv6 (the use-case this implementation is mainly addressing), and the client is using [happy eyeballs](https://en.wikipedia.org/wiki/Happy_Eyeballs) and send a connect on both families in parallel, then the first will succeed and the runner-up will linger in the channel until Accept() is called again (which may be much later)

net/multi_listen.go
uablrek: Thanks for the pointer. I didn't know. But this should work, shouldn't it?
```go
	for _, l := range ml.listeners {
		ml.wg.Add(1)
		// spawn a go routine for every listener to wait for incoming connection requests
		go func(l net.Listener) {
			defer ml.wg.Done()
			for {
				conn, err := l.Accept()
				if err != nil && ml.closed.Load() {
					return
				}
				ml.connErrIndexCh <- connErrIndexTriplet{conn: conn, err: err}
			}
		}(l)
	}
```
(at least it passes `go test`)

net/multi_listen.go
```diff
@@ -0,0 +1,154 @@
...
+	"fmt"
+	"net"
+	"sync"
+	"sync/atomic"
+)
+
+// connErrIndexTriplet packs results of accept call on actual listeners along with the
+// index of the listener. It is used for communication between main and listener goroutines.
+type connErrIndexTriplet struct {
+	index int
```
uablrek, line 30: index is not used. It should be removed, and rename the struct and channel

net/multi_listen.go
uablrek: In https://github.com/kubernetes/utils/pull/311#discussion_r1641776790: do not check err != nil, if ml.closed==true you must never write to the channel (it's closed), and unlock before return.

net/multi_listen.go
```diff
@@ -0,0 +1,160 @@
...
+
+// Close is part of net.Listener interface.
+func (ml *multiListener) Close() error {
+	ml.mu.Lock()
+	ml.closed = true
+	close(ml.connErrCh)
+	for _, l := range ml.listeners {
+		_ = l.Close()
+	}
+	ml.mu.Unlock()
```
uablrek, line 150: Move this above the close of listeners. A critical region should be kept as small as possible

net/multi_listen.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+
+		var network string
+		host, _, err := net.SplitHostPort(addr.String())
+		if err != nil {
+			// close all listeners
+			_ = ml.Close()
+			return nil, err
+		}
+		switch IPFamilyOf(ParseIPSloppy(host)) {
```
thockin, line 90: I think I'd rather see this as an argument, rather than try to figure it out.  It seems like an opportunity to get it wrong, when we don't need to.

net/multi_listen.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// Close is part of net.Listener interface.
+func (ml *multiListener) Close() error {
+	ml.mu.Lock()
+	ml.closed = true
+	close(ml.connErrCh)
+	ml.mu.Unlock()
+	for _, l := range ml.listeners {
+		_ = l.Close()
```
thockin, line 149: I think this could be clearer with comments explaining what is going on.  It's just subtle enough that you can prevent the need to go RTFM.

e.g. does `l.Close()` cause the blocked `Accept()` calls to immediately return?  Saying that here would help.

net/multi_listen.go
```diff
@@ -0,0 +1,160 @@
...
+				}
+				ml.connErrCh <- connErrPair{conn: conn, err: err}
+				ml.mu.Unlock()
+			}
+		}(l)
+	}
+	return ml, nil
+}
+
+// Accept is part of net.Listener interface.
```
thockin, line 132: I would explain here the specific semantics - this will wait for ANY of the sub-listeners to get a connection.

net/multi_listen.go
```diff
@@ -0,0 +1,160 @@
...
+	mu        sync.Mutex
+	connErrCh chan connErrPair
+	closed    bool
+}
+
+// compile time check to ensure *multiListener implements net.Listener.
+var _ net.Listener = &multiListener{}
+
+// MultiListen returns net.Listener which can listen for and accept
+// TCP connections on multiple addresses.
```
thockin, line 48: I think this comment could be more verbose to explain the semantics

uablrek:
The multi-listener _will_ have different behavior compared to the standard Listener, as mentioned in https://github.com/kubernetes/utils/pull/311#discussion_r1641779949. I can't see a way to make them exactly the same, but the differences must be investigated and documented. Otherwise users will assume that they are totally compatible. I can make some tests, but not this week.

Related is [how TCP backlog works in Linux](http://veithen.io/2014/01/01/how-tcp-backlog-works-in-linux.html). I thought I knew (I didn't), and have been hit by [half-open connection](http://en.wikipedia.org/wiki/Half-open_connection) problems at least once (in ~20y, so it is very rare).

net/multi_listen.go
thockin: > The Accept() is "semantically weird".

Excellent callout, Lars.  Is it better to be semantically close but not identical, or to emphasize the semantic difference?  It would require more app changes, but we could do something like:

```
// I am not sure we need the Listener here
acceptHandler := func(listener net.Listener, conn net.Conn, err error) {
    if err != nil {
        // log it
    }
    // handle an accepted socket however I need to, maybe a channel
    // or some other queue or handler-pool mechanism
}
ml := knet.MultiListen("tcp", acceptHandler, "10.9.8.7:80", "10.1.2.3:8000")
```

net/multi_listen.go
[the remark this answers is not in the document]
aroradaman: "0.0.0.0" will bind to all interfaces on IPv4 and IPv6, and if the caller passed ["0.0.0.0:80", [::]:80 ] with network "tcp" expecting to listen on port 80 on all IPv4 and IPv6 addresses then the call will fail.

The primary use case of the multi-listener is to bind on dual-stack IP pairs, so the question comes to should the callers provide ["0.0.0.0:80"] or  ["0.0.0.0:80", [::]:80 ] to listen on port 80 on all interfaces.

If we only want 0.0.0.0 to bind on IPv4 then we can internally add the listener using the similar approach that [LocalPort](https://github.com/kubernetes/utils/blob/master/net/port.go#L107-L113)  uses.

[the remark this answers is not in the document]
thockin: I always get this messed up, so I had to write a program.

> if the caller passed ["0.0.0.0:80", [::]:80 ] with network "tcp" expecting to listen on port 80 on all IPv4 and IPv6 addresses then the call will fail

We could be smart and parse the IPs and if there's a zero-value, just do that?  We do that in other places

> should the callers provide ["0.0.0.0:80"] or ["0.0.0.0:80", [::]:80 ] to listen on port 80 on all interfaces.

Seems like "Be liberal in what you accept" applies here.  The docs should say ":80" is how you get ALL interfaces, but we can accept more flexibly

> If we only want 0.0.0.0 to bind on IPv4 then

This is exactly why taking a `network` arg makes sense to me - the user CAN pass "tcp4" and their meaning is unambiguous.  My fear here is that your logic may just be too clever, for little real value.  After writing this all out, I guess I don't HATE it, it just makes me uncertain.  `ResolveTCPAddr("tcp", ":80")` returns ":80", so we need to test and handle that, too

```
$ cat main.go
package main

import (
	"fmt"
	"net"
	"os"
)

func main() {
	if len(os.Args) != 3 {
		fmt.Printf("USAGE: %s <net> <addr>\n", os.Args[0])
		os.Exit(1)
	}

	l, err := net.Listen(os.Args[1], os.Args[2])
	if err != nil {
		fmt.Printf("FATAL: %v\n", err)
		os.Exit(1)
	}
	for {
		c, err := l.Accept()
		if err != nil {
			fmt.Printf("FATAL: %v\n", err)
			os.Exit(1)
		}
		c.Close()
	}
}


$ ### Prove that nothing works

$ date | nc 127.0.0.1 9999 && echo pass || echo fail
fail

$ date | nc ::1 9999 && echo pass || echo fail
fail


$ ### Listen on "tcp" with IPv4 localhost

$ go run main.go "tcp" 127.0.0.1:9999 &

$ date | nc 127.0.0.1 9999 && echo pass || echo fail
pass

$ date | nc ::1 9999 && echo pass || echo fail
fail

$ kill %1


$ ### Listen on "tcp" with IPv6 localhost

$ go run main.go "tcp" [::1]:9999 &

$ date | nc ::1 9999 && echo pass || echo fail
pass

$ date | nc 127.0.0.1 9999 && echo pass || echo fail
fail

$ kill %1


$ ### Listen on "tcp" with no addr

$ go run main.go "tcp" :9999 &

$ date | nc 127.0.0.1 9999 && echo pass || echo fail
pass

$ date | nc ::1 9999 && echo pass || echo fail
pass

$ kill %1


$ ### Listen on "tcp4" with IPv4 localhost

$ go run main.go "tcp4" 127.0.0.1:9999 &

$ date | nc 127.0.0.1 9999 && echo pass || echo fail
pass

$ date | nc ::1 9999 && echo pass || echo fail
fail

$ kill %1


$ ### Listen on "tcp4" with IPv6 localhost

$ go run main.go "tcp4" [::1]:9999 &
FATAL: listen tcp4: address ::1: no suitable address found
exit status 1


$ ### Listen on "tcp4" with no addr

$ go run main.go "tcp4" :9999 &

$ date | nc 127.0.0.1 9999 && echo pass || echo fail
pass

$ date | nc ::1 9999 && echo pass || echo fail
fail

$ kill %1


$ ### Listen on "tcp6" with IPv4 localhost

$ go run main.go "tcp6" 127.0.0.1:9999 &
FATAL: listen tcp6: address 127.0.0.1: no suitable address found
exit status 1


$ ### Listen on "tcp6" with IPv6 localhost

$ go run main.go "tcp6" [::1]:9999 &

$ date | nc 127.0.0.1 9999 && echo pass || echo fail
fail

$ date | nc ::1 9999 && echo pass || echo fail
pass

$ kill %1


$ ### Listen on "tcp6" with no addr

$ go run main.go "tcp6" :9999 &

$ date | nc 127.0.0.1 9999 && echo pass || echo fail
fail

$ date | nc ::1 9999 && echo pass || echo fail
pass

$ kill %1


$ ### Listen on "tcp" with IPv4 0 addr

$ go run main.go "tcp" 0.0.0.0:9999 &

$ date | nc 127.0.0.1 9999 && echo pass || echo fail
pass

$ date | nc ::1 9999 && echo pass || echo fail
pass

$ kill %1


$ ### Listen on "tcp" with IPv6 0 addr

$ go run main.go "tcp" [::]:9999 &

$ date | nc 127.0.0.1 9999 && echo pass || echo fail
pass

$ date | nc ::1 9999 && echo pass || echo fail
pass

$ kill %1
```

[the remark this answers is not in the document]
danwinship: > "0.0.0.0" will bind to all interfaces on IPv4 and IPv6

One thing I think Tim was trying to get at but did not explicitly say:

    net.Listen("tcp", "0.0.0.0:9999")

will bind to all interfaces on IPv4 and IPv6, but

    net.Listen("tcp4", "0.0.0.0:9999")

will only bind to IPv4.

net/multi_listen.go
```diff
@@ -0,0 +1,160 @@
...
+	for _, address := range addresses {
+		addr, err := resolveTCPAddrFunc("tcp", address)
+		if err != nil {
+			// close all listeners
+			_ = ml.Close()
+			return nil, err
+		}
+
+		var network string
+		host, _, err := net.SplitHostPort(addr.String())
```
danwinship, line 84: while obviously there's something to be said for having `MultiListen` copy `net.Listen`'s API, the flip side of that is that probably most callers start out with separate IP and port, and I'm pretty sure we _never_ want the DNS-resolving behavior, and also, for our current use cases, it's always multiple IPs listening on the same port, so if `MultiListen` took an array of IPs and a single port, you'd have to do less splitting and joining later...

(Curious what other people think about "copy `net.Listen`" vs "make it better for _our_ use case".)

net/multi_listen.go
danwinship, line 90: I have exactly the opposite opinion from Tim on this; passing the family as an argument is an opportunity for the caller to get it wrong (`{IP: "1.2.3.4", Network: "tcp6"}`), and there's no reason to, since we can 100% reliably decide whether a valid IP address is IPv4 or IPv6.

danwinship, line 90: `IPFamilyOfString()`

net/multi_listen.go
```diff
@@ -0,0 +1,160 @@
...
+			return nil, err
+		}
+		switch IPFamilyOf(ParseIPSloppy(host)) {
+		case IPv4:
+			network = "tcp4"
+		case IPv6:
+			network = "tcp6"
+		default:
+			// close all listeners
+			_ = ml.Close()
```
danwinship, line 97: maybe have a `defer` that does this automatically if the function is going to return an error, so you don't have to get it right in each place

net/multi_listen.go
```diff
@@ -0,0 +1,160 @@
...
+				ml.mu.Unlock()
+			}
+		}(l)
+	}
+	return ml, nil
+}
+
+// Accept is part of net.Listener interface.
+func (ml *multiListener) Accept() (net.Conn, error) {
+	connErr, ok := <-ml.connErrCh
```
danwinship, line 134: I don't think this will work. In particular, if a connection arrives when there is not an `Accept()` pending, and then the caller calls `Close()`, it will deadlock, because the listener thread will be blocked trying to write to `ml.connErrCh`, while holding `ml.mu`, so the `Close()` thread will never be able to get `ml.mu`. 

You generally don't want to write to a blocking channel while holding a mutex. (And of course, you can't do it from outside of the mutex the way the code is currently structured because then you might end up writing to it after it's closed.)

The way we usually handle things like this is to have a channel that is _only_ used as a semaphore to wake up the waiting thread, and do the actual data-passing via fields on `ml`, so you don't have to worry about channel buffering. So the listener thread would do

    ml.accepted = append(ml.accepted, connErrPair{conn: conn, err: err})
    // signal acceptReadyCh, unless it has already been signaled
    select {
            case ml.acceptReadyCh <- struct{}{}:
            default:
    }

(The `select` with the `default` keeps it from blocking if the channel is already "full".)

Then the accept thread would check for an item in `ml.accepted`, and if there isn't one, read off `ml.acceptReadyCh` and try again.

(And then various changes to `Close()` to make it all work...)

net/multi_listen.go
```diff
@@ -0,0 +1,160 @@
...
+		_ = l.Close()
+	}
+	ml.wg.Wait()
+	return nil
+}
+
+// Addr is part of net.Listener interface.
+// Addr always returns the address of the first listener, callers should
+// rely on conn.LocalAddr() to get the actual local address of the listener.
+func (ml *multiListener) Addr() net.Addr {
```
danwinship, line 158: FWIW `net.Addr` is itself an interface, so we could return something that just returns dummy data if we wanted. (Nothing cares; the interface is way too underspecified to actually be useful unless you know in advance what the actual object types are.)

net/multi_listen_test.go
```diff
@@ -0,0 +1,282 @@
...
+	return nil
+}
+
+func (f *fakeListener) Addr() net.Addr {
+	return f.addr
+}
+
+var _ net.Listener = &fakeListener{}
+
+func TestMultiListen(t *testing.T) {
```
danwinship, line 84: I think we need some tests of accept-while-close, close-with-pending-connections, etc

net/multi_listen.go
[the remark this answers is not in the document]
thockin: TL;DR at the bottom.

If we are positioning this as a mostly-generic library, with an API we can support (people *will* find it and use it), we should be really clear on behavior.  "tcp" vs "tcp4" vs "tcp6" can be implied by looking at the IP address, except in the case of zero.  For clarity, Go treats ":80", "0.0.0.0:80", and "[::]:80" as identical in the "tcp" network.  There may even be other ways to spell that, I am not an expert in all the IPv6 syntax :)

Assuming we want to not ask the user for a network in this multi-listen API:

0) I think it is reasonable to assert that ":80" means "tcp" - agree or disagree?
1) Do we think "0.0.0.0:80" means "tcp" or "tcp4"?  
2) Do we think "[::]:80" means "tcp" or "tcp6"?

Principle of least least surprise is ambiguous here, IMO.  Given how many times and places we have discussed and documented this, I think it is safe to say that no matter what we choose, someone will be surprised.  On one hand, the "usual" meaning of those zero addresses is "both families", so least-surprise says "be similar".  On the other hand, the "usual" meaning is surprising.

If either of those means "tcp" then I think they both mean "tcp".  AFAIK those are the only ambiguous cases - using "tcp" with any other IP value will only match on that specific IP family.  So we should only ever need to specify "tcp" -- there's no purpose in looking at the IP to decide  "tcp4" or "tcp6".  What about weird IPv6 encodings, like "::ffff:7f00:0001", which is "127.0.0.1" encoded in an IPv6 address (but still using IPv4 as a protocol.  Would this library treat that as "tcp6" ?

```
$ go run main.go "tcp" 127.0.0.1:9376 &
[1] 3502335

$ date | nc 127.0.0.1 9376 && echo pass || echo fail
pass

$ date | nc ::1 9376 && echo pass || echo fail
fail

$ date | nc ::ffff:7f00:0001 9376 && echo pass || echo fail
pass
```

and

```
$ go run main.go "tcp" [::ffff:7f00:0001]:9376 &
[1] 3502767

$ date | nc 127.0.0.1 9376 && echo pass || echo fail
pass

$ date | nc ::1 9376 && echo pass || echo fail
fail

$ date | nc ::ffff:7f00:0001 9376 && echo pass || echo fail
pass
```

In fact, today I learned something.  `net.ResolveTCPAddr("tcp", "::ffff:7f00:0001")` produces "127.0.0.1", so maybe this is already "correct enough"?  It does allow name resolution, which suggests that "localhost" should work, but it does not.  That's SUPER annoying, and suggests that doing such resolution is kind of a trap.

```
$ go run main.go "tcp" localhost:9376 &
[2] 3504806
localhost:9376 resolves to 127.0.0.1:9376

$ date | nc 127.0.0.1 9376 && echo pass || echo fail
pass

$ date | nc ::1 9376 && echo pass || echo fail
fail

$ date | nc ::ffff:7f00:0001 9376 && echo pass || echo fail
pass
```

TL;DR: 

1) Do we think that a user of this library (which may or may not be re-exposed to k8s CLIs) should be able to decide whether "0.0.0.0:80" and "[::]:80", or do we think that zero is REALLY SPECIAL and should always mean both families?

2) Do we think that a user of this library should resolve names like "localhost" into a single family?  Or into multiple families?  Or not at all?

net/multi_listen.go
thockin: My initial instinct was that it should copy `net.Listen()` but I really am OK if we decide otherwise - I just want it to be well thought out.

Passing the network argument, like `net.Listen()`, is explicit and familiar, which has some advantages, even if some of the behavior is surprising to newcomers.

Doing it "the right way" automatically is different, but that can be documented away.  Some of the magic might be surprising to expert users.

Last thing to consider: if we want to use this in k8s, we have to retain the historical semantics, which (I think?) are the Go default semantics for "tcp".  Making this library TOO smart could break kube users.

net/multi_listen.go
thockin: It's the special cases that worry me.  Given a totally blank slate, I think this works great.

* ":80" -> all IPs on all families
* "0.0.0.0:80" -> all IPs on IPv4
* "[::]:80" -> all IPs on IPv6
* Any other IP = auto-detect
* Maybe something clever around localhost?

But doing that would break k8s users, for whom "0.0.0.0:80" already means "all families".   So we can change the magic such that any zero address means "all families", at which point it's the same as if we just hardcode "tcp" (I think?), meaning there's just no way to specify "all IPs on IPv4".

If we want to make this lib flexible AND useful to k8s, we need the network argument.

net/multi_listen.go
thockin: agree generally, but it needs to define `err` at a higher scope and capture it, and never redefine it, which breaks a lot of muscle memory for people

```
var err error
defer func() { if err != nil { ml.Close() }() }

// this breaks and does not close it
if err := something(); err != nil {
    return err
}
```

net/multi_listen_test.go
thockin: and some tests of HTTP serving on a multi-listener

aroradaman:
> Actually, I don't think you have to store it. Just create the Listerners with something like:
> 
> ```go
> l,e := net.ListenConfig{}.Listen(ctx, network, address)
> ```
> 
> Then I suppose all Accept() will exit with an error if the ctx is canceled

@uablrek https://github.com/golang/go/issues/50861

net/multi_listen.go
aroradaman: > while obviously there's something to be said for having MultiListen copy net.Listen's API, the flip side of that is that probably most callers start out with separate IP and port, and I'm pretty sure we never want the DNS-resolving behavior, and also, for our current use cases, it's always multiple IPs listening on the same port, so if MultiListen took an array of IPs and a single port, you'd have to do less splitting and joining later...

If we remove the dns-resolving behaviour (which anyway **net.Listen** will do internally) and consume the network as an argument then there won't be any splitting and joining. 

I can't think of any use case where any consumer would want to use multi-listener for listening on different ports, and the primary use case of being able to listen on dual-stack IP pairs, I think using a list of IPs and a port will better interface. 

By using the network as an argument we don't need to handle many corner cases and the semantics will follow go defaults.  The following signature makes morse sense to me now:

```go
func MultiListen(ctx context.Context, network string, hosts []string, port int) (net.Listener, error)
```


Maybe we should also get rid of context here? as context is not fully utilized by stdlib. Cancellation/Timeout of context doesn't unblock Accept() calls. 
(ref: https://github.com/golang/go/issues/50861)

uablrek:
> > Actually, I don't think you have to store it. Just create the Listerners with something like:
> > ```go
> > l,e := net.ListenConfig{}.Listen(ctx, network, address)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then I suppose all Accept() will exit with an error if the ctx is canceled
> 
> @uablrek [golang/go#50861](https://github.com/golang/go/issues/50861)

Is there _any_ way to break a blocking Accept()??

net/multi_listen.go
thockin: > I can't think of any use case where any consumer would want to use multi listener for listening on different ports

Whether we can think of them or not, should we go out of our way to prevent them?  What I mean is that net.Listen() takes a host:port string, so why would we make this different on purpose?  And how will we expose this in k8s?  I presume the existing CLIs which take "host:port" would now take a list of host:port, no?  In that case the ports CAN be different, even if we don't know why someone would do that.

net/multi_listen.go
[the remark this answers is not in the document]
danwinship: (The behavior of IPv4-mapped IPv6 addresses is not consistent across tools and APIs (even across `net.ParseIP` and `netip.ParseAddr`) and there is never any good reason for using them in properly-dual-stack-aware software, so we should be moving toward banning them everywhere.)

net/multi_listen.go
danwinship: I am fine with passing _one_ `network` argument to `MultiListen` if you want to restrict the listener as a whole to either IPv4 or IPv6, but I thought you were suggesting passing per-IP `network` arguments, which seems to be just annoying and error-prone.

net/multi_listen.go
danwinship:
```
success := false
defer func() { if !success { ml.Close() } }()

...

// this is fine
if err := something(); err != nil {
    return err
}

...

success = true
return nil
```

danwinship:
OK, so...

We could do:

    MultiListen(ctx context.Context, network string, addresses []string) (net.Listener, error)

where the semantics are basically:

    for i := range addresses {
            lc.Listen(ctx, network, addresses[i])
    }

with no added intelligence.

If we want per-address `network` fields, then I think a more logical API would be drop `MultiListen` and instead have the caller do

```
ml := netutils.NewMultiListener()
err := ml.Listen(ctx, "tcp4", "10.0.0.0:80")
if err != nil ...
err = ml.Listen(ctx, "tcp6", "[::]:80")
if err != nil ...
...
```

I think both of these are equally unambiguous, though the latter is more flexible.

I can imagine that for some callers, being able to separately listen on `"tcp4", "0.0.0.0:80"` and `"tcp6", "[::]:80"` might make more sense given their internal organization?


Regardless of which of the above approaches we go to, the other question is: what should happen if you listen on both `"tcp", "0.0.0.0:80"` and `"tcp", "[::]:80"` (which is one of the bugs that got us here, https://github.com/kubernetes/kubernetes/issues/114702)? Should MultiListener DTRT here, or should it just do literally what you asked, and end up returning an error (meaning that it's up to callers to know better than to do that).

net/multi_listen.go
thockin: FWIW, the context is verly limited utility.  It bounds the `resolve` step and is passed back to `ListenConfig.Control()` if that is defined.  It does not cancel a `Listen()` or an `Accept()`.

thockin:
Taking (or building) a list of individual `Listener` is about as flexible as gets.  It's not what I meant, but I can see the appeal.  Given that the primary consumer of this API will be kubernetes, I suspect that's more flexible that we need, but it seems low-cost to do it that way.  I see now the argument that dual-stack is a primary motivator, so a single `network` argument for a list of IPs may be pointless.

WRT zero addresses, if we make the caller call `Listen()`, then it should be on them to not collide - we should not be too smart.  Effectively this becomes an `accept-mux` - we need to nail those semantics.

net/multi_listen_test.go
aroradaman: > and some tests of HTTP serving on a multi-listener

do you mean an integration test that uses real HTTP servers?

thockin: I think it can still be a unit test (since this repo doesn't really have integration tests).  Just use `:0` for a random port.  Create 2 or 3 such ports, pass the multi-listener to `http.Serve` and then connect to each of the ports that were allocated, ensuring they get the correct HTTP response.

net/multi_listen.go
```diff
@@ -0,0 +1,182 @@
...
+			defer ml.wg.Done()
+			for {
+				conn, err := l.Accept()
+				ml.mu.Lock()
+				if ml.closed {
+					ml.mu.Unlock()
+					return
+				}
+				// enqueue the accepted connection
+				ml.connErrQueue = append(ml.connErrQueue, connErrPair{conn: conn, err: err})
```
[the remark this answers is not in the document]
danwinship: > This means the queue is unbounded? Is that really what we want?

No, probably not.

But your version above has the race condition: it's possible for `ml.Close()` to be called in between the `if closed { return }` and the `ch <- connErrPair{...` in the listening thread, resulting in a panic.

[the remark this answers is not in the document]
danwinship: > But as far as I can tell there is no non-blocking `Accept()` in Go (unless maybe we fiddle with the socket FD which seems dangerous).

(and not Windows-compatible)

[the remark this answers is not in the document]
thockin: Indeed, that was bad.  This may be better.  Gets rid of the mutex entirely, at the cost of one more channel, but follows the Go ideal of "senders close channels".

```diff
diff --git a/net/multi_listen.go b/net/multi_listen.go
index d508805..6a94f46 100644
--- a/net/multi_listen.go
+++ b/net/multi_listen.go
@@ -29,20 +29,28 @@ type connErrPair struct {
 	err  error
 }
 
+func (p connErrPair) String() string {
+	if p.conn != nil {
+		return p.conn.RemoteAddr().String()
+	}
+	if p.err != nil {
+		return p.err.Error()
+	}
+	return "neither conn nor err were valid"
+}
+
 // multiListener implements net.Listener
 type multiListener struct {
 	listeners []net.Listener
 
-	wg     sync.WaitGroup
-	mu     sync.Mutex
-	closed bool
+	wg sync.WaitGroup
 
-	// connErrQueue holds the connections accepted by sub-listeners
-	connErrQueue []connErrPair
-
-	// acceptReadyCh is used as a semaphore to wake up the waiting
-	// multiListener.Accept() when new connections are available
-	acceptReadyCh chan any
+	// connChs passes accepted connections, from child listeners to parent.
+	connChs []chan connErrPair
+	// readyCh indicates that a child listener's connCh has data.
+	readyCh chan struct{}
+	// stopCh communicates from parent to child listeners.
+	stopCh chan struct{}
 }
 
 // compile time check to ensure *multiListener implements net.Listener
@@ -58,13 +66,13 @@ var _ net.Listener = &multiListener{}
 //  2. Use "tcp4" or "tcp6" to exclusively listen on IPv4 or IPv6 family, respectively.
 //  3. The host can accept names (e.g, localhost) and it will create a listener for at
 //     most one of the host's IP.
-func MultiListen(ctx context.Context, network string, addrs []string) (net.Listener, error) {
+func MultiListen(ctx context.Context, network string, addrs ...string) (net.Listener, error) {
+	var lc net.ListenConfig
 	return multiListen(
 		ctx,
 		network,
 		addrs,
 		func(ctx context.Context, network, address string) (net.Listener, error) {
-			var lc net.ListenConfig
 			return lc.Listen(ctx, network, address)
 		})
 }
@@ -78,14 +86,16 @@ func multiListen(
 	listenFunc func(ctx context.Context, network, address string) (net.Listener, error),
 ) (net.Listener, error) {
 	if !(network == "tcp" || network == "tcp4" || network == "tcp6") {
-		return nil, fmt.Errorf("network '%s' not supported", network)
+		return nil, fmt.Errorf("network %q not supported", network)
 	}
 	if len(addrs) == 0 {
 		return nil, fmt.Errorf("no address provided to listen on")
 	}
 
 	ml := &multiListener{
-		acceptReadyCh: make(chan any),
+		connChs: make([]chan connErrPair, 0, len(addrs)),
+		readyCh: make(chan struct{}),
+		stopCh:  make(chan struct{}),
 	}
 
 	for _, addr := range addrs {
@@ -98,85 +108,106 @@ func multiListen(
 		ml.listeners = append(ml.listeners, l)
 	}
 
-	for _, l := range ml.listeners {
+	for i, l := range ml.listeners {
+		i := i
 		ml.wg.Add(1)
+		ch := make(chan connErrPair, 1)
+		ml.connChs = append(ml.connChs, ch)
 		go func(l net.Listener) {
+			fmt.Printf("sub[%d]: starting\n", i)
 			defer ml.wg.Done()
 			for {
+				fmt.Printf("sub[%d]: accepting\n", i)
 				conn, err := l.Accept()
-				ml.mu.Lock()
-				if ml.closed {
-					ml.mu.Unlock()
+				pair := connErrPair{conn: conn, err: err}
+				fmt.Printf("sub[%d]: queueing %s\n", i, pair.String())
+
+				// Signal ml.Accept()
+				ch <- pair
+				select {
+				case ml.readyCh <- struct{}{}:
+					fmt.Printf("sub[%d]: sent ready\n", i)
+				case <-ml.stopCh:
+					fmt.Printf("sub[%d]: exiting\n", i)
+					close(ch)
 					return
 				}
-				// enqueue the accepted connection
-				ml.connErrQueue = append(ml.connErrQueue, connErrPair{conn: conn, err: err})
-
-				// signal the waiting ml.Accept() to consume accepted connection from the queue.
-				select {
-				case ml.acceptReadyCh <- struct{}{}:
-				default:
-				}
-				ml.mu.Unlock()
 			}
 		}(l)
 	}
 	return ml, nil
 }
 
-// Accept implements net.Listener.
-// It waits for and returns a connection from any of the sub-listener.
+// Accept implements net.Listener.  It waits for and returns a connection from
+// any of the sub-listener.
 func (ml *multiListener) Accept() (net.Conn, error) {
 	for {
-		// atomically return and remove the first element of the queue if it's not empty
-		ml.mu.Lock()
-		if len(ml.connErrQueue) > 0 {
-			connErr := ml.connErrQueue[0]
-			ml.connErrQueue = ml.connErrQueue[1:]
-			ml.mu.Unlock()
-			return connErr.conn, connErr.err
-		}
-		ml.mu.Unlock()
-
 		// wait for any sub-listener to enqueue an accepted connection
-		_, ok := <-ml.acceptReadyCh
-		if !ok {
-			// The "acceptReadyCh" channel will be closed only when Close() is called on the multiListener.
+		if _, ok := <-ml.readyCh; !ok {
+			// The "readyCh" channel will be closed only when Close() is called on the multiListener.
 			// Closing of this channel implies that all sub-listeners are also closed, which causes a
 			// "use of closed network connection" error on their Accept() calls. We return the same error
 			// for multiListener.Accept() if multiListener.Close() has already been called.
 			return nil, fmt.Errorf("use of closed network connection")
 		}
+
+		for _, ch := range ml.connChs {
+			select {
+			case connErr, ok := <-ch:
+				if !ok {
+					return nil, fmt.Errorf("use of closed network connection")
+				}
+				return connErr.conn, connErr.err
+			default: // nothing there
+			}
+		}
+		// If we get here we got a "ready" but no connection?
 	}
 }
 
-// Close implements net.Listener.
-// It will close all sub-listeners and wait for the go-routines to exit.
+// Close implements net.Listener.  It will close all sub-listeners and wait for
+// the go-routines to exit.
 func (ml *multiListener) Close() error {
-	ml.mu.Lock()
-	if ml.closed {
-		ml.mu.Unlock()
-		return fmt.Errorf("use of closed network connection")
-	}
-	ml.closed = true
-	close(ml.acceptReadyCh)
-	ml.mu.Unlock()
+	// Tell all sub-listeners to stop.
+	close(ml.stopCh)
 
 	// Closing the listeners causes Accept() to immediately return an error,
-	// which serves as the exit condition for the sub-listener go-routines.
+	// which triggers the exit condition for the sub-listener go-routines.
 	for _, l := range ml.listeners {
 		_ = l.Close()
 	}
 
+	// Drain any already-queued connections.
+	for i, ch := range ml.connChs {
+		fmt.Printf("draining [%d]\n", i)
+		for connErr := range ch {
+			fmt.Printf("draining [%d] => %v\n", i, connErr.String())
+			if connErr.conn != nil {
+				connErr.conn.Close()
+			}
+		}
+		fmt.Printf("drained [%d]\n", i)
+	}
+
 	// Wait for all the sub-listener go-routines to exit.
 	ml.wg.Wait()
+	close(ml.readyCh)
+
 	return nil
 }
 
-// Addr is an implementation of the net.Listener interface.
-// It always returns the address of the first listener.
-// Callers should  use conn.LocalAddr() to obtain the actual
-// local address of the sub-listener.
+// Addr is an implementation of the net.Listener interface.  It always returns
+// the address of the first listener.  Callers should  use conn.LocalAddr() to
+// obtain the actual local address of the sub-listener.
 func (ml *multiListener) Addr() net.Addr {
 	return ml.listeners[0].Addr()
 }
+
+// Addrs is like Addr, but returns the address for all registered listeners.
+func (ml *multiListener) Addrs() []net.Addr {
+	ret := []net.Addr{}
+	for _, l := range ml.listeners {
+		ret = append(ret, l.Addr())
+	}
+	return ret
+}
```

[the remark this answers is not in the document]
thockin: If we used `reflect.Select` we could get rid of `readyCh` entirely and just select on the array of `connCh`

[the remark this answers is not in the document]
thockin: yeah, let's not panic/recover inside a library if we can avoid it.  It is costly, if I recall, and I don't think we need it.  We just need clarity on our semantics.

As a general rule, Go asssumes that writers "own" channels and are responsible for closing them "when there's no more data".  It's safe to read a closed channel but not write.

If the semantic we want is approximately "each sub-listener has a bounded number of accepted connections waiting for the app to accept", then that suggests to me that we need a per-listener queue (channels are good at this).  If we prefer "the multi-listener has a bounded number of accepted connections waiting, regardless of which sub-listener", it can get even simpler, at the cost of potentially starving, which I guess can happen anyway unless we iterate the conn channels more "fairly" (e.g. keep track of which one we read last, and then resume there like a ring).  I don't think we need that.

Simpler form.  It may still race, but I think I got them all.  The fun one was between app calling Accept() in a tight loop and sub-listeners.  This presumes we don't yet change to the `AddListener()` model.


```diff
diff --git a/net/multi_listen.go b/net/multi_listen.go
index d508805..6672978 100644
--- a/net/multi_listen.go
+++ b/net/multi_listen.go
@@ -29,20 +29,26 @@ type connErrPair struct {
 	err  error
 }
 
+func (p connErrPair) String() string {
+	if p.conn != nil {
+		return p.conn.RemoteAddr().String()
+	}
+	if p.err != nil {
+		return p.err.Error()
+	}
+	return "neither conn nor err were valid"
+}
+
 // multiListener implements net.Listener
 type multiListener struct {
 	listeners []net.Listener
 
-	wg     sync.WaitGroup
-	mu     sync.Mutex
-	closed bool
+	wg sync.WaitGroup
 
-	// connErrQueue holds the connections accepted by sub-listeners
-	connErrQueue []connErrPair
-
-	// acceptReadyCh is used as a semaphore to wake up the waiting
-	// multiListener.Accept() when new connections are available
-	acceptReadyCh chan any
+	// connCh passes accepted connections, from child listeners to parent.
+	connCh chan connErrPair
+	// stopCh communicates from parent to child listeners.
+	stopCh chan struct{}
 }
 
 // compile time check to ensure *multiListener implements net.Listener
@@ -58,13 +64,13 @@ var _ net.Listener = &multiListener{}
 //  2. Use "tcp4" or "tcp6" to exclusively listen on IPv4 or IPv6 family, respectively.
 //  3. The host can accept names (e.g, localhost) and it will create a listener for at
 //     most one of the host's IP.
-func MultiListen(ctx context.Context, network string, addrs []string) (net.Listener, error) {
+func MultiListen(ctx context.Context, network string, addrs ...string) (net.Listener, error) {
+	var lc net.ListenConfig
 	return multiListen(
 		ctx,
 		network,
 		addrs,
 		func(ctx context.Context, network, address string) (net.Listener, error) {
-			var lc net.ListenConfig
 			return lc.Listen(ctx, network, address)
 		})
 }
@@ -78,14 +84,15 @@ func multiListen(
 	listenFunc func(ctx context.Context, network, address string) (net.Listener, error),
 ) (net.Listener, error) {
 	if !(network == "tcp" || network == "tcp4" || network == "tcp6") {
-		return nil, fmt.Errorf("network '%s' not supported", network)
+		return nil, fmt.Errorf("network %q not supported", network)
 	}
 	if len(addrs) == 0 {
 		return nil, fmt.Errorf("no address provided to listen on")
 	}
 
 	ml := &multiListener{
-		acceptReadyCh: make(chan any),
+		connCh: make(chan connErrPair),
+		stopCh: make(chan struct{}),
 	}
 
 	for _, addr := range addrs {
@@ -98,85 +105,94 @@ func multiListen(
 		ml.listeners = append(ml.listeners, l)
 	}
 
-	for _, l := range ml.listeners {
+	for i, l := range ml.listeners {
+		i := i
 		ml.wg.Add(1)
 		go func(l net.Listener) {
+			fmt.Printf("sub[%d]: starting\n", i) //DBG
 			defer ml.wg.Done()
 			for {
+				fmt.Printf("sub[%d]: accepting\n", i) //DBG
+				// Accept() is blocking, unless ml.Close() is called, in which
+				// case it will return immediately with an error.
 				conn, err := l.Accept()
-				ml.mu.Lock()
-				if ml.closed {
-					ml.mu.Unlock()
+				pair := connErrPair{conn: conn, err: err}
+				terminate := (err != nil)
+
+				select {
+				case ml.connCh <- pair:
+					fmt.Printf("sub[%d]: sent conn %s\n", i, pair.String()) //DBG
+				case <-ml.stopCh:
+					terminate = true
+				}
+				if terminate {
+					fmt.Printf("sub[%d]: exiting\n", i) //DBG
 					return
 				}
-				// enqueue the accepted connection
-				ml.connErrQueue = append(ml.connErrQueue, connErrPair{conn: conn, err: err})
-
-				// signal the waiting ml.Accept() to consume accepted connection from the queue.
-				select {
-				case ml.acceptReadyCh <- struct{}{}:
-				default:
-				}
-				ml.mu.Unlock()
 			}
 		}(l)
 	}
 	return ml, nil
 }
 
-// Accept implements net.Listener.
-// It waits for and returns a connection from any of the sub-listener.
+// Accept implements net.Listener.  It waits for and returns a connection from
+// any of the sub-listener.
 func (ml *multiListener) Accept() (net.Conn, error) {
-	for {
-		// atomically return and remove the first element of the queue if it's not empty
-		ml.mu.Lock()
-		if len(ml.connErrQueue) > 0 {
-			connErr := ml.connErrQueue[0]
-			ml.connErrQueue = ml.connErrQueue[1:]
-			ml.mu.Unlock()
-			return connErr.conn, connErr.err
-		}
-		ml.mu.Unlock()
-
-		// wait for any sub-listener to enqueue an accepted connection
-		_, ok := <-ml.acceptReadyCh
-		if !ok {
-			// The "acceptReadyCh" channel will be closed only when Close() is called on the multiListener.
-			// Closing of this channel implies that all sub-listeners are also closed, which causes a
-			// "use of closed network connection" error on their Accept() calls. We return the same error
-			// for multiListener.Accept() if multiListener.Close() has already been called.
-			return nil, fmt.Errorf("use of closed network connection")
-		}
+	// wait for any sub-listener to enqueue an accepted connection
+	connErr, ok := <-ml.connCh
+	if !ok {
+		// The channel will be closed only when Close() is called on the
+		// multiListener.  Closing of this channel implies that all
+		// sub-listeners are also closed, which causes a "use of closed
+		// network connection" error on their Accept() calls. We return the
+		// same error for multiListener.Accept() if multiListener.Close()
+		// has already been called.
+		return nil, fmt.Errorf("use of closed network connection")
 	}
+	return connErr.conn, connErr.err
 }
 
-// Close implements net.Listener.
-// It will close all sub-listeners and wait for the go-routines to exit.
+// Close implements net.Listener.  It will close all sub-listeners and wait for
+// the go-routines to exit.
 func (ml *multiListener) Close() error {
-	ml.mu.Lock()
-	if ml.closed {
-		ml.mu.Unlock()
-		return fmt.Errorf("use of closed network connection")
-	}
-	ml.closed = true
-	close(ml.acceptReadyCh)
-	ml.mu.Unlock()
+	// Tell all sub-listeners to stop.
+	close(ml.stopCh)
 
-	// Closing the listeners causes Accept() to immediately return an error,
-	// which serves as the exit condition for the sub-listener go-routines.
+	// Closing the listeners causes Accept() to immediately return an error in
+	// the sub-listener go-routines.
 	for _, l := range ml.listeners {
 		_ = l.Close()
 	}
 
 	// Wait for all the sub-listener go-routines to exit.
 	ml.wg.Wait()
+	close(ml.connCh)
+
+	// Drain any already-queued connections.
+	fmt.Printf("draining\n") //DBG
+	for connErr := range ml.connCh {
+		fmt.Printf("draining => %v\n", connErr.String()) //DBG
+		if connErr.conn != nil {
+			connErr.conn.Close()
+		}
+	}
+	fmt.Printf("drained\n") //DBG
+
 	return nil
 }
 
-// Addr is an implementation of the net.Listener interface.
-// It always returns the address of the first listener.
-// Callers should  use conn.LocalAddr() to obtain the actual
-// local address of the sub-listener.
+// Addr is an implementation of the net.Listener interface.  It always returns
+// the address of the first listener.  Callers should  use conn.LocalAddr() to
+// obtain the actual local address of the sub-listener.
 func (ml *multiListener) Addr() net.Addr {
 	return ml.listeners[0].Addr()
 }
+
+// Addrs is like Addr, but returns the address for all registered listeners.
+func (ml *multiListener) Addrs() []net.Addr {
+	ret := []net.Addr{}
+	for _, l := range ml.listeners {
+		ret = append(ret, l.Addr())
+	}
+	return ret
+}
```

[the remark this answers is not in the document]
aroradaman: I tried this approach panic/race is still there. 

Added a unit test to mock race conditions.
```shell
go test k8s.io/utils/net -run TestMultiListen_Accept/accept_some_connections -v --count 10
```
Have tried `reflect.Select` with one channel per sub-listener, race is still there (but the test rarely panics now), will try to come up with a reliable solution.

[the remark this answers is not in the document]
thockin: Running your test against https://github.com/kubernetes/utils/pull/311#discussion_r1655283651 I get no fails, even with 1M runs.  I tried to tickle the races by forcing a sleep in strategic places, but I can't make it happen.  I added some comments.

It only closes `connCh` AFTER the waitgroup is done, so sub-listeners can't write to it after close, and nobody ever writes to `stopCh`, it's just used as a `select` case, so sub-listeners don't block on sending a connection.

https://github.com/kubernetes/utils/compare/master...thockin:kubernetes-utils:pr-311-plus-thockin?expand=1

net/multi_listen.go
```diff
@@ -0,0 +1,195 @@
...
+		go func(l net.Listener) {
+			defer ml.wg.Done()
+			for {
+				// Accept() is blocking, unless ml.Close() is called, in which
+				// case it will return immediately with an error.
+				conn, err := l.Accept()
+				// This assumes that ANY error from Accept() will terminate the
+				// sub-listener. We could maybe be more precise, but it
+				// doesn't seem necessary.
+				terminate := err != nil
```
aroradaman, line 107: @thockin should I just return the error to the consumer instead of closing the listener here?
The behaviour of `multiListener` will be similar to `net.LIstener` if we return the error.

thockin: no, I think we need to return it up thru the channel so the main Accpet can return it

thockin (commented):
Thanks!

/lgtm
/approve

PR #312, opened 2024-06-20
Add WithDeadlineTimer interface to clock

tatsuhiro-t:
**What type of PR is this?**
/kind feature

This is a part of effort to fix https://github.com/kubernetes/kubernetes/issues/125581

WithDeadlineTimer interface extends Clock to add NewDeadlineTimer. NewDeadlineTimer is similar to NewTimer, but it returns a timer which fires at the given timestamp.  It makes unit testing easier with FakeClock.Step when a test subject is making a timer inside goroutine (e.g., workqueue.NewDelayingQueue).  With NewTimer, in some race conditions, a timer never fire if FakeClock.Step is called after getting the base time Now() and calling NewTimer().  This is because the duration passed to timer is calculated against the old timestamp, but inside FakeClock.NewTimer, the targetTime is calculated against the time adjusted by Step.  With NewDeadlineTimer, targetTime is not adjusted, which works around this issue.

Because workqueue.DelayingQueueConfig uses clock.WithTicker and changing it breaks API compatibility, I plan to do type assertion like so to use NewDeadlineTimer if available:

```patch
diff --git a/staging/src/k8s.io/client-go/util/workqueue/delaying_queue.go b/staging/src/k8s.io/client-go/util/workqueue/delaying_queue.go
index 958b96a80c3..e3bebd36987 100644
--- a/staging/src/k8s.io/client-go/util/workqueue/delaying_queue.go
+++ b/staging/src/k8s.io/client-go/util/workqueue/delaying_queue.go
@@ -251,6 +251,14 @@ func (q *delayingType[T]) AddAfter(item T, duration time.Duration) {
        }
 }
 
+type deadlineTimerClock struct {
+       clock.Clock
+}
+
+func (c deadlineTimerClock) NewDeadlineTimer(ts time.Time) clock.Timer {
+       return c.NewTimer(ts.Sub(c.Now()))
+}
+
 // maxWait keeps a max bound on the wait time. It's just insurance against weird things happening.
 // Checking the queue every 10 seconds isn't expensive and we know that we'll never end up with an
 // expired item sitting for more than 10 seconds.
@@ -271,6 +279,11 @@ func (q *delayingType[T]) waitingLoop() {
 
        waitingEntryByData := map[t]*waitFor{}
 
+       deadlineClock, ok := q.clock.(clock.WithDeadlineTimer)
+       if !ok {
+               deadlineClock = deadlineTimerClock{Clock: q.clock}
+       }
+
        for {
                if q.TypedInterface.ShuttingDown() {
                        return
@@ -297,7 +310,7 @@ func (q *delayingType[T]) waitingLoop() {
                                nextReadyAtTimer.Stop()
                        }
                        entry := waitingForQueue.Peek().(*waitFor)
-                       nextReadyAtTimer = q.clock.NewTimer(entry.readyAt.Sub(now))
+                       nextReadyAtTimer = deadlineClock.NewDeadlineTimer(entry.readyAt)
                        nextReadyAt = nextReadyAtTimer.C()
                } 
```

https://github.com/kubernetes/kubernetes/issues/125581

**Release note**:
```
This PR adds WithDeadlineTimer interface extends Clock to add NewDeadlineTimer. NewDeadlineTimer is similar to NewTimer, but it returns a timer which fires at the given timestamp.
```

### DIFF
--- a/clock/clock.go
+++ b/clock/clock.go
@@ -117,10 +117,10 @@ func (RealClock) NewTimer(d time.Duration) Timer {
 	}
 }
 
-// NewDeadlineTimer is the same as time.NewTimer(ts.Sub(time.Now()))
+// NewDeadlineTimer is the same as time.NewTimer(time.Until(ts))
 func (RealClock) NewDeadlineTimer(ts time.Time) Timer {
 	return &realTimer{
-		timer: time.NewTimer(ts.Sub(time.Now())),
+		timer: time.NewTimer(time.Until(ts)),
 	}
 }
 
--- a/clock/clock.go
+++ b/clock/clock.go
@@ -81,6 +81,15 @@ type Ticker interface {
 
 var _ = WithTicker(RealClock{})
 
+// WithDeadlineTimer extends Clock to add NewDeadlineTimer which fires at a given moment of time.
+type WithDeadlineTimer interface {
+	Clock
+	// NewDeadlineTimer returns a new Timer which fires at ts.
+	NewDeadlineTimer(ts time.Time) Timer
+}
+
+var _ = WithDeadlineTimer(RealClock{})
+
 // RealClock really calls time.Now()
 type RealClock struct{}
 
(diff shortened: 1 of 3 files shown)


clock/clock.go
```diff
@@ -108,6 +117,13 @@ func (RealClock) NewTimer(d time.Duration) Timer {
 	}
 }
 
+// NewDeadlineTimer is the same as time.NewTimer(ts.Sub(time.Now()))
+func (RealClock) NewDeadlineTimer(ts time.Time) Timer {
+	return &realTimer{
+		timer: time.NewTimer(ts.Sub(time.Now())),
```
mauri870, line 123: `ts.Sub(time.Now)` can be written as `time.Until(ts)`

tatsuhiro-t: Fixed via 50647950f2abc950ddc3c2e245982966a51d6fb1

mauri870:
@deads2k @mengqiy Could you take a look at this when you have a chance? Thanks!

PR #313, opened 2024-07-01
Support checking if obj is a nil ptr

zhifei92:
**What type of PR is this?**
/kind feature

IsNilPtr checks if the provided object is a nil pointer. This is specially useful when dealing with interfaces that may hold pointers.

Relevant issues:  https://github.com/kubernetes/kubernetes/issues/125561,https://github.com/kubernetes/kubernetes/pull/125563

> the nil checking for interface RuntimeHandlerResolver in #L109 will never check whether the underlying struct is nil.
A reflect nil checking is needed to avoid panic in running 'rcManager.LookupRuntimeHandler'(#L111)

Fixes #

**Release note**:
```

```

### DIFF
--- a/ptr/ptr.go
+++ b/ptr/ptr.go
@@ -71,3 +71,10 @@ func Equal[T comparable](a, b *T) bool {
 	}
 	return *a == *b
 }
+
+// IsNilPtr checks if the provided object is a nil pointer. This is specially useful when dealing
+// with interfaces that may hold pointers.
+func IsNilPtr(obj any) bool {
+	v := reflect.ValueOf(obj)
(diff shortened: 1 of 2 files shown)


ptr/ptr.go
```diff
@@ -71,3 +71,10 @@ func Equal[T comparable](a, b *T) bool {
 	}
 	return *a == *b
 }
+
+// IsNilPtr checks if the provided object is a nil pointer. This is specially useful when dealing
+// with interfaces that may hold pointers.
+func IsNilPtr(obj any) bool {
```
googs1025, line 77: Is this method called from anywhere other than testing?

zhifei92: yes

PR #315, opened 2024-08-20
Deprecate strings/slices functions covered by stdlib slices

tklauser:
**What type of PR is this?**

/kind cleanup

As suggested by @thockin in [1], deprecate the functions in the k8s.io/utils/strings/slices package that are already covered by generic functions in the Go standard library slices package. Also reimplement them in terms of slices functions.

[1] https://github.com/kubernetes/kubernetes/pull/126715#pullrequestreview-2241525172

Fixes #

**Release note**:
```
Deprecate k8s.io/utils/strings/slices functions covered by the Go standard libary slices package.
```

### DIFF
--- a/strings/slices/slices.go
+++ b/strings/slices/slices.go
@@ -20,21 +20,14 @@ limitations under the License.
 // replace "stringslices" if the "slices" package becomes standard.
 package slices
 
+import goslices "slices"
+
 // Equal reports whether two slices are equal: the same length and all
 // elements equal. If the lengths are different, Equal returns false.
 // Otherwise, the elements are compared in index order, and the
 // comparison stops at the first unequal pair.
-func Equal(s1, s2 []string) bool {
-	if len(s1) != len(s2) {
-		return false
-	}
-	for i, n := range s1 {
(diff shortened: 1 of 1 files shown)


strings/slices/slices.go
```diff
@@ -20,21 +20,14 @@ limitations under the License.
 // replace "stringslices" if the "slices" package becomes standard.
 package slices
 
+import "slices"
```
thockin, line 23: I think this should alias the import to "goslices" or "stdslices" for clarity?

thockin (commented):
super minor point.  Is the `use [other] instead" syntax documented somewhere?

tklauser:
> super minor point. Is the `use [other] instead" syntax documented somewhere?

The syntax is documented at https://go.dev/doc/comment#doclinks

strings/slices/slices.go
tklauser: Changed to "goslices", thanks.

thockin:
Cool.TIL.  Thanks!

/lgtm
/approve

Issue #316, opened 2024-08-27
Switching to stdlib slices breaks golang 1.18-1.20 compatibility


gibizer:
**What happened**:
Bumping to the latest version including https://github.com/kubernetes/utils/commit/4965b0f13691de9f60deba4e3e0b4526f27aa76e in a project using golang 1.20 (e.g.https://github.com/openstack-k8s-operators/nova-operator/pull/847) breaks as slices is only added to sdlib in golang 1.21. The commit 4965b0f13691de9f60deba4e3e0b4526f27aa76e did not changed the go.mod file to indicate that this repo now only support 1.21 either. 

**What you expected to happen**:

I expect that such change only happens after https://github.com/kubernetes/utils/pull/308 has been landed.

**How to reproduce it**:

bump to version https://github.com/kubernetes/utils/commit/4965b0f13691de9f60deba4e3e0b4526f27aa76e or newer in an env using golang < 1.21

PR #317, opened 2024-08-27
Revert "Deprecate strings/slices functions covered by stdlib slices"

gibizer:
This reverts commit 4965b0f13691de9f60deba4e3e0b4526f27aa76e as it
incorrectly started using stdlib slices without declaring that this repo
is now only compatible with golang >=1.21.

Closes: #316

### DIFF
--- a/strings/slices/slices.go
+++ b/strings/slices/slices.go
@@ -20,14 +20,21 @@ limitations under the License.
 // replace "stringslices" if the "slices" package becomes standard.
 package slices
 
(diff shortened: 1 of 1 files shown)


gibizer:
@thockin could you check this please? (Sorry for pinging directly but you were the approver of the commit I propose to revert here)

thockin:
Oof. OK.  We could add those "deprecated" comments sooner and change the impl later.

/lgtm
/approve

PR #275, opened 2023-02-09
lru: add support for setting eviction func

enj:
Signed-off-by: Monis Khan <mok@microsoft.com>

/kind feature

**Release note**:

```
lru: add support for setting eviction func
```

PR #318, opened 2024-11-02
Add a helper method to set eviction function after construction

dims:
We have a use case in https://github.com/kubernetes/kubernetes/pull/128507 where we set `OnEvicted` later.

https://github.com/dims/utils/pull/new/add-a-helper-method-to-set-eviction-function-after-construction

**What type of PR is this?**

/kind feature

Fixes #

**Release note**:
```
Add a new `SetEvictionFunc` in lru's Cache
```

### DIFF
--- a/lru/lru.go
+++ b/lru/lru.go
@@ -16,6 +16,7 @@ limitations under the License.
 package lru
 
 import (
+	"fmt"
 	"sync"
 
 	groupcache "k8s.io/utils/internal/third_party/forked/golang/golang-lru"
@@ -42,6 +43,15 @@ func NewWithEvictionFunc(size int, f EvictionFunc) *Cache {
 	c := New(size)
 	c.cache.OnEvicted = f
 	return c
+}
+
+// SetEvictionFunc updates the eviction func
+func (c *Cache) SetEvictionFunc(f EvictionFunc) error {
+	if c.cache.OnEvicted != nil {
+		return fmt.Errorf("lru cache eviction function is already set")
+	}
+	c.cache.OnEvicted = f
+	return nil
 }
 
 // Add adds a value to the cache.
--- a/lru/lru_test.go
+++ b/lru/lru_test.go
@@ -130,3 +130,31 @@ func TestEviction(t *testing.T) {
 		t.Errorf("unexpected eviction data: key=%v val=%v", seenKey, seenVal)
 	}
 }
+
+func TestSetEviction(t *testing.T) {
+	var seenKey Key
+	var seenVal interface{}
+
+	lru := New(1)
+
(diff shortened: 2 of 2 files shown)


lru/lru.go
```diff
@@ -44,6 +44,11 @@ func NewWithEvictionFunc(size int, f EvictionFunc) *Cache {
 	return c
 }
 
+// SetEvictionFunc updates the eviction func
+func (c *Cache) SetEvictionFunc(f EvictionFunc) {
+	c.cache.OnEvicted = f
```
aojea, line 49: should be thread safe and lock before setting the value?

same for `NewWithEvictionFunc`?

aojea:
It seems correct, but it will be nice if you can just add an unit test, looking at this https://github.com/kubernetes/utils/pull/275/files it seems you can just copy paste that but instead of using the constructor you use this new helper

dims:
> It seems correct, but it will be nice if you can just add an unit test, looking at this https://github.com/kubernetes/utils/pull/275/files it seems you can just copy paste that but instead of using the constructor you use this new helper

indeed! i tightened the api a bit, by allowing the set function only once as well. added tests.

aojea:
/lgtm
/approve

Thanks

lru/lru.go
```diff
@@ -44,6 +45,15 @@ func NewWithEvictionFunc(size int, f EvictionFunc) *Cache {
 	return c
 }
 
+// SetEvictionFunc updates the eviction func
+func (c *Cache) SetEvictionFunc(f EvictionFunc) error {
+	if c.cache.OnEvicted != nil {
+		return fmt.Errorf("lru cache eviction function is already set")
```
aojea, line 51: yeah, this seems much safer and reduce the risk of supporting weird behaviors

lru/lru.go
liggitt: > good point! added it here

was that added in a follow-up? not seeing it here

lru/lru.go
dims: ✅

lru/lru.go
dims: @liggitt would you recommend both the lock/unlock in addition to the "return error if it is already set"? i can cut another PR if so

liggitt: > @liggitt would you recommend both the lock/unlock in addition to the "return error if it is already set"?

yeah... going from nil to set still requires a lock since the function can be read/used concurrently from other methods

liggitt: (and disallowing resetting it once set still sgtm)

dims: Follow up is here https://github.com/kubernetes/utils/pull/319

PR #320, opened 2024-11-22
[Fake clock] Make Stop / Reset return false if Timer stopped

antoninbas:
**What type of PR is this?**

> /kind bug

This is a bug in the fake clock implementation, but not one that impacts K8s users.

In order to conform to the standard library Timer, Stop and Reset should return false if the Timer has already been stopped, which was not the case previously.

We add unit tests to validate the new implementation. The updated TestFakeStop test case and the new TestFakeReset/reset_stopped_timer case fail with the old implementation.

We also simplify the implementation of Stop and Rest in the following way: there is no need to check f.waiter.fired to determine whether a Timer has expired. For Timers, this flag is set atomically with the waiter being removed from the waiters list. As a result, we only check for absence of the waiter from the list, which either indicates that the Timer has been stopped or that it has already expired (fired).

**Release note**:
```

```

### DIFF
--- a/clock/testing/fake_clock.go
+++ b/clock/testing/fake_clock.go
@@ -305,44 +305,49 @@ func (f *fakeTimer) C() <-chan time.Time {
 	return f.waiter.destChan
 }
 
-// Stop stops the timer and returns true if the timer has not yet fired, or false otherwise.
+// Stop prevents the Timer from firing. It returns true if the call stops the
+// timer, false if the timer has already expired or been stopped.
 func (f *fakeTimer) Stop() bool {
 	f.fakeClock.lock.Lock()
 	defer f.fakeClock.lock.Unlock()
 
+	active := false
 	newWaiters := make([]*fakeClockWaiter, 0, len(f.fakeClock.waiters))
 	for i := range f.fakeClock.waiters {
 		w := f.fakeClock.waiters[i]
 		if w != &f.waiter {
 			newWaiters = append(newWaiters, w)
+			continue
 		}
+		// If timer is found, it has not been fired yet.
+		active = true
 	}
 
 	f.fakeClock.waiters = newWaiters
 
-	return !f.waiter.fired
+	return active
 }
 
-// Reset resets the timer to the fake clock's "now" + d. It returns true if the timer has not yet
-// fired, or false otherwise.
+// Reset changes the timer to expire after duration d. It returns true if the
+// timer had been active, false if the timer had expired or been stopped.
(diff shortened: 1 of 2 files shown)


liggitt:
thanks for the PR, will defer review to others, but opened https://github.com/kubernetes/kubernetes/pull/128944 in the meantime to make sure unit tests in kubernetes/kubernetes still pass with this change

antoninbas:
Thanks @liggitt - I am happy to help look into any test failures that may arise

antoninbas:
@liggitt it seems that the 2 test failures for your validation PR are related to dependencies, and maybe expected in this case?

I can see that `pull-kubernetes-unit` ran sucessfully.

antoninbas:
@cheftako @dashpole Could I get a review for this?

dims:
/assign @MadhavJivrajani @thockin 

if you want to take a peek, it looks good to me and the change works fine in k/k's unit tests as well.

clock/testing/fake_clock.go
```diff
@@ -305,44 +305,49 @@ func (f *fakeTimer) C() <-chan time.Time {
...
+// Reset changes the timer to expire after duration d. It returns true if the
+// timer had been active, false if the timer had expired or been stopped.
 func (f *fakeTimer) Reset(d time.Duration) bool {
 	f.fakeClock.lock.Lock()
 	defer f.fakeClock.lock.Unlock()
 
-	active := !f.waiter.fired
+	active := false
 
 	f.waiter.fired = false
```
thockin, line 339: Why not get rid of the `fired` field?  AFAICT it is no longer used.

thockin (commented):
This change LGTM, but I think can be cleaned up further - agree or did I miss something?

clock/testing/fake_clock.go
antoninbas: Thanks for pointing this out. I agree it is not used anymore, and I have removed it.

clock/testing/fake_clock.go
```diff
@@ -198,12 +198,10 @@ func (f *FakeClock) setTimeLocked(t time.Time) {
 			if w.skipIfBlocked {
 				select {
 				case w.destChan <- t:
-					w.fired = true
 				default:
```
thockin, line 201: This field is still defined in the struct, L51

antoninbas: oops, sorry about that
removed

thockin:
Thanks!

/lgtm
/approve

antoninbas:
Thanks for the careful review @thockin !

PR #321, opened 2024-12-10
Add Len and Capacity methods for ring buffer

xigang:
**What type of PR is this?**
/kind feature

Add `Len()` and `Capacity()` methods for Informer metrics to calculate `numberOfPendingNotifications` and `sizeOfRingGrowing`.

PR: https://github.com/kubernetes/kubernetes/pull/129160

KEP: https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/4346-informer-metrics

```
type eventHandlerMetrics struct {

  // number of pending data
  numberOfPendingNotifications GaugeMetric

  // size of RingGrowring data
  sizeOfRingGrowing GaugeMetric

  // how long processing an item from informer reflector
  prcoessDuration  HistogramMetric
        
}
```

Fixes #

**Release note**:
```
NONE
```

### DIFF
--- a/buffer/ring_growing.go
+++ b/buffer/ring_growing.go
@@ -70,3 +70,13 @@ func (r *RingGrowing) WriteOne(data interface{}) {
 	r.data[(r.readable+r.beg)%r.n] = data
 	r.readable++
 }
+
+// Len returns the number of items in the buffer.
+func (r *RingGrowing) Len() int {
+	return r.readable
+}
+
+// Cap returns the capacity of the buffer.
+func (r *RingGrowing) Cap() int {
+	return r.n
+}


buffer/ring_growing.go
```diff
@@ -70,3 +70,13 @@ func (r *RingGrowing) WriteOne(data interface{}) {
 	r.data[(r.readable+r.beg)%r.n] = data
 	r.readable++
 }
+
```
liggitt, line 73: does anything ever shrink this buffer?

xigang: @liggitt  No, this ring buffer implementation (RingGrowing) only grows and never shrinks. Looking at the code:

1. The buffer starts with an initialSize specified in `NewRingGrowing`.
2. When the buffer becomes full (r.readable == r.n), the `WriteOne` method doubles the size (newN := r.n * 2).
3. There are no methods that reduce the size of the buffer.

If shrinking capability is needed, a method can be added to shrink the buffer when it is underutilized (for example, when readability falls below a certain threshold, such as 25% of the capacity).

xigang:
ping @deads2k

xigang:
@liggitt Can this PR be approved?

xigang:
xref: https://github.com/kubernetes/kubernetes/pull/129160

xigang:
xref: https://github.com/kubernetes/kubernetes/pull/129160
@thockin @liggitt Could you please review this PR?

buffer/ring_growing.go
```diff
@@ -70,3 +70,13 @@ func (r *RingGrowing) WriteOne(data interface{}) {
...
+// Len returns the number of items in the buffer.
+func (r *RingGrowing) Len() int {
+	return r.readable
+}
+
+// Capacity returns the capacity of the buffer.
+func (r *RingGrowing) Capacity() int {
```
thockin, line 80: Sorry to nitpick, but why short `Len` and long `Capacity`?  Go's `len` and `cap` seem like they should convert to `Len` and `Cap` ?

thockin (commented):
Thanks!

/lgtm
/approve

PR #325, opened 2025-05-02
fake clock: expose the number of waiters

aojea:
By using `Waiters()`, tests can wait until the expected number of waiters are present before manipulating the clock's time, ensuring that the goroutine under test has reached a stable state regarding its timers.

 /kind feature
 /kind flake

When working on https://github.com/kubernetes/kubernetes/pull/131502 it took me a while to find the source of flakiness, the code under test uses two timers and runs in a goroutinge.

There is a race between the test code when updating the fake clock time, that may happen before the timers are reset.

The fake clock already has the `HasWaiters()` method, but that only works for one timer, leaving a small window for the race to happen. In order to avoid that, we need to expose the actual number of waiters to be able to synchronize the test .

**Release note**:
```
NONE
```

/assign @thockin @danwinship

### DIFF
--- a/clock/testing/fake_clock.go
+++ b/clock/testing/fake_clock.go
@@ -221,12 +221,24 @@ func (f *FakeClock) setTimeLocked(t time.Time) {
 	f.waiters = newWaiters
 }
 
-// HasWaiters returns true if After or AfterFunc has been called on f but not yet satisfied (so you can
-// write race-free tests).
+// HasWaiters returns true if Waiters() returns non-0 (so you can write race-free tests).
 func (f *FakeClock) HasWaiters() bool {
 	f.lock.RLock()
 	defer f.lock.RUnlock()
 	return len(f.waiters) > 0
+}
+
+// Waiters returns the number of "waiters" on the clock (so you can write race-free
+// tests). A waiter exists for:
+//   - every call to After that has not yet signaled its channel.
+//   - every call to AfterFunc that has not yet called its callback.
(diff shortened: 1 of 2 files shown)


aojea:
/assign @dims 

cross referencing https://github.com/kubernetes/kubernetes/pull/131502/commits/c805d32ad6791780a1e25d7427875c35e08d3cd5#r2070948260

clock/testing/fake_clock.go
```diff
@@ -229,6 +229,14 @@ func (f *FakeClock) HasWaiters() bool {
 	return len(f.waiters) > 0
 }
 
+// HasWaiters returns the number of waiters, if After or AfterFunc has been called or
+// if a Timer or a Ticker was created on f but not yet satisfied (so you can write race-free tests).
+func (f *FakeClock) Waiters() int {
```
danwinship, line 234: wrong function name in comment

But the comment needs to be more explicit if you want people to be able to use it reliably. Something like:

```
// Waiters returns the number of "waiters" on the clock. A waiter exists for:
//   - every call to After that has not yet signaled its channel.
//   - every call to AfterFunc that has not yet called its callback.
//   - every timer created with NewTimer which is currently ticking.
//   - every ticker created with NewTicker which is currently ticking.
//   - every ticker created with Tick.
```

and then maybe make HasWaiters say that it returns true if Waiters returns non-0.

aojea: ack

PR #328, opened 2025-07-29
netutilsv2

danwinship:
This is the latest version of what was originally https://github.com/kubernetes/kubernetes/pull/122549.

Big ideas:
- remove deprecated stuff
- support netip.Addr and netip.Prefix (for parsing, IPFamily checking, etc)
- add net/netip conversion functions
- make netutils.IPFamily compatible with corev1.IPFamily
- use generics in a few places
- add some more useful stuff we want

I'm not sure that "k8s.io/utils/net/v2" is necessarily the right way to do this. An alternative would be to add all the new stuff to "k8s.io/utils/ip" instead (and deprecate the IP-related methods in utils/net).

https://github.com/danwinship/kubernetes/commits/netutilsv2/ is an example of porting k/k to this, which helps to motivate some of the new stuff (particularly the "MustParse*" and "Parse*List" functions).

/kind feature
/kind cleanup
/cc @aojea


net/v2/ipfamily.go
```diff
@@ -55,29 +95,56 @@ func IsDualStackIPs(ips []net.IP) bool {
...
-// - at least one IP from each family (v4 and v6) is present
-func IsDualStackIPStrings(ips []string) bool {
-	parsedIPs := make([]net.IP, 0, len(ips))
-	for _, ip := range ips {
-		parsedIP := ParseIPSloppy(ip)
-		if parsedIP == nil {
-			return false
+// IsDualStackPair returns true if vals contains exactly 1 IPv4 address and 1 IPv6 address
+// (in either order).
+func IsDualStackPair[T ipOrString](vals []T) bool {
```
danwinship, line 110: This is new. I kept noticing places in k/k that wanted this (half of https://github.com/danwinship/kubernetes/commit/4a88052a89cebe7b209bbdaa5a27c4d100f8cfb9)

net/v2/ipfamily.go
```diff
@@ -90,101 +157,24 @@ func IsDualStackCIDRs(cidrs []*net.IPNet) bool {
...
-	switch {
-	case ip.To4() != nil:
-		return IPv4
-	case ip.To16() != nil:
+// OtherIPFamily returns the other IP family from ipFamily.
+//
+// Note that "k8s.io/utils/net/v2".IPFamily intentionally has identical values to
+// "k8s.io/api/core/v1".IPFamily and "k8s.io/discovery/v1".AddressType, so you can cast
+// the input/output values of this function between these types.
+func OtherIPFamily(ipFamily IPFamily) IPFamily {
```
danwinship, line 187: stolen from `k/k/pkg/proxy/util.go`, but not heavily used in k/k yet

net/v2/convert.go
```diff
@@ -113,6 +114,40 @@ func IPFromAddr(addr netip.Addr) net.IP {
 	return net.IP(addr.AsSlice())
 }
 
+// IPFromInterfaceAddr can be used to extract the underlying IP address value (as a
+// net.IP) from the return values of net.InterfaceAddrs(), net.Interface.Addrs(), or
+// net.Interface.MulticastAddrs(). (net.Addr is also used in some other APIs, but this
+// function should not be used on net.Addrs that are not "interface addresses".)
+func IPFromInterfaceAddr(ifaddr net.Addr) net.IP {
```
danwinship, line 121: See https://github.com/danwinship/kubernetes/commit/922f93d5b21f7ced360c70553085f8801facb2e1

net/v2/parse.go
```diff
@@ -17,17 +17,59 @@ limitations under the License.
...
-// leading '0' characters on numbers.  Go used to allow this and then changed
-// the behavior in 1.17.  We're choosing to keep it for compat with potential
-// stored values.
-var ParseCIDRSloppy = forkednet.ParseCIDR
+// ParseIP parses an IPv4 or IPv6 address to a net.IP. This accepts both fully-valid IP
+// addresses and irregular/ambiguous forms, making it usable for both validated and
+// non-validated input strings. It should be used instead of net.ParseIP (which rejects
+// some strings we need to accept for backward compatibility) and the old
+// netutilsv1.ParseIPSloppy.
+func ParseIP(ipStr string) (net.IP, error) {
```
danwinship, line 34: The addition of the error return value is frequently annoying for callers (https://github.com/danwinship/kubernetes/commit/8cb7e603a921592386189d2b93806227c6e763de, though in other places it's easy to ignore, https://github.com/danwinship/kubernetes/commit/9adeb456a0f9e48207e9bf2731e0b7912de8732e). But this is more consistent (and the addition of `MustParseIP` later helps for unit tests).

Also, making the return value consistent (`(X, error)`) helps with the generic Must and List methods later...

net/v2/parse.go
```diff
@@ -17,17 +17,59 @@ limitations under the License.
...
+// ParseIPNet parses an IPv4 or IPv6 CIDR string representing a subnet or mask, to a
+// *net.IPNet. This accepts both fully-valid CIDR values and irregular/ambiguous forms,
+// making it usable for both validated and non-validated input strings. It should be used
+// instead of net.ParseCIDR (which rejects some strings that we need to accept for
+// backward-compatibility) and the old netutilsv1.ParseCIDRSloppy.
+//
+// The return value is equivalent to the second return value from net.ParseCIDR. Note that
+// this means that if the CIDR string has bits set beyond the prefix length (e.g., the "5"
+// in "192.168.1.5/24"), those bits are simply discarded.
+func ParseIPNet(cidrStr string) (*net.IPNet, error) {
```
danwinship, line 84: Most ParseCIDRSloppy callers ignore the first argument anyway (https://github.com/danwinship/kubernetes/commit/dce0e89a48a09d23e90fb420667775763ded3182). Half of the ones that don't were trying to manually implement what `IPFromInterfaceAddr` does, so they can use that now. (The remainder will use ParseIPAsIPNet, coming in a few commits.)

And again, the consistent return value helps with generics.

net/v2/parse.go
```diff
@@ -77,25 +77,51 @@ func ParseAddr(ipStr string) (netip.Addr, error) {
...
+// strings. It should be used instead of net.ParseCIDR (which rejects some strings that we
+// need to accept for backward-compatibility) and the old netutilsv1.ParseCIDRSloppy.
+//
+// The return value combines the two return values of net.ParseCIDR; the returned
+// *net.IPNet's IP field will contain a net.IP corresponding to the full IP address from
+// the CIDR string, while the Mask field will contain a net.IPMask corresponding to the
+// CIDR's prefix length.
+//
+// Compare ParseAddrAsPrefix, which returns a netip.Prefix, but is otherwise identical.
+func ParseIPAsIPNet(cidrStr string) (*net.IPNet, error) {
```
danwinship, line 101: I don't love the naming here... I occasionally feel like it should be the other way around? (`ParseIPNetAsIP`... though the idea of the existing name is that you're parsing an IP, and getting the value as an IPNet). Maybe I should just use the "ifaddr" naming. (`ParseIFAddrAsIPNet`?)

net/v2/parse.go
```diff
@@ -165,3 +165,33 @@ func ParseAddrAsPrefix(cidrStr string) (netip.Prefix, error) {
...
+		ret, err := parse(str)
+		if err != nil {
+			panic(err)
+		}
+		return ret
+	}
+}
+
+// MustParseIP is like ParseIP, but it panics on error instead of returning an error value.
+var MustParseIP = must(ParseIP)
```
danwinship, line 182: So these get used A LOT in k/k in unit tests, eg https://github.com/danwinship/kubernetes/commit/88ba5a54ba09aea6d4d2ad9add4e674c20bac018, https://github.com/danwinship/kubernetes/commit/e47bc9b8383f4884dbeda588f3ef509cd40a5656.

In my first version of this, I actually had an exported generic method rather than exported individual functions, so callers would do `ip := netutils.MustParse[net.IP]("1.2.3.4")` or `prefix := netutils.MustParse[netip.Prefix]("1.2.3.0/24")`. One problem with this though was that there was no way to use the macros with both `ParseIPNet` _and_ `ParseIPAsIPNet` though, since `netutils.MustParse[*net.IPNet]` can only mean one thing. Oh, another annoying thing is that you need to include the `*` in `*net.IPNet`, but of course not with any of the other types.

So then I thought about `netutils.MustParse[netutils.ParseIP]("1.2.3.4")` or `netutils.MustParse(netutils.ParseIP("1.2.3.4"))` (yeah, you can make that work, https://github.com/kubernetes/kubernetes/pull/127882). But I didn't like the fact that you have to say `netutils` twice... So anyway, that led to what you see here.

net/v2/parse.go
```diff
@@ -195,3 +195,65 @@ var MustParseIPAsIPNet = must(ParseIPAsIPNet)
...
+var MustParseAddrList = mustlist(ParseAddrList)
+
+// MustParsePrefixList parses a list of strings with ParsePrefix and returns a []netip.Prefix or else panics on error.
+var MustParsePrefixList = mustlist(ParsePrefixList)
+
+// MustParseIPAsIPNetList parses a list of strings with ParseIPAsIPNet and returns a []*net.IPNet or else panics on error.
+var MustParseIPAsIPNetList = mustlist(ParseIPAsIPNetList)
+
+// MustParseAddrAsPrefixList parses a list of strings with ParseAddrAsPrefix and returns a []netip.Prefix or else panics on error.
+var MustParseAddrAsPrefixList = mustlist(ParseAddrAsPrefixList)
```
danwinship, line 259: I'd also thought about having a canonicalization wrapper, that returns an error if the parsed result doesn't re-stringify back to what you passed in. But then, like, do we need to have `MustParseIPListCanonical` and every other possible combination? So, that's a point in favor of exposing the generic functions rather than exposing the specific methods, so then someone could do `netutils.MustParse[netutils.ParseList[netutils.Canonical[netutils.ParseIP]]]`...

or not...

net/v2/parse.go
danwinship: oh, the renaming from `ParseCIDR` to `ParseIPNet` is to be consistent with `ParseIP`, `ParseAddr`, and `ParsePrefix`; each function uses the basename of the type it returns.

net/v2/parse.go
danwinship: The other reason I'd originally had `netutils.MustParse[net.IP]()` was because I was also originally trying to use `netutils.Parse[net.IP]` and `netutils.Parse[netip.Addr]` rather than having separate functions for the two types. But that was based on the idea that golang would do more type inference than it actually does and so you'd be able to omit the type constraints just like you can with the genericized `IPFamilyOf`. But that doesn't actually work; golang never infers the type of a return value, so you end up always needing to include the constraint, so it ends up being ugly.

net/v2/parse.go
danwinship: > Why is this "ParseIP..." and not "ParseIPNet..." or "ParseCIDR..." if the input is "cidrStr"?

So as I said, I'm still not quite sure how to name these functions correctly, but the important point is that `cidrStr` is an IP address written in CIDR form.

danwinship:
> Some of the "when is it a pointer and when is it a value" is annoying, but that seems like vestiges of legacy `net` APIs.

Yes. `*net.IPNet` should be the only visible `*`s in the API, and making the API use plain `net.IPNet` instead would be weird and would just force callers to `*` or `&` values elsewhere.

> Should we mark all new APIs which use `net` types as deprecated-at-launch and start the process of converting to newer `netip` APIs?
> 
> Should we lean on the word "Legacy" or "Old" ?

When we had first started talking about this back when `netip` first appeared, I think that's how we were thinking about it. But the ecosystem as a whole is not moving toward `netip` very quickly (there are currently only 4 files in k/k/vendor/ that import `"net/netip"`) so we will still need to use the `net` types in new code for quite a while.

net/v2/parse.go
```diff
@@ -0,0 +1,259 @@
...
+// addresses and irregular/ambiguous forms, making it usable for both validated and
+// non-validated input strings. It should be used instead of net.ParseIP (which rejects
+// some strings we need to accept for backward compatibility) and the old
+// netutilsv1.ParseIPSloppy.
+//
+// Compare ParseAddr, which returns a netip.Addr but is otherwise identical.
+func ParseIP(ipStr string) (net.IP, error) {
+	// Note: if we want to get rid of forkednet, we should be able to use some
+	// invocation of regexp.ReplaceAllString to get rid of leading 0s in ipStr.
+	ip := forkednet.ParseIP(ipStr)
```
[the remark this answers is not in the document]
danwinship: KEP-4858 doesn't do anything about config files/CLI arguments. We currently don't even _warn_ about leading 0s in that case.

In the larger ecosystem, there may also be aggregated APIs that didn't get updated along with KEP-4858, CRDs with pre-kubernetes#121912 IP/CIDR validation, and IPs/CIDRs in annotations.

We _could_ explicitly distinguish legacy-vs-"correct" parsing in the API like we now do for `utilvalidation.IsValidIP` / `utilvalidation.IsValidIPForLegacyField`, but we'd have to do it separately for the `net` and `netip` types, so that would be 4 new base functions, plus, potentially, corresponding `Must`, `List`, etc functions... (unless we can find a better way to deal with the API [combinatoric explosion](https://github.com/kubernetes/utils/pull/328#discussion_r2241070784)...)

net/v2/ipfamily.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// IPFamilyOfCIDR returns the IP family of val (or IPFamilyUnknown if val is nil or
+// invalid). IPv6-encoded IPv4 addresses (e.g., "::ffff:1.2.3.0/120") are considered IPv4.
+// val can be a *net.IPNet, a netip.Prefix, or a string containing a single CIDR value.
+//
+// Note that "k8s.io/utils/net/v2".IPFamily intentionally has identical values to
+// "k8s.io/api/core/v1".IPFamily and "k8s.io/discovery/v1".AddressType, so you can cast
+// the return value of this function to those types.
+func IPFamilyOfCIDR[T cidrOrString](val T) IPFamily {
```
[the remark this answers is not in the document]
danwinship: Oh, yeah, it's totally un-go-like, as seen by the ugliness of the implementation. (But it feels right when you actually _use_ the functions...) As commented elsewhere, I did it this way to avoid needing to add another set of methods for the `netip` types, but given that we need separate parsing methods for `net` and `netip` types, maybe having separate categorization methods for them too isn't awful.

net/v2/parse.go
```diff
@@ -0,0 +1,259 @@
...
+	return func(str string) T {
+		ret, err := parse(str)
+		if err != nil {
+			panic(err)
+		}
+		return ret
+	}
+}
+
+// MustParseIP is like ParseIP, but it panics on error instead of returning an error value.
```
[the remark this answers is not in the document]
thockin: > I feel like we ought to be able to use MustParseIP when parsing already-validated fields from API objects.

I guess that's OK, but is it better than just assigning the error to `_` ?

> Like, change all the IP-valued fields from string to "metav1.IP", which is something like struct { Raw string; Canonical string; Parsed netip.Addr } so that then people who need the parsed value already have it...)

I love that, and I wish we had done that.  The problem is that the type definitions we use for clients of the API (e.g. k,ube-proxy) are also used to emit protobuf and docs and everything else.  Changing anything is a breaking change.

If I have my way (declarative validation leading the charge), we end up with an IDL which describes the on-the-wire schema (and generates openapi docs and proto, etc) and ALSO generates client code which would be wonderful for what you are describing.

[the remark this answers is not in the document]
thockin: As for a testing sub-pkg, I don't think we need to go that far, I think comments alone are useful.

net/v2/ipfamily.go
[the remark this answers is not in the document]
thockin: It's more go-ish to just explode the number of functions and then fan-in on the impl.  CF `go doc regexp`, which says:

```
There are 16 methods of Regexp that match a regular expression and identify the
matched text. Their names are matched by this regular expression:

    Find(All)?(String)?(Submatch)?(Index)?
```

thockin:
Kubernetes is big in the Go ecosystem -- we should embody the change we want to see.  This looks like a great move directionally.

net/v2/parse.go
danwinship: Yes, it's syntactically a CIDR string, but it's semantically an IP address.

99% of CIDR strings in Kubernetes represent either a subnet (as in `node.status.podCIDRs[0]`) or a mask (as in `networkPolicy.spec.egress[0].to[0].ipBlock.cidr`), and that's what `ParseIPNet`/`ParsePrefix` are for.

This function is for CIDR strings that represent an IP address with associated information about its subnet attached (eg "192.168.1.5/24") as in `resourceClaim.status.devices[0].networkData.ips[0]`. (And nowhere else in core k8s, though they're common in CNI plugins because that's how IPs on network interfaces are normally represented.)

(We need separate functions for subnet/mask-like CIDRs and ifaddr-like CIDRs to preserve backward-compatible handling of pre-KEP-4858 CIDR fields: traditionally, if a user put a value like "192.168.1.5/24" into a NetworkPolicy ipBlock, code using `ParseCIDRSloppy` would see it as though you had said "192.168.1.0/24", and so `ParseIPNet`/`ParsePrefix` work the same way. You need the new function for cases like the DRA field, where you _don't_ want that backward-compatible ignoring of the lower bits.)

Since this is mostly used for addresses on interfaces, we call it "ifaddr" format in some places, and I'd mentioned we could do that here (`ParseIFAddrAsIPNet` / `ParseIFAddrAsPrefix`).

net/v2/parse.go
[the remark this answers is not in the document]
danwinship: > WRT validation, I think we can get to a place where we do not allow new uses of leading-zero IPs in the API and config files.

Right; we already basically have that for API. I forget what the last new legacy IP field was, but it was actually added several releases prior to the start of KEP-4858.

I kind of ended up punting on config files, and maybe I need to get back to that. Config validation generally doesn't use `utilvalidation`, so they are _not_ using strict validation yet, even conditionally. (Or even getting warnings). (I think the failure to use `utilvalidation` is because the same code gets used to validate config file and CLI args, and it would be confusing to give errors in terms of field paths when the user had provided the value via the CLI?)

But config files are also impossible to ratchet, because they have no way of distinguishing config values the admin just added to the file, from config files that the admin added to the file 5 releases ago. So for config files, we'd need to warn about bad values for a while, and then eventually just start rejecting them unconditionally.

> can we ever get to a place where we don't pass the value on to clients (so libraries like this don't need to handle them)? I forget if your KEP covered this?

We deferred any decisions about fixing existing data (https://github.com/kubernetes/enhancements/pull/4899#discussion_r1949955099). This has been brought up multiple times since we first started scrambling to deal with the original golang behavior change (eg, initial comment of https://github.com/kubernetes/kubernetes/issues/108074) but I don't think we've ever _deeply_ thought about the pros and cons.

> Or are we FOREVER going to have `forkednet`

We can get rid of `forkednet` with some `regex` if we really want to...

net/v2/convert.go
```diff
@@ -0,0 +1,202 @@
...
+
+package net
+
+import (
+	"net"
+	"net/netip"
+	"strings"
+)
+
+// AddrFromIP converts a net.IP to a netip.Addr. Given valid input this will always
```
aojea, line 25: This should be also part of golang/net library IMHO , it is a very dangerous behavior and subtle

net/v2/convert.go
aojea: seems very specific for our specific use case for a library, despite the name and the description clarifies that perfectly, the output of the InterfaceAddrs is `net.Addr` that is a very generic interface https://pkg.go.dev/net#Addr , it also may have protocol ip and port or ip and network

net/v2/ipfamily.go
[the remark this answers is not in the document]
aojea: I was trying to find examples in other libraries of this usage but I couldn't find it ... generics provide the benefit of type safety at compile time, but since one of the types set is a string I wonder the benefits vs casting an interface

net/v2/multi_listen.go
```diff
@@ -0,0 +1,195 @@
...
+	// Make sure this can be called repeatedly without explosions.
+	select {
+	case <-ml.stopCh:
+		return fmt.Errorf("use of closed network connection")
+	default:
+	}
+
+	// Tell all sub-listeners to stop.
+	close(ml.stopCh)
+
```
aojea, line 161: I think this was my fault on the original PR , anyway you should update the code to fix this possible panic

https://github.com/kubernetes/utils/pull/329

net/v2/parse.go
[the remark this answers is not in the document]
aojea: are config and flags really a problem? I do not feel they are, IIRC the problem was with the APIs because different actors can interpret different meanings but flags and configs are interpreted only by the component so there can not be a problem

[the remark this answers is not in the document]
danwinship: Yeah, for config/flags the only thing we would really have to worry about is a component reading a "bad" value and then passing it to someone else.

But actually, since there's no "reconciliation loop" problem with config/cli, we could just do repair-on-read and always pass the canonicalized value to the component, right?

So I guess the fix for config/CLI is to add some helpers to log warnings and then canonicalize the values when reading in bad config values.

PR #330, opened 2025-08-20
Replace forked net parsers with regexp hacks

danwinship:
Proof-of-concept, since I keep mentioning we could do this if we wanted. I'm not sure we _do_ want to, but...

Note that we only use the regexps if the initial attempt at parsing fails, so we only take the regexp hit on invalid or legacy inputs.

/kind cleanup
/cc @aojea @thockin


danwinship:
/close

I did this because I thought Antonio wanted to get rid of the forked parsers, but he says he'd rather have those than regexp hacks.

aojea:
mandatory xkcd XD

image

PR #331, opened 2025-08-27
Ensure `Close()` waits for goroutines when called concurrently

ash2k:
**What type of PR is this?**

/kind bug

See the discussion at https://github.com/kubernetes/utils/pull/329#issuecomment-3206258339.

Fixes #

**Release note**:
```
NONE
```

### DIFF
--- a/net/multi_listen.go
+++ b/net/multi_listen.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"net"
 	"sync"
-	"sync/atomic"
 )
 
 // connErrPair pairs conn and error which is returned by accept on sub-listeners.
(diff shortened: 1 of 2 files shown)


net/multi_listen.go
```diff
@@ -152,29 +151,27 @@ func (ml *multiListener) Accept() (net.Conn, error) {
...
-			_ = connErr.conn.Close()
+		// Drain any already-queued connections.
+		for connErr := range ml.connCh {
+			if connErr.conn != nil {
+				_ = connErr.conn.Close()
+			}
 		}
-	}
+	})
 	return nil
```
ash2k, line 175: I think it doesn't make sense to return an error for double close. Either we return the actual underlying errors or no error at all. Any reason to create "artificial" errors? I.e. nothing bad happened because of the second call, why return an error?

PR #336, opened 2025-12-17
Migrate google/btree to k8s util

pjsharath28:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

The `github.com/google/btree` library used by k8s and etcd has been archived and is no longer maintained. This raises long-term concerns around security fixes and CVEs.

We discussed multiple options to address this in the [issue](https://github.com/etcd-io/etcd/issues/20991#issuecomment-3616630633). Given that we only rely on a single source file from the library, we concluded that copying the code into `k8s.io/utils` is the simplest option. This avoids introducing or maintaining a separate fork while preserving identical behaviour.
 
#### Which issue(s) this PR is related to:
Discussion issue: https://github.com/etcd-io/etcd/issues/20991

#### Special notes for your reviewer:
Upstream Google attribution is preserved and the file header follows Kubernetes conventions while remaining Apache-2.0 compliant.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```


btree/btree.go
```diff
@@ -0,0 +1,860 @@
...
+	return
+}
+
+// ItemIterator allows callers of {A/De}scend* to iterate in-order over portions of
+// the tree.  When this function returns false, iteration will stop and the
+// associated Ascend* function will immediately return.
+type ItemIterator[T any] func(item T) bool
+
+// Ordered represents the set of types for which the '<' operator work.
+type Ordered interface {
```
skitt: Ah, I missed this previously, sorry — please use `cmp.Ordered` instead.

pjsharath28: Good catch, thanks
I've resolved it as well

pjsharath28:
@skitt, @dims, I've addressed all comments.
Please review when you're back. Thanks.

dims:
@pjsharath28 can you please move the code to `third_party/forked` so it is pretty clear where we got this from? 

see the existing code in there and the additional files that give it context (we have one in there forked from golang)

dims:
@pjsharath28 please squash and we are good to goo

pohly:
@pjsharath28: thanks for your patience!

/lgtm
/hold cancel

PR #343, opened 2026-02-05
Add dump package from k8s.io/apimachinery

dims:
Move the dump package from k8s.io/apimachinery/pkg/util/dump to k8s.io/utils to make it available for use by other projects without requiring the full apimachinery dependency.

Example: https://github.com/kubernetes-sigs/kustomize/blob/6c8c9cc0e882d5d0a7903543216a402d355185b9/kyaml/yaml/fns.go#L854

This will make it easier to ensure k8s code itself does not have direct calls to `go-spew` package itself as well.

The package provides the same three utility functions for pretty-printing Go objects using spew as was there in kubernetes:
- Pretty(): multi-line indented output with disabled methods
- ForHash(): deterministic output suitable for hashing (sorted keys)
- OneLine(): compact single-line output

code is from:
https://github.com/kubernetes/kubernetes/tree/dc1ec1211e4f54064ba6dafd8aac46ac3d4379b4/staging/src/k8s.io/apimachinery/pkg/util/dump

**What type of PR is this?**

/kind feature

>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

**Release note**:
```
Adding a new dump package with 3 new methods.
```

### DIFF
--- a/dump/dump.go
+++ b/dump/dump.go
@@ -1,0 +1,54 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package dump
+
+import (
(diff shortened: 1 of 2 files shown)


skitt:
/hold for review by @thockin and/or @liggitt, feel free to cancel @dims if my review is enough

I’ve checked that this is strictly identical to the files in `apimachinery`. +1 for making useful packages available without pulling in all the dependency tree…

/lgtm

liggitt:
> /hold for review by @thockin and/or @liggitt, feel free to cancel @dims if my review is enough



I don't have a strong opinion either way

dims:
/hold cancel

thanks @liggitt @skitt

PR #348, opened 2026-04-14
Add ParallelStep method to Trace to support logging concurrent work with an explicit duration

richabanker:
**What type of PR is this?**
/kind feature

Adds `ParallelStep` method to Trace to support logging concurrent work with an explicit duration.

Issue #https://github.com/kubernetes/kubernetes/issues/136002

**Release note**:
```

```

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -82,6 +82,10 @@ type traceStep struct {
 	stepTime time.Time
 	msg      string
 	fields   []Field
+	// explicit duration override; nil means infer from timestamps
+	duration *time.Duration
+	// if true, renders with ||| prefix to indicate concurrent work.
+	parallel bool
 }
 
 // rLock doesn't need to do anything because traceStep instances are immutable.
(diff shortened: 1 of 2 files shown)


trace/trace.go
richabanker, line 276: @serathius @dashpole I am wondering how should lastStepTime behave for a parallel step? 

If ParallelStep advances lastStepTime normally, any sequential step that follows is measured from when ParallelStep was recorded. But if we consider a case like

```
t=0    trace starts; background i/o goroutine launched, CPU work begins on main thread
t=50  i/o finishes - ParallelStep("i/o", 50ms) advances lastStepTime to t=50
t=70   CPU finishes - Step("cpu computation") shows 20ms ---> should this be 70ms ?
``` 

so, should ParalleStep freeze lastStepTime to support this case ?

serathius:
cc @p0lyn0mial

trace/trace.go
richabanker: I think we could also just call this `AggregatedStep` or something, if the only use case is to provide the aggregate of durations spent in various stages of a streamed response. `ParallelStep` seems to indicate that this could happen in the middle of a trace, which I am not sure is something we want (probably needs bigger discussion, and is also not the use case we are trying to solve in https://github.com/kubernetes/kubernetes/pull/137181)

serathius: SG

dashpole:
Here is an alternative: https://github.com/kubernetes/kubernetes/pull/138873

richabanker:
/close

lets try to push https://github.com/kubernetes/kubernetes/pull/138873, thanks @dashpole !

PR #349, opened 2026-05-07
[WIP] fswatch: add cross-platform filesystem watcher

dims:
A small Kubernetes compatibility layer that removes direct github.com/fsnotify/fsnotify dependencies from the Linux build closure. Linux uses raw inotify(7) (epoll + pipe wakeup) inline; non-Linux platforms wrap fsnotify behind a build-tag-isolated adapter so fsnotify is absent from production Linux binaries.

WatchDir watches dir non-recursively; on Add(dir) failure (initial or self Remove/Rename) it reports through the error handler and keeps ticking onChange on the recheck interval, retrying Add(dir) on each tick until the watch comes back online — so manifest reloaders do not silently stop on transient ENOENT/ENOSPC.

WatchFile watches the parent directory so atomic-rename updates are observed; WithRecheckInterval fires onChange unconditionally each tick to support callers that retry transient apply failures, while filesystem events still go through an lstat-based change check. WithInitialCallback fires after the watch (or fallback poll) is in place to avoid a race window with the first event.

Add and Remove hold the lifecycle mutex across the underlying syscall so a concurrent Close cannot tear down the FD mid-call.

**What type of PR is this?**

/kind feature

Fixes #

**Release note**:
```
Adding a new fswatch package with implementation for linux and redirects to fsnotify/fsnotify for non-linux environments
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,11 @@ go 1.23
 
 require (
 	github.com/davecgh/go-spew v1.1.1
+	github.com/fsnotify/fsnotify v1.9.0
 	k8s.io/klog/v2 v2.80.1
 )
 
-require github.com/go-logr/logr v1.2.0 // indirect
+require (
+	github.com/go-logr/logr v1.2.0 // indirect
+	golang.org/x/sys v0.13.0 // indirect
+)
--- a/fswatch/doc.go
+++ b/fswatch/doc.go
@@ -1,0 +1,21 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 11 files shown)


dims:
xref: https://github.com/kubernetes/kubernetes/issues/138812

dims:
xref: https://github.com/kubernetes/kubernetes/pull/138862

go.mod
```diff
@@ -4,7 +4,11 @@ go 1.23
...
 require (
 	github.com/davecgh/go-spew v1.1.1
+	github.com/fsnotify/fsnotify v1.9.0
 	k8s.io/klog/v2 v2.80.1
 )
 
-require github.com/go-logr/logr v1.2.0 // indirect
+require (
+	github.com/go-logr/logr v1.2.0 // indirect
+	golang.org/x/sys v0.13.0 // indirect
```
BenTheElder, line 13: This needs CVE patches pretty often, and now gets brought into the tree for anyone using k8s.io/utils which *almost* didn't have dependencies before (IMHO we should really try to factor out go-spew at some point, klog seems reasonable)

danwinship:
ok but why?

If we don't trust fsnotify, then we shouldn't trust it on Windows either, and if we do, then we don't need to get rid of it on Linux...

(Alternatively, to the extent that there's an implied "we don't actually care about Windows", we could just have `fswatch_others.go` do polling from a goroutine rather than using OS-specific APIs, which then solves Ben's problem.)

PR #353, opened 2026-06-16
Provide a Go 1.26 new() version of ptr.To()

skitt:
**What type of PR is this?**

/kind cleanup

This provides a Go 1.26-specific version of `ptr.To()` delegating to `new()`. The main purpose of this is to allow Go 1.26 codebases relying on `ptr.To()` to automatically migrate to `new()` using `go fix`.

Fixes #

**Release note**:
```
NONE
```

### DIFF
--- a/ptr/ptr.go
+++ b/ptr/ptr.go
@@ -46,11 +46,6 @@ func AllPtrFieldsNil(obj interface{}) bool {
 	return true
 }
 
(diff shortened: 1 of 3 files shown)


skitt:
Ah, this is useless for now, `go fix` doesn’t support type parameter inference.

PR #354, opened 2026-06-26
cpuset: add NumCPU to read the host's online CPU count from sysfs

dims:
NumCPU reads /sys/devices/system/cpu/online and returns the number of online CPUs. Unlike runtime.NumCPU it is not affected by the calling process's CPU affinity (its cgroup cpuset), so it reflects the whole machine, which is what callers want when sizing a host-wide resource by CPU count.

**What type of PR is this?**

/kind feature

Fixes #

Based on reviewer request from https://github.com/kubernetes/kubernetes/pull/140021

**Release note**:
```
Add a new "NumCPU" method that returns the number of CPUs that are online
```

### DIFF
--- a/cpuset/numcpu.go
+++ b/cpuset/numcpu.go
@@ -1,0 +1,50 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


sbueringer:
Thx!

/lgtm

PR #355, opened 2026-07-06
Bump to Go 1.25 and apply automated fixes (go fix)

skitt:
**What type of PR is this?**

/kind cleanup

This bumps to Go 1.25, which is the oldest version used by currently-supported `k/k` branches.

It also applies all automated `go fix` fixes; the commits before the 1.25 bump are valid for Go 1.23, those after it require Go 1.25. The main purpose of this is to set a baseline so that `go fix` can be used in future PRs (for example with `go:inline` directives).

Fixes #

**Release note**:
```
NONE
```

### DIFF
--- a/dump/dump_test.go
+++ b/dump/dump_test.go
@@ -77,7 +77,7 @@ func TestPretty(t *testing.T) {
 	tebw := embedwrap{teb, &teb}
 
 	testCases := []struct {
-		a    interface{}
+		a    any
 		want string
 	}{
 		{int8(93), "(int8) 93\n"},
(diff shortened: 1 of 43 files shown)


dump/dump_test.go
```diff
@@ -130,7 +130,7 @@ func TestPretty(t *testing.T) {
 		{tebw, "(dump.embedwrap) {\n  embed: (dump.embed) {\n    s: (string) (len=4) \"test\"\n  },\n  e: (*dump.embed)({\n    s: (string) (len=4) \"test\"\n  })\n}\n"},
 		{map[string]int{}, "(map[string]int) {\n}\n"},
 		{map[string]int{"one": 1}, "(map[string]int) (len=1) {\n  (string) (len=3) \"one\": (int) 1\n}\n"},
-		{map[string]interface{}{"one": 1}, "(map[string]interface {}) (len=1) {\n  (string) (len=3) \"one\": (int) 1\n}\n"},
+		{map[string]any{"one": 1}, "(map[string]interface {}) (len=1) {\n  (string) (len=3) \"one\": (int) 1\n}\n"},
```
skitt, line 133: go-spew doesn’t know about `any`…

dims: ack

PR #359, opened 2026-09-16
Add tests for Cache RemoveOldest, Len and Clear

magic-peach:
**What type of PR is this?**
/kind test

Cache.RemoveOldest, Cache.Len, and Cache.Clear had zero test coverage. Adds direct tests for each, including RemoveOldest and Clear on an empty cache.

Fixes #

Verified with go test ./lru/...: all 10 tests pass, including the pre-existing race and eviction tests.

**Release note**:
```
NONE
```


skitt:
See https://kubernetes.slack.com/archives/C01672LSZL0/p1789992857325439

/close

